PR #55, opened 2021-11-02
Tuned schemas for Java code generator

ValentinZakharov:
In java, we have POJO generated automatically from JSON schemas. When class generated it use json-file name or title field for generated class name. Since, most of filenames are versions (e.g. `0.1.0.json`) it's generates ambiguous class names like `010`, `010_1`, `010_2` etc.
To bring better naming for generated class, I've added title field, which used for class names.

Another one problem, there is difficult consistently describe dictionary in json-schema and there is no suitable way to generate java classes which uses Map (i.e. dictionary).
The best workaround I found was to add java-specific tag `existingJavaType`, which used by java code generator to define Map. The `existingJavaType` tag is ignored by any other parsers, so should not be any problem.

### DIFF
--- a/utils/interfaces/schemas/miscs/appsec/contexts/actor/0.1.0.json
+++ b/utils/interfaces/schemas/miscs/appsec/contexts/actor/0.1.0.json
@@ -1,6 +1,7 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
   "$id": "/miscs/appsec/contexts/actor/0.1.0.json",
+  "title": "Actor",
   "definitions": {
   },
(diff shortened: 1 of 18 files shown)


nizox (approved):

cbeauchesne (approved):
@ValentinZakharov , If your lib relies on the "java validity" of the schemas, I strongly advise you to create a small Dockerfile file that validate those schemas from a Java point on view, and add it in "test the test" CI. Otherwise, We're 100% sure that we'll break it one day, and by murphy's law, at the worst moment ever for you :)

PR #56, opened 2021-11-03
AppSec addresses don't have semi-colons

nizox:
This PR changes the WAF attack assertion to check for the address and the key path separately.


ValentinZakharov (approved):

cbeauchesne:
@nizox Fixed. Though, @Julio-Guerra told me that `key_path` will fully be supported later. It may be a good idea to not check it in current tests, and add a dedicated test class for it (skipped for everybody but node) WDYT ?

cbeauchesne (approved):

nizox:
I think it's fine to skip tests using the key path for incompatible libraries.

PR #58, opened 2021-11-03
Fix trace tests

Julio-Guerra:
- APPSEC_KEEP didn't get approved so MANUAL_KEEP should be validated instead
- Check for every appsec span tags in case of an event, and not only the event-specific tags.
- Rework the test file.
- Add the missing user agent span tag in Go.

### DIFF
--- a/tests/appsec/test_traces.py
+++ b/tests/appsec/test_traces.py
@@ -2,9 +2,9 @@
 # This product includes software developed at Datadog (https://www.datadoghq.com/).
 # Copyright 2021 Datadog, Inc.
 
-from utils import BaseTestCase, context, interfaces, released, bug, missing_feature
 import pytest
 
+from utils import BaseTestCase, context, interfaces, released, bug, missing_feature
 
 if context.weblog_variant == "echo-poc":
     pytestmark = pytest.mark.skip("not relevant: echo is not instrumented")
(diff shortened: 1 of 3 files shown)


tests/appsec/test_traces.py
```diff
@@ -15,9 +15,14 @@
...
 class Test_AppSecEventSpanTags(BaseTestCase):
     def test_appsec_event_span_tags(self):
-        """Spans with AppSec events should have the general AppSec span tags, along with the appsec.event and _sampling_priority_v1 tags"""
+        """
+        Spans with AppSec events should have the general AppSec span tags, along with the appsec.event and
+        _sampling_priority_v1 tags
+        """
 
         @missing_feature(library="ruby", reason="can't report user agent with dd-trace-rb")
+        @missing_feature(library="dotnet",
```
cbeauchesne, line 24: This decorator should be on the test function (`test_appsec_event_span_tags`), not here

Julio-Guerra: Oops, my bad, now fixed.

tests/appsec/test_traces.py
```diff
@@ -33,42 +39,48 @@ def validate_appsec_span(span):
...
                 raise Exception("Metric _sampling_priority_v1 should be set on traces that are manually kept")
 
-            if span["metrics"]["_sampling_priority_v1"] != APPSEC_KEEP:
-                raise Exception(f"Trace id {span['trace_id']} , sampling priority should be {APPSEC_KEEP}")
+            MANUAL_KEEP = 2
+            if span["metrics"]["_sampling_priority_v1"] != MANUAL_KEEP:
+                raise Exception(f"Trace id {span['trace_id']} , sampling priority should be {MANUAL_KEEP}")
 
-            return True
+            return validate_appsec_span_tags(span)
```
cbeauchesne, line 46: Useles no? `test_custom_span_tags` test it for all span.

tests/appsec/test_traces.py
```diff
@@ -33,42 +39,48 @@ def validate_appsec_span(span):
...
+
+        interfaces.library.add_span_validation(validator=validate_custom_span_tags)
+
+
+RUNTIME_FAMILY = ["nodejs", "ruby", "jvm", "dotnet", "go", "php", "python"]
+
 
-            if "_dd.appsec.enabled" not in span["metrics"]:
-                raise Exception("Can't find _dd.appsec.enabled in span's metrics")
+def validate_appsec_span_tags(span):
```
cbeauchesne, line 68: If my remark above is right, you can move this test logic inside `validate_custom_span_tags`

tests/appsec/test_traces.py
Julio-Guerra: Those span tags are not checked otherwise. I would check them when an attack happens too.

cbeauchesne (approved):

PR #59, opened 2021-11-03
go: simpler support for variants

Julio-Guerra:
- Add gorilla variant
- Simplify how variants are managed with go files, one per variant
- As a result, share the same Dockerfile and pass the variant as build
- Rename net-http-poc into net-http


cbeauchesne (changes requested):
> As a result, share the same Dockerfile and pass the variant as build

I'd like to keep the same build logic for all weblog (e.g. one dockerfile per variant). 

I'm open to discuss about revamp it, and make one dockerfile per lang (but I bet it'll be not that relevant for other lang), but it must be done on all weblogs, and in a dedicated PR.

I understand that both dockerfile are almost the same for golang, but  I'm ok with dealing with a little bit of boilerplates if I can avoid complexity un `run.sh` (and if this boilerplate became too big, we can always push some build logic in a script, as we did for `install_ddtrace.sh`.

cbeauchesne (approved):

PR #79, opened 2021-11-26
PHP: testing appsec/misc fixes in tests

cataphract:
There are two failures that have no details and that I couldn't figure out:

```
============================================ short async test summary info =============================================
 FAILED tests/test_sampling.py::Test_SamplingDecisions::test_sampling_decision_added - 1 fails
 FAILED tests/appsec/test_traces.py::Test_AppSecEventSpanTags::test_custom_span_tags - 1 fails
================================================ 2 failed, 97 passed ================================================
```

Also requires https://github.com/DataDog/dd-appsec-php/releases/download/v0.1.0/dd-appsec-php-0.1.0-amd64.tar.gz in `/binaries`, as the repository is not open yet, so it cannot be downloaded.


cbeauchesne:
For the lint, you can do, with python 3 

```
pip install -r requirements.txt
black .
```

cataphract:
> For the lint, you can do, with python 3
> 
> ```
> pip install -r requirements.txt
> black .
> ```

done

cbeauchesne:
Checking ...

tests/appsec/waf/test_addresses.py
```diff
@@ -11,8 +11,9 @@
     pytestmark = pytest.mark.skip("not relevant")
 
 
-@released(golang="?", dotnet="?", java="?", php="?", python="?", ruby="?")
+@released(golang="?", dotnet="?", java="?", php="0.1.0", python="?", ruby="?")
```
cbeauchesne, line 14: if it's not releqsed at all

```suggestion
@released(golang="?", dotnet="?", java="?", php="?", python="?", ruby="?")
```

tests/appsec/waf/test_addresses.py
```diff
@@ -150,14 +152,16 @@ def test_cookies(self):
     @bug(library="ruby")
     def test_cookies_with_special_chars(self):
         """Other cookies patterns, to be merged once issue are corrected"""
-        r = self.weblog_get("/waf", cookies={"value": ";shutdown--"})
+        # weblog_get doesn't properly escape semicolons in the cookie value,
```
cbeauchesne, line 155: Is it specific for java ?

tests/appsec/waf/test_rules.py
```diff
@@ -242,7 +245,7 @@ def test_sqli(self):
         interfaces.library.assert_waf_attack(r, rules.sql_injection.crs_942_160)
 
         r = self.weblog_get("/waf", cookies={"value": "/*!*/"})
-        interfaces.library.assert_waf_attack(r, rules.sql_injection)
+        interfaces.library.assert_waf_attack(r, rules.sql_injection.crs_942_100)
```
cbeauchesne, line 248:
```suggestion
        interfaces.library.assert_waf_attack(r, rules.sql_injection)
```

Several rules reports this pattern, the WAF is not deterministic here => validate rule type, eg `rules.sql_injection`

utils/_context/core.py
```diff
@@ -63,8 +63,10 @@ def __init__(self):
             self.waf_rule_set = Version("1.0.0")
         elif self.library == "java":
             self.waf_rule_set = Version("0.0.1")
+        elif self.library == "php":
+            self.waf_rule_set = Version("1.0.0")
         else:
-            self.waf_rule_set = Version("0.0.1")
+            self.waf_rule_set = Version("1.0.0")
```
cbeauchesne, line 69: `else` old version :)

```suggestion
            self.waf_rule_set = Version("0.0.1")
```

utils/build/docker/php/install_ddtrace.sh
```diff
@@ -2,25 +2,37 @@
```
cbeauchesne, line 2: The requirement for this script : 

1. if it found some binaries inside /binaries folder then it install it
2. else, it downlaod the very last version of the tracer.
3. then, it extract the version of the tracer with whatever dark magic, and save it in a file named /binaries/SYSTEM_TESTS_LIBRARY_VERSION

From what I see, it starts by step 2, am I right ?

cbeauchesne (changes requested):

utils/interfaces/_logs/core.py
```diff
@@ -154,6 +164,12 @@ def _clean_line(self, line):
 
         return line
 
+    def _get_standardized_level(self, level):
+        if context.library == "php":
+            return level.upper()
+        else:
+            super(_LibraryStdout, self)._get_standardized_level(level)
```
cbeauchesne, line 171: The cause of the error in java 

```suggestion
            return super(_LibraryStdout, self)._get_standardized_level(level)
```

cbeauchesne (changes requested):

tests/appsec/test_reports.py
```diff
@@ -10,7 +10,8 @@
     pytestmark = pytest.mark.skip("not relevant")
 
 
-@released(golang="?", dotnet="1.28.6", nodejs="2.0.0-appsec-beta.1", php="?", python="?", ruby="0.51.0")
+@released(golang="?", dotnet="1.28.6", nodejs="2.0.0-appsec-beta.1", python="?", ruby="0.51.0")
+@missing_feature(library="php", reason="reporting outside traces is not and will not be supported")
```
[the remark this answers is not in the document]
cataphract: I thought it checked both. I'll try to enable it.

tests/appsec/waf/test_addresses.py
cataphract: This test shouldn't work on any agent because it's a limitation of the WAF/rules. I thought this merited an explanation, but I can leave it as a comment instead.

tests/appsec/waf/test_addresses.py
cataphract: I'm doing PHP, not Java, but no, it's not specific. If the cookie header has a semicolon, the http servers will interpret that as starting a new cookie, not as part of the value of the cookie. Arguably, the method `weblog_get` should be changed instead so that int correct encodes the value.

tests/appsec/waf/test_rules.py
cataphract: When I tried it failed if I didn't specify the rule though.

utils/build/docker/php/install_ddtrace.sh
cataphract: I'm installing two things here: the tracer and appsec. What's stored in SYSTEM_TESTS_LIBRARY_VERSION is the version of the tracer, not the version of appsec. It doesn't support installing the latest version though, but that support should go into the installer that I'm using, not the script, as it's planned to include such support in the script. Right now, it installs 0.67.0 of the tracer if there's nothing in /binaries and fails if there's no appsec in /binaries. This is because the appsec repository is not yet public.

cataphract:
I've pushed a new commit. Tests are now passing locally.

tests/appsec/waf/test_rules.py
cbeauchesne: Do you have the log output of this fail ?

tests/appsec/test_traces.py
```diff
@@ -53,8 +53,8 @@ def test_custom_span_tags(self):
         """AppSec should store in APM spans some tags when enabled."""
 
         def validate_custom_span_tags(span):
-            if span.get("name") != "init.service":
-                return
+            if span.get("name") not in ("init.service", "web.request"):
+                return True
```
cbeauchesne, line 57: Returning true here means that it will validate the test if it founds any span with name not in  ("init.service", "web.request"), which is not what we want.

Return None means that the validation is not validated neither invalidated.

Furthermore, I thnink alsmost all span as name = `web.request`, so it's probably not what we want. Can you simply skip the test for now, I'll have a look.

tests/appsec/test_traces.py
```diff
@@ -65,7 +65,7 @@ def validate_custom_span_tags(span):
 
 def validate_appsec_span_tags(span):
     if span.get("parent_id") not in (0, None):  # do nothing if not root span
-        return
+        return True
```
cbeauchesne, line 68: same comment as above for `validate_custom_span_tags`

cbeauchesne (changes requested):

utils/build/docker/php/install_ddtrace.sh
cbeauchesne: > What's stored in SYSTEM_TESTS_LIBRARY_VERSION is the version of the appsec, not the version of the tracer

It must be the tracer version. If you need to use the `appsec` version to do some filters with `irrelevant`, `bug`... we need to add it. Tell me, and I will do it.

----

Ok, from what I understand, you look for a `.tar.gz` in `/binaries`, and it was previously a `.apk` right ? If yes, you also need to modify the `./utils/scripts/load-binary.sh` script => it's used in my nightly CI to get the latest build from your `master` branch.

cataphract: The version of the tracer is mostly irrelevant, since its only role is submitting the trace. For the functionality that we're testing, its version is not nearly as important as the version of the appsec library. So having `SYSTEM_TESTS_LIBRARY_VERSION` be the tracer version may make sense in other runtimes where the two integrated, but no so much in PHP. In any case, if you want `SYSTEM_TESTS_LIBRARY_VERSION` to be the tracer version, then yes, another mechanism will be needed to test against the appsec version.

cataphract:
Tests now pass on CI

cataphract:
@cbeauchesne  Is there anything left for me to do here?

cbeauchesne:
@cataphract so far, I don't think so. On my side, I need to make a change on the versionning system, will be ok today or tomorrow

cbeauchesne:
Ok, I've added a commit that handles the php/AppSec component. In the `released` decorators, `php` refers to the tracers (current version 0.67) and `php_appsec` your component.

Now, several questions : 

1. If there is nothing in binary folder, it's the 0.67 version of the tracers who is installed. `0.67` is hardcoded, is there a way to get the "latest" instead ?
2. If there is nothing in binary folder, the `0.1` version of the php_appsec component is installed => I have the lock issue.
3. `./utils/scripts/load-binary.sh php_appsec` should be working. Basically, it loads the latest build from your `master` branch and put it binary. For instance, for the tracer, I load it from CircleCI.

Note : turns out I may have introduced some mess on the CI, fixing it.

cataphract:
> Ok, I've added a commit that handles the php/AppSec component. In the `released` decorators, `php` refers to the tracers (current version 0.67) and `php_appsec` your component.
> 
> Now, several questions :
> 
> 1. If there is nothing in binary folder, it's the 0.67 version of the tracers who is installed. `0.67` is hardcoded, is there a way to get the "latest" instead ?

With the latest version of the installer, this is possible. I've added a commit that enabled this.

> 2. If there is nothing in binary folder, the `0.1` version of the php_appsec component is installed => I have the lock issue.

This shouldn't happen with the latest version of the install script or with the upcoming 0.2.0 version of ddappsec.

> 3. `./utils/scripts/load-binary.sh php_appsec` should be working. Basically, it loads the latest build from your `master` branch and put it binary. For instance, for the tracer, I load it from CircleCI.

We currently don't publish snapshots. I'll add this to our todo list.

cbeauchesne:
> With the latest version of the installer, this is possible. I've added a commit that enabled thhis

:+1: 

> We currently don't publish snapshots. I'll add this to our todo list.

ok, we can move forward without it, just ping me once it'll be done. Side question : does the repo be open sourced one day ? 

----

Ok, the last step is the support of `DD_APPSEC_WAF_TIMEOUT=5s`. Regarding the current setup, the tests will be too flaky without it :(

cbeauchesne:
@cataphract @Anilm3 merging main into this PR is a little bit a pain, I'm splitting it in two parts : 

* the `php_appsec` mechanism, to be merged this morning
* unskipping test for php_appsec, to be merged once the waf badget can be set threw an env var.

PR #88, opened 2021-12-03
Check request headers instead of response

anna-git:
Fix of checking response instead of request headers


cbeauchesne (commented):
If your fix takes more than a day, I suggest to add a `@bug` on the test :)

anna-git:
It should be good, we already have Robert's PR to fix it, and as its very small, should be approved quickly!

cbeauchesne (approved):

cbeauchesne (approved):

PR #93, opened 2021-12-13
go/net-http: fix root route

Julio-Guerra:
Fix the `net/http` variant of weblog so that:
- An access to `/` returns the status code 200 Ok
- Accesses to anything else such as `/i-dont-exist` returns the status 404 Not Found.

This is due to the way the `net/http.ServeMux` works and matches its pattern: `/` will match anything starting with `/`. This problem doesn't exist the two other variants gorilla and echo whose matching works differently.

### DIFF
--- a/utils/build/docker/golang/app/net-http.go
+++ b/utils/build/docker/golang/app/net-http.go
@@ -15,7 +15,7 @@ func main() {
 	mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
 		// "/" is the default route when the others don't match
 		// cf. documentation at https://pkg.go.dev/net/http#ServeMux
-		// Therefore, we need to check the URL path to only handler the `/` case
+		// Therefore, we need to check the URL path to only handle the `/` case
 		if r.URL.Path != "/" {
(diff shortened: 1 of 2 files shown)


utils/build/docker/golang/app/net-http.go
```diff
@@ -13,6 +13,13 @@ func main() {
 	mux := httptrace.NewServeMux()
 
 	mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		// "/" is the default route when the others don't match
+		// cf. documentation at https://pkg.go.dev/net/http#ServeMux
+		// Therefore, we need to check the URL path to only handler the `/` case
```
Hellzy, line 18: Nitpick: handler -> handle

Hellzy (approved):
LGTM! (Small comment nitpick)

utils/build/docker/golang/app/net-http.go
Julio-Guerra: Fixed

cbeauchesne (approved):

PR #97, opened 2021-12-17
Add rules file from the staging environment

robertpi:


robertpi:
Looks like this approach doesn't work well as the URL requires authentication.

PR #102, opened 2021-12-21
Some data on msgpack may be bytes[] -> decode them

cbeauchesne:


utils/interfaces/_deserializer.py
```diff
@@ -40,23 +40,41 @@ def deserialize_http_message(path, message, data, interface):
...
         return []
 
     return data
 
 
+def _convert_bytes_values(item):
+    if isinstance(item, dict):
+        for key in item:
+            if isinstance(item[key], bytes):
+                item[key] = item[key].decode()
```
lloeki, line 70: Does `decode` assume `ASCII-8BIT` encoding? Any error raised if it isn't?

cbeauchesne: Nope, `utf-8`. For ruby, it's ascii, but it may be not the same for others ?

I don't know what is the <del>best</del> less worst to assume :)

PR #112, opened 2022-01-10
Header tag test has been fixed in .net

pierotibou:


tests/test_library_conf.py
```diff
@@ -8,10 +8,6 @@
 
 
 @irrelevant(library="cpp")
```
cbeauchesne, line 10:
```suggestion
@irrelevant(library="cpp")
@released(dotnet="xx.xx.xx")
```

where `xx.xx.xx` is the good version number (tests are executed on previous versions)

cbeauchesne (changes requested):

cbeauchesne (approved):

PR #125, opened 2022-01-14
Add some discovery scans test (404 rules)

anna-git:
Adds 3 tests to test we get the right attacks reported for discovery attacks


cbeauchesne:
I was working on #137 with @stombre and I realize it's almost the same purpose. Is it ok for you if we move forward with Daix' one ?

PR #128, opened 2022-01-19
Force service name from env for Ruby integrations

lloeki:


cbeauchesne:
Will be fixed in tracer

PR #133, opened 2022-01-21
go: revert "enforce init.service priority to be greater than 0 (#126)"

knusbaum:
Reverts DataDog/system-tests#126

This test was failing for real. We want to re-enable it.


cbeauchesne:
Hi @knusbaum,

Merging it now implies that it will fail on your CI until you fix it, is it ok for you ?

knusbaum:
Yes, this is a real bug and real failure. CI *should* fail until this is fixed.

cbeauchesne (approved):

cbeauchesne:
ok, you can merge it whenever you need

PR #143, opened 2022-01-26
Update to ruby ddtrace 1.0

lloeki:


lloeki:
I'm looking into the sampling failure, it seems there's a change in ddtrace 1.0 we need to adjust to.

cbeauchesne (approved):

PR #147, opened 2022-01-28
Add some properties on schemas check for agent/traces

cbeauchesne:


cbeauchesne:
flaky on java, 80s is not enough

PR #157, opened 2022-02-02
Use ruby 1.0 branch

lloeki:


lloeki:
And here's another one!

FYI @cbeauchesne there should be three more like this in the coming days:

- once I rename the namespace from `Datadog::Security` to `Datadog::AppSec`: use that new namespace instead
- once branch `1.0` is merged into `master`: one that will move dev test from branch `1.0` to `master`
- once `1.0.0.beta1` gem is released: one that will move prod test to use `1.0.0.beta1`

cbeauchesne (approved):

PR #165, opened 2022-02-04
PHP-FPM Weblog Variant

Anilm3:


utils/build/docker/php/apache-mod/entrypoint.sh
```diff
@@ -14,6 +14,26 @@ LOGS_APACHE=(/var/log/apache2/{access.log,error.log})
 touch "${LOGS_APACHE[@]}"
 chown root:adm "${LOGS_APACHE[@]}"
 
+# Fix the timeout
```
Anilm3, line 17: This timeout fix is temporary, until we implement the ability to parse the time + unit value (tracked [here](https://github.com/DataDog/dd-appsec-php/issues/68)).

tests/appsec/waf/test_rules.py
```diff
@@ -3,7 +3,7 @@
 # Copyright 2021 Datadog, Inc.
 
 """Exhaustive tests on WAF default rule set"""
-
+import sys
```
cbeauchesne, line 6:
```suggestion
```

cbeauchesne (changes requested):

tests/appsec/waf/test_rules.py
cbeauchesne: useless if I'm not wrong

cbeauchesne (approved):

PR #184, opened 2022-02-15
Bump prod tracer, point dev to master, and use auto_instrument

lloeki:


lloeki:
Also ends up pointing Ruby dev to master

cbeauchesne (approved):

PR #194, opened 2022-02-25
Add profiling package loading

cbeauchesne:


cbeauchesne:
PHP profiler has been integrated inside php tracer repo

PR #199, opened 2022-03-02
Sync tests with rules release 1.2.6

robertpi:


cbeauchesne:
The issue as now is that we'll break tests for everything else not using `1.2.6`. We need to do something like this for all test that have changed : 

```
    @irrelevant(context.waf_rule_set_version < "1.2.6")
    def test_headers_legacy(self):
        """
        Via server.request.headers.no_cookies
        """
        # Note: we do not check the returned key_path nor rule_id for the alpha version
        r = self.weblog_get("/waf/", headers={"MyHeader": "../../../secret.txt"})
        interfaces.library.assert_waf_attack(r, pattern="../", address="server.request.headers.no_cookies")

    @irrelevant(context.waf_rule_set_version >= "1.2.6")
    def test_headers(self):
        """
        Via server.request.headers.no_cookies
        """
        # Note: we do not check the returned key_path nor rule_id for the alpha version
        r = self.weblog_get("/waf/", headers={"MyHeader": "../../../secret.txt"})
        interfaces.library.assert_waf_attack(r, pattern="/../", address="server.request.headers.no_cookies")
```

cbeauchesne:
FYI : #200

robertpi:
> 
> And as now, we do not have a way to get `waf_rule_set_version` :( cc @netcode

Why would be need a rule version, I'm see pattern is a regex, it could easily be adapted to match the old and new output.

Also, now I realize that it's a regex, it's clearly buggy. As you're using `.` which will match any character, when you actually mean `\.`.

cbeauchesne:
> Also, now I realize that it's a regex, it's clearly buggy. As you're using . which will match any character, when you actually mean ..

Mmmh, are you sure ?

robertpi:
> > Why would be need a rule version
> 
> Because you can have rules deletion.
> 

Do we really care about testing rules that will be deleted in old versions?

cbeauchesne:
Yes, because we constantly test old supported versions of libs.

tests/appsec/waf/test_addresses.py
```diff
@@ -45,8 +45,8 @@ def test_query_encoded(self):
 
     def test_query_with_strict_regex(self):
         """ AppSec catches attacks in URL query value, even with regex containing"""
-        r = self.weblog_get("/waf/", params={"value": "0000012345"})
-        interfaces.library.assert_waf_attack(r, pattern="0000012345", address="server.request.query")
+        r = self.weblog_get("/waf/", params={"value": "sleep(1)"})
+        interfaces.library.assert_waf_attack(r, pattern="sleep(1)", address="server.request.query")
```
Julio-Guerra, line 49: are they really equivalent?
I wonder what "_with_strict_regex" really is about 🤔

Julio-Guerra (approved):

cbeauchesne:
I'll propose another PR that correctly test the 1.2.6 rules without breaking tests on old versions

cbeauchesne:
#202

PR #202, opened 2022-03-02
Adapt test to 1.2.6 version of AppSec rules

cbeauchesne:


tests/appsec/test_alpha.py
```diff
@@ -47,7 +47,8 @@ def test_headers(self):
         """
         # Note: we do not check the returned key_path nor rule_id for the alpha version
         r = self.weblog_get("/waf/", headers={"MyHeader": "../../../secret.txt"})
-        interfaces.library.assert_waf_attack(r, pattern="/../", address="server.request.headers.no_cookies")
+        pattern = "/../" if context.appsec_event_rules < "1.2.6" else "../"
```
Julio-Guerra, line 50: I thinking testing the rule ID would be more relevant, but equivalent for sure.

PR #214, opened 2022-03-04
Adding an endpoint to generate multiple spans

adrianmenendezvalle:
We want to have an endpoint that generates a lot of load in APM to test our services. This new endpoint takes URL parameters to generate a number of spans and tags with random data (to increase the size of the payload)

We're using this in our load tests through https://github.com/DataDog/k9-appsec-testing-env

### DIFF
--- a/tests/test_miscs.py
+++ b/tests/test_miscs.py
@@ -17,5 +17,4 @@ class Test_Basic(BaseTestCase):
 
     def test_spans_generation(self):
         r = self.weblog_get("/spans")
-        assert r.status_code == 200
         interfaces.library.assert_trace_exists(r)
--- a/tests/test_miscs.py
+++ b/tests/test_miscs.py
@@ -1,0 +1,21 @@
+# Unless explicitly stated otherwise all files in this repository are licensed under the the Apache License Version 2.0.
+# This product includes software developed at Datadog (https://www.datadoghq.com/).
+# Copyright 2022 Datadog, Inc.
+
+from utils import BaseTestCase, interfaces, missing_feature
+
+
+@missing_feature(library="cpp")
+@missing_feature(library="golang")
+@missing_feature(library="nodejs")
+@missing_feature(library="java")
+@missing_feature(library="php")
+@missing_feature(library="python")
+@missing_feature(library="ruby")
+class Test_Basic(BaseTestCase):
+    """ Make sure the spans endpoint is successful """
+
+    def test_spans_generation(self):
+        r = self.weblog_get("/spans")
+        assert r.status_code == 200
+        interfaces.library.assert_trace_exists(r)
--- a/utils/build/docker/dotnet/Helper.cs
+++ b/utils/build/docker/dotnet/Helper.cs
@@ -1,0 +1,32 @@
+using Datadog.Trace;
+using System;
+using System.Linq;
+using System.Threading;
+
+
+namespace weblog
+{
+    public class Helper
+    {
+        private static Random random = new Random();
+        private const string chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
+
+        public void GenerateSpan(int garbageTags)
+        {
+            using (var childScope = Tracer.Instance.StartActive("spans.child"))
+            {
+                childScope.Span.ResourceName = "span_" + System.Guid.NewGuid().ToString();
+                for (int i = 0; i < garbageTags; i++)
+                {
+                    childScope.Span.SetTag("garbage" + i, RandomString(50));
+                }
+            }
+        }
+
+        private static string RandomString(int length)
+        {
+            return new string(Enumerable.Repeat(chars, length)
+                .Select(s => s[random.Next(s.Length)]).ToArray());
+        }
+    }
+}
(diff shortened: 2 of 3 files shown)


cbeauchesne:
Hi @adrianmenendezvalle ,

Could you add a simple test on this endpoint? it will ensure all weblogs will have the same endpoints. It should look like this, in a new file `tests/test_miscs.py`: 

``` py 
# Unless explicitly stated otherwise all files in this repository are licensed under the the Apache License Version 2.0.
# This product includes software developed at Datadog (https://www.datadoghq.com/).

from utils import BaseTestCase, missing_feature


if context.library == "cpp":


@missing_feature(library="cpp")
@missing_feature(library="golang")
@missing_feature(library="nodejs")
@missing_feature(library="php")
@missing_feature(library="ruby")
class Test_Basic(BaseTestCase):
    """ Find a good short description :) """

    def test_main(self):
        r = self.weblog_get("/spans")
        assert r.status_code == 200
        interfaces.library.assert_trace_exists(r)
```

We can also add a slightly better test on the trace content if you want.

adrianmenendezvalle:
> Hi @adrianmenendezvalle ,
> 
> Could you add a simple test on this endpoint? it will ensure all weblogs will have the same endpoints. It should look like this, in a new file `tests/test_miscs.py`:
> 
> ```python
> # Unless explicitly stated otherwise all files in this repository are licensed under the the Apache License Version 2.0.
> # This product includes software developed at Datadog (https://www.datadoghq.com/).
> # Copyright 2021 Datadog, Inc.
> 
> from utils import BaseTestCase, missing_feature
> 
> 
> @missing_feature(library="cpp")
> @missing_feature(library="golang")
> @missing_feature(library="nodejs")
> @missing_feature(library="java")
> @missing_feature(library="php")
> @missing_feature(library="python")
> @missing_feature(library="ruby")
> class Test_Basic(BaseTestCase):
>     """ Find a good short description :) """
> 
>     def test_main(self):
>         r = self.weblog_get("/spans")
>         assert r.status_code == 200
>         interfaces.library.assert_trace_exists(r)
> ```
> 
> We can also add a slightly better test on the trace content if you want.

Added! Thanks for the quick review ❤️

cbeauchesne:
My bad, the test is not the good one, let me fix it

adrianmenendezvalle:
> My bad, the test is not the good one, let me fix it

Oh! I tested it locally and it worked fine 🤔

cbeauchesne:
> Oh! I tested it locally and it worked fine 🤔

On .NET yes.

Other one should have not failed

tests/test_miscs.py
```diff
@@ -0,0 +1,21 @@
...
+@missing_feature(library="java")
+@missing_feature(library="php")
+@missing_feature(library="python")
+@missing_feature(library="ruby")
+class Test_Basic(BaseTestCase):
+    """ Make sure the spans endpoint is successful """
+
+    def test_spans_generation(self):
+        r = self.weblog_get("/spans")
+        assert r.status_code == 200
```
cbeauchesne, line 20: Simply remove this line

```suggestion
```

adrianmenendezvalle:
> > Oh! I tested it locally and it worked fine 🤔
> 
> On .NET yes.
> 
> Other one should have not failed

Isn't the `missing_feature` to pick which languages should run the test?

rtfpessoa (approved):

cbeauchesne:
> Isn't the missing_feature to pick which languages should run the test?

Yes it is, but we have a flaw. `missing_feature` do actually run the test. If it fails, it says nothing (expected failure). If it succeed, you'll have a warning, something like `test X was expected to fail, but is a success`. 

The pain point is that system tests framework uses only asynchronous test, meaning that the test function is always ok at execution, but a mechanism validate later the real state of the test function when data are available. And the mechanism handle `missing_feature` meta data.

But even if we use this custom mechanism, it's based on `pytest`, a python good old testing framework. We only use `pytest` for its file crawler and log manager, and don't use at all other testing mechanisms. The pain point is that the line `assert code == 200` is a synchronous test, and is seen by pytest as an error, as it don't know anything about `missing_feature`.

I have a plan to, or get rid of pytest, or adapt our decorator to fit into pytest, but I miss time ...

cbeauchesne (approved):

adrianmenendezvalle:
> > Isn't the missing_feature to pick which languages should run the test?
> 
> Yes it is, but we have a flaw. `missing_feature` do actually run the test. If it fails, it says nothing (expected failure). If it succeed, you'll have a warning, something like `test X was expected to fail, but is a success`.
> 
> The pain point is that system tests framework uses only asynchronous test, meaning that the test function is always ok at execution, but a custom mechanism validate later the real state of the test function when data are available. And this mechanism handles `missing_feature` decorator.
> 
> But even if we use this custom mechanism, it's based on `pytest`, a python good old testing framework. We only use `pytest` for its file crawler and log manager, and don't use at all other testing mechanisms. The pain point is that the line `assert code == 200` is a synchronous test, and is seen by pytest as an error, as it don't know anything about `missing_feature`.
> 
> I have a plan to, or get rid of pytest, or adapt our decorator to fit into pytest, but I miss time ...

Thanks a lot for the explanation and all your help! Makes sense now 🙇

PR #251, opened 2022-03-29
Make tests avoid running on cookies

Taiki-San:
Cookies will be disabled from rules starting with 1.2.7


cbeauchesne:
you can safely ignore fails on node/dev

cbeauchesne (approved):
You'll probably have errors on old version, has query `params` was not supported.

But I'd say if it happens, the simpler will be do another PR with `@missing_feature` for older versions

tests/appsec/waf/test_rules.py
```diff
@@ -41,7 +41,7 @@ class Test_HttpProtocol(BaseTestCase):
     @missing_feature(library="golang", reason="cookie decoding ?")
```
cbeauchesne, line 41: Maybe removing this one ?

cbeauchesne (approved):

PR #263, opened 2022-04-01
Java only got the compatible WAF with 0.98

Taiki-San:


tests/appsec/waf/test_rules.py
```diff
@@ -298,7 +298,7 @@ def test_nosqli_value(self):
         r = self.weblog_get("/waf/", headers={"x-attack": "$nin"})
         interfaces.library.assert_waf_attack(r, rules.nosql_injection.sqr_000_007)
 
-    @missing_feature(context.library != "java", reason="Need to use last WAF version")
+    @missing_feature(context.library != "java", java < "0.96.0", reason="Need to use last WAF version")
```
cbeauchesne, line 301: Without bug, and a little bit more clear

```suggestion
    @missing_feature(context.library != "java", reason="Need to use last WAF version")
    @missing_feature(context.library < "java@0.96.0", reason="Was using a too old WAF version")
```

cbeauchesne (changes requested):

cbeauchesne (approved):

PR #271, opened 2022-04-04
go: fix server.request.cookies tests

Julio-Guerra:
Make two `missing_feature` tests `irrelevant` as we rely on the standard Go parser and it doesn't handle those corner cases properly. We don't plan to do anything about it, even in the long term when we will have dynamic instrumentation.


cbeauchesne:
You also have a `Test_Cookies` to fix in scenario folders. Actually, as cookies has been removed from next rules files, it's this class that used for feature status.

cbeauchesne (approved):

PR #277, opened 2022-04-05
vertx3 app; set max memory on java apps

cataphract:

PR #281, opened 2022-04-05
vertx3 app; set max memory on java apps

cbeauchesne:
Restauration of #277 from @cataphract

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,6 +90,9 @@ jobs:
           - library: java
             weblog-variant: ratpack
             version: prod
+          - library: java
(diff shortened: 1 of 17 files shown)


cbeauchesne:
@cataphract I can't approve my own PR.
If it's ok for you, I merge it

cataphract:
yeah should be good, the only question is if my prs get merged before the next release of the java agent, o/wise there will be some failures on ratpack and vertx3

cbeauchesne:
on your CI, only spring-bbot is used (would be a good idea to include other weblogs though), so it won't break.

If the version number are not good, it'll break my nightly, but without consequence, I'll fix with good version numbers

PR #288, opened 2022-04-07
enable dd_env test for Python

vdeturckheim:


vdeturckheim:
Thanks, I think the test itself fails - I am not yet sure why

cbeauchesne:
So egg-chicken issue, the feature is still on a PR on dd-trace-py

PR #293, opened 2022-04-11
Activate on test for .NET

cbeauchesne:


cbeauchesne:
Should be ok, but is not.

anna-git:
yes, it s because the app is crashing before anything can reach us, cause by default, asp.net mvc json deserializer doesnt accept single quotes.  So we get something like 
`System.Text.Json.JsonReaderException: '''' is an invalid start of a property name. Expected a '"'. LineNumber: 0 | BytePositionInLine: 1.'`

anna-git (changes requested):
r = self.weblog_post("/waf", data={"value": "<vmlframe src=\"xss\">"})

cbeauchesne:
Seems not to be a JSON issue, it's a basic URL form encoded payload.

BTW, I can see that in logs for this request : 

```
DDAS-0010-00: Executing AppSec In-App WAF with parameters: { server.request.method: POST, server.request.uri.raw: http://weblog:7777/waf, server.request.query: {  }, server.request.headers.no_cookies: Error: couldn't format type: System.Collections.Generic.Dictionary`2[System.String,System.String[]], server.request.cookies: {  }, server.response.status: 405 }  { MachineName: ".", Process: "[10 dotnet]", AppDomain: "[1 app]", AssemblyLoadContext: "\"Default\" System.Runtime.Loader.DefaultAssemblyLoadContext #2", TracerVersion: "2.7.0.0" }
```

anna-git:
Regarding the logging error 
`server.request.headers.no_cookies: Error: couldn't format type: System.Collections.Generic.Dictionary`2[System.String,System.String[]],`
this should be fixed in https://github.com/DataDog/dd-trace-dotnet/pull/2676/commits/d7cce1574cccfbf4c1f8ebec3a9af765b76c6702 from [this PR](https://github.com/DataDog/dd-trace-dotnet/pull/2676)

Regarding the error itself, there's no route expecting a POST on waf so the body is never read by the framework, I will adapt the weblog.

anna-git (approved):

PR #311, opened 2022-04-19
Add some docs describing the endpoints a weblog needs

robertpi:


cbeauchesne:
Good initiative @robertpi, thank you! It'll help to implement endpoint. Though, I think that keeping the test the "source of truth" for weblog endpoint is the way to go, otherwise, we'll spend too much energy writing a complete doc, and correcting N weblogs just because we incorrectly understand one endpoint when we wrote the doc. So, I would add in preambule something like : 

> This document describes endpoints implemented on weblog. Though, it's not a complete description, and can contains mistakes. The source of truth are the test itself. If a weblog endpoint passes system tests, then you can consider it as ok. And if it does not passes it, then you must correct it, even if it's in line with this document.

docs/weblog/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+Where `repeats` and `garbage` are the parameters read from the query string.
+
+## GET /sqli
+
+The end point should accept a query string parameter `q`. This parameter should be used as part of a query that is executed against a real database (to ensure a database span is created).
+
+The output of the query should be written to the body of the response.
+
+## GET /waf
```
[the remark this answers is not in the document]
robertpi: I've made the update.

cbeauchesne (approved):

docs/weblog/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+Where `repeats` and `garbage` are the parameters read from the query string.
+
+## GET /sqli
+
+The end point should accept a query string parameter `q`. This parameter should be used as part of a query that is executed against a real database (to ensure a database span is created).
+
+The output of the query should be written to the body of the response.
+
+## \[All HTTP verbs\] /waf
```
lloeki, line 91: it should accept: `/waf` (slashless) and `/waf/*` (e.g `/waf/foo/bar/baz/qux`) as both are in use in system tests

docs/weblog/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Disclaimer
+
+This document describes endpoints implemented on weblog. Though, it's not a complete description, and can contains mistakes. The source of truth are the test itself. If a weblog endpoint passes system tests, then you can consider it as ok. And if it does not passes it, then you must correct it, even if it's in line with this document.
+
+**You are strongly encouraged to help others by submitting corrections when you notice issues with this document.**
+
+# Endpoints
+
+## GET / (aka plain text end point)
```
lloeki, line 13: IIRC it must accept `POST` as well

docs/weblog/README.md
```diff
@@ -0,0 +1,99 @@
...
+Hello world!\n
+```
+
+## GET /headers
+
+This end point should set the following headers:
+
+```
+content-type: text
+content-length: 16
```
lloeki, line 27: watch out for setting `Content-Length` here: it can produce a wait til a timeout occurs as some parts of a framework or user agent may wait for data that never comes. It is often set by the framework anyway, the important part appears to be that it MUST be present, whether manually or via framework facilities.

docs/weblog/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+This document describes endpoints implemented on weblog. Though, it's not a complete description, and can contains mistakes. The source of truth are the test itself. If a weblog endpoint passes system tests, then you can consider it as ok. And if it does not passes it, then you must correct it, even if it's in line with this document.
+
+**You are strongly encouraged to help others by submitting corrections when you notice issues with this document.**
+
+# Endpoints
+
+## GET / (aka plain text end point)
+
+The following text should be written to the body of the response:
```
lloeki, line 15: I don't think the tests check for actual response body, RFC  `SHOULD` or `MAY`?

docs/weblog/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+The end point should accept two query string parameters:
+
+* `repeats` - this is the number of spans that should be manually created
+* `garbage` - this is the number of tags that should added to each a span
+
+The following text should be written to the body of the response:
+
+```
+Generated {repeats} spans with {garbage} garbage tags\n
```
lloeki, line 80: I don't think the tests check for actual response body.

docs/weblog/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+The following text should be written to the body of the response:
+
+```
+Hello world!\n
+```
+
+## GET /spans
+
+The end point should accept two query string parameters:
```
lloeki, line 72:
```suggestion
The end point MUST accept two query string parameters:
```

docs/weblog/README.md
```diff
@@ -0,0 +1,99 @@
...
+content-language: en-US
+```
+
+The following text should be written to the body of the response:
+
+```
+Hello headers!\n
+```
+
+# GET /identify
```
lloeki, line 37: This endpoint MUST respond to the unslashed resource path (`/identify`) and MAY respond to the slashed resource path (`/identify/`)

docs/weblog/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+## GET / (aka plain text end point)
+
+The following text should be written to the body of the response:
+
+```
+Hello world!\n
+```
+
+## GET /headers
```
lloeki, line 21: This endpoint MUST respond to the unslashed resource path (`/headers`) and MAY respond to the slashed resource path (`/headers/`)

docs/weblog/README.md
```diff
@@ -0,0 +1,99 @@
...
+usr.session_id
+usr.role
+usr.scope
+```
+
+The value of each tag should be the tag name, for example `usr.id` should be set to `usr.id`.
+
+## GET /params/%s
+
+This end point should accept a parameter that is a string and is part of the URL path.
```
lloeki, line 54:
```suggestion
This end point MUST accept a parameter that is a string and is part of the URL path.
```

docs/weblog/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+```
+Hello world!\n
+```
+
+## GET /spans
+
+The end point should accept two query string parameters:
+
+* `repeats` - this is the number of spans that should be manually created
```
lloeki, line 74: MUST/SHOULD/MAY they be flat or nested?

lloeki:
I would advise calling the test app `weblog` kept internal/an implementation detail, as it is far removed from the historical weblog used to test Sqreen (which was an actual fake blog app).

docs/weblog/README.md
lloeki: Also, it seems these are optional (a test case hits it without passing some of the args), so:
- the endpoint MUST accept these
- when absent the endpoint MUST behave as if the respective value is a given default (`0`?)

cbeauchesne (approved):

cbeauchesne:
I'e commited suggestion made by @lloeki . Is it ok for you @robertpi ?

robertpi:
For sure!

PR #313, opened 2022-04-19
Fix identify endpoint

lloeki:
The endpoint call ends with `/`, which messess up with Sinatra and Rack somehow (IIRC slahs and no-slash can be two different HTTP resources)


Hellzy (approved):
Approving for Go.
The rest LGTM as well but leaving this up for Anil/Charles to confirm

cbeauchesne (approved):

PR #316, opened 2022-04-21
Robert/activate various tests

robertpi:


robertpi:
Comes off a version too old.

PR #323, opened 2022-04-27
fix query key rule

simon-id:
This PR skips the non-encoded query key test because there is no rule for it yet, and modifies the encoded query key rule as it wasn't matching with any keys_only rule

### DIFF
--- a/tests/appsec/waf/test_addresses.py
+++ b/tests/appsec/waf/test_addresses.py
@@ -12,10 +12,11 @@ if context.library == "cpp":
 
 
 # WAF/current ruleset don't support looking at keys at all
-@released(golang="?", dotnet="?", java="?", nodejs="?", php="?", python="?", ruby="1.0.0.beta1")
+@released(golang="?", dotnet="?", java="?", nodejs="2.6.0", php="?", python="?", ruby="1.0.0.beta1")
 class Test_UrlQueryKey(BaseTestCase):
     """Appsec supports keys on server.request.query"""
 
+    @irrelevant(reason="no non-encoded rule on query yet")
     def test_query_key(self):
         """AppSec catches attacks in URL query key"""
         r = self.weblog_get("/waf/", params={"appscan_fingerprint": "attack"})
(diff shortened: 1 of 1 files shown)


tests/appsec/waf/test_addresses.py
```diff
@@ -12,10 +12,11 @@
 
 
 # WAF/current ruleset don't support looking at keys at all
-@released(golang="?", dotnet="?", java="?", nodejs="?", php="?", python="?", ruby="1.0.0.beta1")
+@released(golang="?", dotnet="?", java="?", nodejs="2.6.0", php="?", python="?", ruby="1.0.0.beta1")
 class Test_UrlQueryKey(BaseTestCase):
     """Appsec supports keys on server.request.query"""
 
+    @irrelevant(reason="no non-encoded rule on query yet")
```
simon-id, line 19: I think the only reason this test was working for ruby is because he makes the querystring a list of kv, and the WAF interprets this as a value, not as a key ;)

simon-id: @lloeki feel free to confirm my hypothesis and update the test accordingly

cbeauchesne: Waiting form confirmation from @lloeki then you can remove the test.

cbeauchesne (changes requested):

tests/appsec/waf/test_addresses.py
```diff
@@ -12,10 +12,11 @@
 
 
 # WAF/current ruleset don't support looking at keys at all
-@released(golang="?", dotnet="?", java="?", nodejs="?", php="?", python="?", ruby="1.0.0.beta1")
+@released(golang="?", dotnet="?", java="?", nodejs="2.6.0", php="?", python="?", ruby="1.0.0.beta1")
```
lloeki, line 15:
```suggestion
@released(golang="?", dotnet="?", java="?", nodejs="2.6.0", php="?", python="?", ruby="?")
```

tests/appsec/waf/test_addresses.py
lloeki: Confirmed, this should not read releasd for ruby above.

tests/appsec/waf/test_addresses.py
```diff
@@ -24,8 +25,8 @@ def test_query_key(self):
     @missing_feature(library="ruby")
```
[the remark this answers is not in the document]
lloeki: correct

lloeki, line 25:
```suggestion

```

lloeki (approved):
Approved, with additional changes batched

lloeki:
Rebased

cbeauchesne (approved):

PR #325, opened 2022-04-28
Active obfuscator tests for .NET

robertpi:


cbeauchesne (approved):

scenarios/appsec/test_traces.py
```diff
@@ -21,16 +21,14 @@ def test_appsec_obfuscator_key(self):
         # Validate that the AppSec events do not contain the following secret value.
         # Note that this value must contain an attack pattern in order to be part of the security event data
         # that is expected to be obfuscated.
-        SECRET = "this is a very secret value having the attack"
+        SECRET = "this-is-a-very-secret-value-having-the-attack"
```
cbeauchesne, line 24: Does it mean that it is not obfuscated when the value contains space ? 🤔

robertpi: No, the cookie would be removed, so not available to app and therefore not analyzed by the waf or sent to the server.

PR #330, opened 2022-04-29
enable more python tests

vdeturckheim:


scenarios/appsec/test_customconf.py
```diff
@@ -29,7 +29,6 @@ class Test_CorruptedRules(_BaseNoAppSec):
 
     @missing_feature(library="golang")
     @missing_feature(library="nodejs")
```
cbeauchesne, line 31: Please add on `released` decorator the python lib's version for this feature (otherwise, it'll break test whem we run them on older versions)

cbeauchesne (changes requested):

PR #334, opened 2022-05-02
Paullgdc/test with latest main branch agent

paullegranddc:
Add the possibility to test against the agent built from the `main` branch instead of the last released version.
To achieve this:
* The datadog-agent is cloned and queried for the last merged commits
* The registry containing images built from the agent's CI is queried to get existing tags, that follow the following format `v<pipeline number>-<12 first char of commit hash>-<agent flavor>`

We pick the agent container with the highest pipeline number, and the latest commit.

The tests are probably only runnable locally, or on gitlab as it requires access to the build-stable aws account.
I only enabled it for php because it's the only language tested on gitlab. Do we want to test with all weblogs in cross-component tests?

Also currently, the main branch has a bug that make the agent ignore proxy settings, which cause system-test to fail.


utils/build/docker/agent.Dockerfile
```diff
@@ -1,21 +1,12 @@
-FROM debian:10
-
-# Install the datadog agent
+ARG AGENT_IMAGE=datadog/agent:7
```
cbeauchesne, line 1: latest ?

utils/build/build.sh
```diff
@@ -82,14 +83,22 @@ do
         docker build -f utils/build/docker/runner.Dockerfile -t system_tests/runner $EXTRA_DOCKER_ARGS .
 
     elif [[ $IMAGE_NAME == agent ]]; then
+        BUILD_ARGS=""
+        if [ -f ./binaries/agent-image ]; then
+            read -r AGENT_BASE_IMAGE < ./binaries/agent-image
```
cbeauchesne, line 88: AGENT_BASE_IMAGE=$(cat ...)

utils/build/build.sh
```diff
@@ -69,6 +69,7 @@ echo ""
 echo "TEST_LIBRARY:      $TEST_LIBRARY"
 echo "WEBLOG_VARIANT:    $WEBLOG_VARIANT"
 echo "BUILD_IMAGES:      $BUILD_IMAGES"
+echo "AGENT_BASE_IMAGE:" $AGENT_BASE_IMAGE
```
cbeauchesne, line 72: Remove this (not needed)

cbeauchesne (changes requested):
Could you have a look on their CircleCi build job ?

paullegranddc:
🔥

PR #343, opened 2022-05-06
Disable sampling test on Ruby

lloeki:


tests/test_sampling.py
```diff
@@ -31,6 +31,7 @@ def next_request_id(cls):
     @bug(context.library < "java@0.92.0")
     @flaky(context.library < "python@0.57.0")
     @flaky(context.library >= "java@0.98.0", reason="APMJAVA-743")
+    @flaky(library="ruby", reason="fails randomly for Sinatra on JSON body that dutifully keeps")
```
cbeauchesne, line 34:
```suggestion
    @flaky(library="ruby", weblog_variant="sinatra", reason="fails randomly for Sinatra on JSON body that dutifully keeps")
```

cbeauchesne (changes requested):

tests/test_sampling.py
cbeauchesne: if it fails only on sinatra :)

lloeki: Given it appears something is unfit in the test and there is no significant behaviour difference between Rails, Sinatra, and Rack except timing, I was keen on disabling the test altogether until it is not flaky anymore.

lloeki: Applied anyway.

cbeauchesne (approved):

lloeki:
Thanks!

PR #354, opened 2022-05-13
Don't used DD_TRACE_SAMPLE_RATE in the rate limit test

robertpi:
…rate limiter overrides what appsec does


cbeauchesne:
Duplicated with #355

PR #356, opened 2022-05-13
Add test covering the agent <-> tracer update script test

ameske:


cbeauchesne:
Fill free to push commit directly to `cbeauchesne/test-remote-configuration`, it'll be easier

ameske:
Wanted to get your thoughts on the approach first as I'm worried it might not line up with the project's design. I haven't run anything yet, but I think I have enough code down to share with you the structure and get your thoughts.

I have a program that, based on a defined script, generates Agent -> Tracer JSON payloads and at the same time generates the important expected fields that we want to check in the next Tracer -> Agent request. My thought was to just store those in the repo, keep track of what request we're on, and validate against the corresponding pre-generated JSON payload. That keeps us from having to have all the validation logic for each request be hard coded and makes it easier to add more tests by just adding more JSON payloads.

ameske:
Any thoughts on the high-level approach of using pre-calculated payloads that we load and compare requests to? This of course requires us to be able to pre-load responses impersonating the agent in the proxy.

I've tested out this patten using my own test server that plays the role that system-tests would and a WIP Go tracer client. It works really well for us, so hopefully this design is suitable for system-tests.

ameske:
I think so?

The remote config client in the tracer is started when a tracer starts up, and polls for updates on a periodic interval. So I'd need the command to let me say "respond first with X, then Y, then Z" so I can queue all the responses up.

cbeauchesne:
FYI @ameske : #366 (sorry for the delay).

Let me merge this PR, rebase, and we'll be able to start something.

ameske:
No worries at all! :-) Sounds like a plan.

tests/test_remote_configuration.py
```diff
@@ -34,3 +35,57 @@ def validator(data):
...
+
+    def test_tracer_update_sequence(self):
+        def load_expected_payload(test_no):
+            path = "???/r{}.json".format(test_no)
+            f = open(path, "r")
+            payload = json.load(f)
+            f.close()
+            return payload
+
+        def validator(data):
```
ameske, line 55: I think I'm having trouble understanding how this all connects together with #366.

It sounds like I need to do something like:

```python
from utils import proxies

proxies.set("mock_remote_config_backend", True)
```

But then I'm not sure how I go about queuing up the responses. I see that there is a modify_response function within the proxy forwarder where the flag is checked and the response is modified if True, but I don't see a way to set this here in the test definition. https://github.com/DataDog/system-tests/blob/8988a10f4fb4db35ae12472351e98933a513b787/utils/proxy/forwarder.py#L27-L35

I also don't understand how to control the timing of all of this. As soon as the weblog application that embeds the tracer starts up, the remote configuration client within it will start making a request behind the scenes for an update every X seconds (X is based on how we configure it). It's my understanding that `validator` defined here is called for every request to the /v0.7/config endpoint that the proxy captures. I want the first 13 requests from the tracer to be met with 13 different responses in sequence. So for this style test to work I'd need the proxy preconfigured before the tracer starts its request polling.

Let me know if it'd be better to have a quick sync up. I currently have a Go program that runs the tracer that mimics this interception of requests that I've used in the interim, so I can show a concrete example of what we'd need.

cbeauchesne: let's meet :)

utils/build/docker/golang/app/net-http.go
```diff
@@ -8,8 +8,13 @@ import (
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
 )
 
+var tufRoot = `{"signed":{"_type":"root","spec_version":"1.0","version":1,"expires":"2032-05-29T12:49:41.030418-04:00","keys":{"ed7672c9a24abda78872ee32ee71c7cb1d5235e8db4ecbf1ca28b9c50eb75d9e":{"keytype":"ed25519","scheme":"ed25519","keyid_hash_algorithms":["sha256","sha512"],"keyval":{"public":"7d3102e39abe71044d207550bda239c71380d013ec5a115f79f51622630054e6"}}},"roles":{"root":{"keyids":["ed7672c9a24abda78872ee32ee71c7cb1d5235e8db4ecbf1ca28b9c50eb75d9e"],"threshold":1},"snapshot":{"keyids":["ed7672c9a24abda78872ee32ee71c7cb1d5235e8db4ecbf1ca28b9c50eb75d9e"],"threshold":1},"targets":{"keyids":["ed7672c9a24abda78872ee32ee71c7cb1d5235e8db4ecbf1ca28b9c50eb75d9e"],"threshold":1},"timestsmp":{"keyids":["ed7672c9a24abda78872ee32ee71c7cb1d5235e8db4ecbf1ca28b9c50eb75d9e"],"threshold":1}},"consistent_snapshot":true},"signatures":[{"keyid":"ed7672c9a24abda78872ee32ee71c7cb1d5235e8db4ecbf1ca28b9c50eb75d9e","sig":"d7e24828d1d3104e48911860a13dd6ad3f4f96d45a9ea28c4a0f04dbd3ca6c205ed406523c6c4cacfb7ebba68f7e122e42746d1c1a83ffa89c8bccb6f7af5e06"}]}`
+
 func main() {
-	tracer.Start()
+	tracer.Start(
+		tracer.WithServiceVersion("1.2.3"),
```
ameske, line 15: We'll need the tracers to explicitly set this across all possible "flavors". I only did net/http for Golang at the moment in case there's a better way to set this that you know of.

utils/build/docker/golang/app/net-http.go
```diff
@@ -8,8 +8,13 @@ import (
...
+		tracer.WithRemoteConfigTUFRoot(tufRoot),
```
ameske, line 16: For the Go tracers we'll need a way to configure the TUF root file as well. I imagine for production we'll embed this, but testing assets use a testing root, so the ability to override this is important.

Eventually other tracers will need this too, but we're only doing TUF validations in Go for now. I only added this to net-http as an example, but I can do so for others. The other thing is we're still working with other teams on the Go tracer, so this is a very experimental version that's not finalized at all. How we configure this root will change eventually.

utils/proxy/forwarder.py
```diff
@@ -28,7 +28,7 @@ def __init__(self):
 
         # This is the proxy state. Basically, true or false values that tells the proxy to enable, or not
         # specific behavior. You can get/modify it using a direct GET/POST /_system_tests_state request to the proxy
-        self.state = json.loads(os.environ.get("INITIAL_PROXY_STATE", "") or "{}")
+        self.state = json.loads(os.environ.get("INITIAL_PROXY_STATE", "") or '{"mock_remote_config_backend": true}')
```
ameske, line 31: I ended up with timing problems if I tried to set the state within the test class. I need the first RC update request from the tracer on the /v0.7/config endpoint to be given a mocked response. This worked for now but if you have a better less hacky way to do so please let me know. 😄

utils/build/docker/golang/app/net-http.go
```diff
@@ -8,8 +8,13 @@ import (
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
 )
 
+var tufRoot = `{"signed":{"_type":"root","spec_version":"1.0","version":1,"expires":"2032-05-29T12:49:41.030418-04:00","keys":{"ed7672c9a24abda78872ee32ee71c7cb1d5235e8db4ecbf1ca28b9c50eb75d9e":{"keytype":"ed25519","scheme":"ed25519","keyid_hash_algorithms":["sha256","sha512"],"keyval":{"public":"7d3102e39abe71044d207550bda239c71380d013ec5a115f79f51622630054e6"}}},"roles":{"root":{"keyids":["ed7672c9a24abda78872ee32ee71c7cb1d5235e8db4ecbf1ca28b9c50eb75d9e"],"threshold":1},"snapshot":{"keyids":["ed7672c9a24abda78872ee32ee71c7cb1d5235e8db4ecbf1ca28b9c50eb75d9e"],"threshold":1},"targets":{"keyids":["ed7672c9a24abda78872ee32ee71c7cb1d5235e8db4ecbf1ca28b9c50eb75d9e"],"threshold":1},"timestsmp":{"keyids":["ed7672c9a24abda78872ee32ee71c7cb1d5235e8db4ecbf1ca28b9c50eb75d9e"],"threshold":1}},"consistent_snapshot":true},"signatures":[{"keyid":"ed7672c9a24abda78872ee32ee71c7cb1d5235e8db4ecbf1ca28b9c50eb75d9e","sig":"d7e24828d1d3104e48911860a13dd6ad3f4f96d45a9ea28c4a0f04dbd3ca6c205ed406523c6c4cacfb7ebba68f7e122e42746d1c1a83ffa89c8bccb6f7af5e06"}]}`
```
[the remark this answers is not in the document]
ameske: Looking at adding support for doing this in an env variable, should be easy to do and would keep the weblog clean and backwards compatible.

scenarios/remote_config/rc_tuf_root.json
```diff
@@ -0,0 +1 @@
+{"signed":{"_type":"root","spec_version":"1.0","version":1,"expires":"2032-05-29T12:49:41.030418-04:00","keys":{"ed7672c9a24abda78872ee32ee71c7cb1d5235e8db4ecbf1ca28b9c50eb75d9e":{"keytype":"ed25519","scheme":"ed25519","keyid_hash_algorithms":["sha256","sha512"],"keyval":{"public":"7d3102e39abe71044d207550bda239c71380d013ec5a115f79f51622630054e6"}}},"roles":{"root":{"keyids":["ed7672c9a24abda78872ee32ee71c7cb1d5235e8db4ecbf1ca28b9c50eb75d9e"],"threshold":1},"snapshot":{"keyids":["ed7672c9a24abda78872ee32ee71c7cb1d5235e8db4ecbf1ca28b9c50eb75d9e"],"threshold":1},"targets":{"keyids":["ed7672c9a24abda78872ee32ee71c7cb1d5235e8db4ecbf1ca28b9c50eb75d9e"],"threshold":1},"timestsmp":{"keyids":["ed7672c9a24abda78872ee32ee71c7cb1d5235e8db4ecbf1ca28b9c50eb75d9e"],"threshold":1}},"consistent_snapshot":true},"signatures":[{"keyid":"ed7672c9a24abda78872ee32ee71c7cb1d5235e8db4ecbf1ca28b9c50eb75d9e","sig":"d7e24828d1d3104e48911860a13dd6ad3f4f96d45a9ea28c4a0f04dbd3ca6c205ed406523c6c4cacfb7ebba68f7e122e42746d1c1a83ffa89c8bccb6f7af5e06"}]}
```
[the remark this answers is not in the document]
ameske: Ah, yes it is indeed. I was originally going to store it as a file rather than straight in the dockerfile but I couldn't figure out a way to set the env variable nicely with the data stored in the file. So I just put the contents directly in the dockerfile.

cbeauchesne (approved):

PR #358, opened 2022-05-20
Tests for standard tags required by ASM

robertpi:


scenarios/test_library_conf.py
```diff
@@ -0,0 +1,41 @@
...
+from utils.tools import logger
+
+# PHP JAVA ok
+
+
+@irrelevant(library="cpp")
+class Test_HeaderTags(BaseTestCase):
+    """DD_TRACE_HEADER_TAGS env var support"""
+
+    @released(dotnet="2.1.0", golang="?", java="?", nodejs="?", php_appsec="?", python="?", ruby="?")
```
cbeauchesne, line 15: `released` decorator is only valid on class. Basically: 

* one class <=> on feature
* N method => N use cases

scenarios/test_library_conf.py
```diff
@@ -0,0 +1,41 @@
...
+# Copyright 2021 Datadog, Inc.
+
+from utils import context, BaseTestCase, interfaces, released, bug, irrelevant, missing_feature
+from utils.tools import logger
+
+# PHP JAVA ok
+
+
+@irrelevant(library="cpp")
+class Test_HeaderTags(BaseTestCase):
```
cbeauchesne, line 12: Could you add `coverage` decorator ? Reminder : 

* `@coverage.basic` => most simple test case, it ensures at least that the main use case of the feature is working
* `@coverage.complete` => You are confortable to says that a bug can't happen if the test is sucessful (at least for contexts covered by your weblogs)
* `@coverage.good` => everything else

tests/test_standard_tags.py
```diff
@@ -0,0 +1,93 @@
+# Unless explicitly stated otherwise all files in this repository are licensed under the the Apache License Version 2.0.
+# This product includes software developed at Datadog (https://www.datadoghq.com/).
+# Copyright 2022 Datadog, Inc.
+
+from utils import context, BaseTestCase, interfaces, irrelevant, released
+
+
+class Test_StandardTags(BaseTestCase):
```
cbeauchesne, line 8: Same remark : move `released` here and add `coverage`

scenarios/test_library_conf.py
cbeauchesne, line 12: Could you add `coverage` decorator ? Reminder : 

* `@coverage.basic` => most simple test case, it ensures at least that the main use case of the feature is working
* `@coverage.complete` => You are confortable to says that a bug can't happen if the test is sucessful (at least for contexts covered by your weblogs)
* `@coverage.good` => everything else

utils/interfaces/_library/miscs.py
```diff
@@ -121,6 +121,46 @@ def check(self, data):
...
+            for span in trace:
+                if self.rid:
+                    if self.rid != _get_rid_from_span(span):
+                        continue
+
+                    self.log_debug(f"Found a trace for {m(self.message)}")
+
+                try:
+                    for tagKey in self.tags:
+                        if tagKey not in self.tags:
```
cbeauchesne, line 149: maybe that instead ? 

```suggestion
                        if tagKey not in span["meta"]:
```

utils/interfaces/_library/miscs.py
```diff
@@ -121,6 +121,46 @@ def check(self, data):
...
+
+                        if expectValue != actualValue:
+                            raise Exception(
+                                f'{tagKey} tag in span\'s meta should be "{expectValue}", not "{actualValue}"'
+                            )
+
+                    self.log_debug(f"Trace in {data['log_filename']} validates {m(self.message)}")
+                    self.is_success_on_expiry = True
+                except Exception as e:
+                    self.set_failure(f"{m(self.message)} not validated: {e}\nSpan is: {span}")
```
cbeauchesne, line 163: Giving the file name helps a lot : 

```suggestion
                    self.set_failure(f"{m(self.message)} not validated in {data['log_filename']}:\n{e}\nSpan is: {span}")
```

cbeauchesne (changes requested):

cbeauchesne:
For the linter, here is an how to : https://github.com/DataDog/system-tests/blob/main/docs/edit/lint.md

cbeauchesne:
I may know why it's failing. We put a watermark in this headers, to be able to correlate the request and the trace . But as there is no appsec event, the user-agent header is not reported in the trace. Give me some time, I'll have a look.

cbeauchesne:
I think you need to add `user-agent:http.request.headers.user-agent` in your custom `DD_TRACE_HEADER_TAGS`, it will solve issues

cbeauchesne:
Don't worry about failures on sinatra 14, 20 and 21, it's not related and being corrected soon by Loic.

robertpi:
@cbeauchesne Seems, everything's passing expect the sinatra tests. I'm out Monday, but could we get this merged anyway?

cbeauchesne (approved):

PR #368, opened 2022-06-02
Tests for url redaction

robertpi:


cbeauchesne (changes requested):
could you just add `@rfc("...")` with the rfc link at the top of the class ?  :pray:

simon-id:
@robertpi @cbeauchesne I added the rfc decorator as it looks like Robert is busy these days

cbeauchesne (commented):
just wait for the CI before merging

simon-id:
I think the test cases are wrong @robertpi when i try on regex101 it doesn't have the expected outcome

cbeauchesne:
@simon-id I think they are not regex

simon-id:
@robertpi https://regex101.com/r/LsLd9L/1

robertpi:
> @robertpi https://regex101.com/r/LsLd9L/1

Yes, I misunderstood the standard regex when writing the test, corrected now.

PR #384, opened 2022-06-20
Handle several version of .NET tracer

cbeauchesne:
extract ddtrace_version from the loaded binary with regex
read env variable DDTRACE_VERSION from msbuild 
define constants dependin on the tracer version and different partial class executed depending on it


cbeauchesne:
FYI, testing with `2.2.0` : 

```
/app/app.csproj(9,17): error MSB4086: A numeric comparison was attempted on "$(DDTRACE_VERSION)" that evaluates to "" instead of a number, in condition "'$(DDTRACE_VERSION)' != '' OR '$(DDTRACE_VERSION)' >= '2.7.0'".
```

PR #385, opened 2022-06-20
Verify telemetry proxy correctly enriches proxied requests

pawelchcki:


utils/interfaces/_library/telemetry.py
```diff
@@ -18,6 +18,26 @@ def check(self, data):
         if not 200 <= repsonse_code < 300:
             self.set_failure(f"Got response code {repsonse_code} telemetry message {data['log_filename']}")
 
+class _TelemetryProxyEnrichmentValidation(BaseValidation):
+    is_success_on_expiry = True # TODO: why this needs to be set?
```
cbeauchesne, line 22: There is two validation logic : 

* I need to check at least one value. Once I saw it, it's validated => `is_success_on_expiry` is set to false because if I see nothing, it's a failure
* I need to check all values. Having one positive check is not enough. So you never set a success by yourself, you just tell that if at the end of the session, if you don't explicitly set it as a failure, it will be a success  => `is_success_on_expiry` is true

cbeauchesne (approved):

PR #401, opened 2022-07-06
go: @irrelevant test_actor_ip

Hellzy:
Mark the test as irrelevant since the Go library now relies on the backend copying what's in `http.client_ip` into `actor.ip`


Hellzy:
Will set as ready for review once tests pass

cbeauchesne (approved):

PR #402, opened 2022-07-07
Test all supported php versions

estringana:
Right now PHP is only tested with version 8.0. This PR is adding tests for all supported versions.

The way it is done is by extracting as much as possible from the exiting dockerfile to a bash script which will be reused by all the different variants.

The naming strategy for variants also change. Now the name of the variant will contain the target php version. Currently `apache-mod`, now `apache-mod-8.0`, `apache-mod-7.4`, etc


estringana:
Hi @cbeauchesne, I forgot to add the PR is still in WIP. How do you tag WIP PRs? I can see there is no tag on this repo. Do you use the description for doing so?

utils/build/docker/php/apache-mod-7.0.Dockerfile
```diff
@@ -0,0 +1,25 @@
+ARG PHP_VERSION=7.0
+ARG VARIANT=release-zts
+
+FROM datadog/dd-appsec-php-ci:php-$PHP_VERSION-$VARIANT
+
+ENV DD_TRACE_ENABLED=1
+ENV DD_TRACE_GENERATE_ROOT_SPAN=1
+ENV DD_TRACE_AGENT_FLUSH_AFTER_N_REQUESTS=0
+ENV DD_TRACE_HEADER_TAGS=user-agent
```
estringana, line 9: **HELP WANTED**

I am trying to remove as much duplication as possible from all the dockerfiles. These 4 lines are duplicated in all dockerfiles and I would like to move them to `utils/build/docker/php/apache-mod/build.sh`. I have tried to add them to the `.zshrc` file and also to add a SH file into the `profile.d` folder but none of them worked.

estringana: In fact, all dockerfiles are 99% the same. The only thing which changes is the first line where the PHP version is specified. Ideally, calling a generic dockerfile with a build argument which specifies the version would be good to avoid duplication. That said, current duplication is not the end of the world. Any idea is welcome

cbeauchesne: Hi,

You can't set a env var by calling an `.sh` in a dockerfile.

Though, common env var can be set in `utils/build/docker/set-system-tests-weblog-env.Dockerfile`, **if and only if** they are implemented the sane in all tracers. It's not the case for instance for DD_TRACE_HEADER_TAGS.

That being said, we can imagine modifying `utils/build/build.sh ` to have a common dockerfile for each tracer : 

on line 119 , somthing like 

```
IF utils/build/docker/${TEST_LIBRARY}/set-common.Dockerfile EXISTS THEN
        docker build \
            --progress=plain \
            -f utils/build/docker/${TEST_LIBRARY}/set-common.Dockerfile \
            -t system_tests/weblog \
            $EXTRA_DOCKER_ARGS \
            .
fi
```

And you set in utils/build/docker/${TEST_LIBRARY}/set-common.Dockerfile eveything common across all your weblog

estringana: Thanks for info about the env vars!

Regarding the common dockerfile, how would it get the version of php that it needs to build?

cbeauchesne: it doesn't need to have it. Or did I missed something ?

estringana: Just to keep this PR up to date. We had a call and spoke about this. The agreed is that there will be have a kinda dockerfile extending weblog for common things within a language. That will avoid duplication. This will be done in a different PR when FPM gets done(so we can provide a robust solution which works with both)

cbeauchesne:
Well, deserialization is working, trying to ignore UTF-8 errors. But as the value are not the good one, some other tests are failing. I'm digging into those errors

estringana:
Thanks @cbeauchesne for all your support 🙏

cbeauchesne:
mmmh, I think I made a mistake

estringana:
@Anilm3 @cbeauchesne PR is ready to review :)

cbeauchesne (approved):
Great ! 

Can you just add the failing weblogs in the CI, and comment the lines ?

PR #403, opened 2022-07-07
feat: :white_check_mark: Enable collected response headers for Python 1.2.3

gnufede:
Enable collected response headers test for Python


gnufede:
Oh, OK, so we should ship `1.2.3` prior to enabling this test, right? I don't know why it is fetching `1.2.0` to be honest

cbeauchesne:
ouch, I think I know why, i'm getting `1.1` instead of `1.x` branch, let me fix this

cbeauchesne:
Well, even with `1.x` : 

```bash
> pip install git+https://github.com/DataDog/dd-trace-py.git
...
> python -c "import ddtrace; print(ddtrace.__version__)"
1.2.0rc2.dev58+g06e7207f
```

tests/appsec/test_traces.py
```diff
@@ -247,7 +247,7 @@ def validate_appsec_span_tags(span, appsec_data):
 
 
 @rfc("https://datadoghq.atlassian.net/wiki/spaces/APS/pages/2186870984/HTTP+header+collection")
-@released(dotnet="2.5.1", php_appsec="0.2.2", python="?", ruby="1.0.0.beta1")
+@released(dotnet="2.5.1", php_appsec="0.2.2", python="1.2.3", ruby="1.0.0.beta1")
```
brettlangdon, line 250:
```suggestion
@released(dotnet="2.5.1", php_appsec="0.2.2", python="1.3.0", ruby="1.0.0.beta1")
```

The release process for `1.3.0` has already been started, so right now we have `1.3.0rc2` released so any scm version up to date with `1.x` should be above `1.3.0rc2`.


This might get tricky since we won't know what version your change will get released in until it is merged into `1.x`

cbeauchesne (approved):

PR #407, opened 2022-07-08
Split remote config update sequence tests up by supported product

ameske:
The initial approach to adding remote configuration support to system
tests involve a pregenerated "script" of mocked responses that we
delivered to the tracers whenever a tracer requested an update. The
responses tested a variety of scenarios described in the RFC. These
mocked responses involved configuration files for two products supported
by remote configuration (ASM, and FEATURES).

Unfortunately, tracers won't all support the same set of products. This
wouldn't be an issue if tracers stored config files as raw bytes
internally, but part of the RFC mandates that clients deserialize the
config and reject any malformed config files for a given product. As a
result, if a tracer client is not aware of a given product, it's not
going to know how to deserialize its config and it'll fail the test.

To solve this, we consider support for a remote configuration product a
separate "feature" in system tests. We now create a test script for each
product allowing a tracer to opt in once it supports the given product.


ameske:
@cbeauchesne -  I think I might have a bug in the mocked responses file that I'm working on...but I wanted you to be able to take a look at my high level approach here while I'm working on that. I have a feeling there's a more elegant way to do it 😄 . Nothing else will change on my end except I'll likely have an updated `mocked` JSON file.

cbeauchesne (approved):

PR #408, opened 2022-07-11
Test all supported php versions fpm

estringana:
Until now, on the pipeline the only version tested was PHP 8.0. This PR add tests for all supported version of PHP


utils/build/docker/php/php-fpm/entrypoint.sh
```diff
@@ -15,11 +15,11 @@ touch "${LOGS_APACHE[@]}"
 chown root:adm "${LOGS_APACHE[@]}"
 
 # Unused at the moment
-env | sed -rn 's#^([^=]+)=([^=]+)$#env[\1] = "\2"#p' | tee /dev/stderr >> /etc/php/8.0/fpm/pool.d/www.conf
-sed -i "s/;clear_env = no/clear_env = no/" /etc/php/8.0/fpm/pool.d/www.conf
+env | sed -rn 's#^([^=]+)=([^=]+)$#env[\1] = "\2"#p' | tee /dev/stderr >> /etc/php/PHP_VERSION/fpm/pool.d/www.conf
```
cbeauchesne, line 18: how this is working Oo it should be $PHP_VERSION, or did I miss something ?

estringana: This file is kind a template. On the building I´m sedding it and changing `PHP_VERSION` for the built PHP version. It is done on here https://github.com/DataDog/system-tests/pull/408/files#diff-cb8b3f596af52d706c2371c13b60fe4cce0e410a35be4b10ed56b96584511d15R42

cbeauchesne: ah ok !

cbeauchesne (approved):

PR #411, opened 2022-07-12
Local security tool complained about world exposed ports on my laptop.

pawelchcki:
Lets see if we can just expose them locally only


paullgdc:
Not sure that does anything. Any port exposed to the docker host can be accessed by someone on the same lan btw https://gist.github.com/guns/1dc1742dce690eb560a3a2d7581a9632

cbeauchesne:
regarding @paullgdc remark, does the  bot stop complaining ?

pawelchcki:
@cbeauchesne Yeah the tools seems happy. 

Re: Paul's comment, I should probably remove docker from my Private Router :D

cbeauchesne (approved):

colin-higgins (approved):

PR #424, opened 2022-07-21
remote config: Create a cached and a no cached version of the scenarios

ameske:
Whether or not a tracer caches config files is an opt in feature. If a
tracer does opt-in to caching the script needs to change to only send
new or modified config files to the tracer so that the tracer is
properly challenged to detect and verify unmodified files by using its
cache. Since everything is static as far as the scripts go, this means
we need a CACHE and a NOCACHE scenario for each product.


run.sh
```diff
@@ -143,6 +143,21 @@ elif [ $SYSTEMTESTS_SCENARIO = "REMOTE_CONFIG_MOCKED_BACKEND_ASM_DD" ]; then
     export SYSTEMTESTS_LOG_FOLDER=logs_remote_config_mocked_backend
     export SYSTEMTESTS_LIBRARY_PROXY_STATE='{"mock_remote_config_backend": "ASM_DD"}'
 
+elif [ $SYSTEMTESTS_SCENARIO = "REMOTE_CONFIG_MOCKED_BACKEND_FEATURES_NOCACHE" ]; then
+    export RUNNER_ARGS="scenarios/remote_config/test_remote_configuration.py::Test_RemoteConfigurationFields scenarios/remote_config/test_remote_configuration.py::Test_RemoteConfigurationUpdateSequenceFeaturesNoCache"
+    export SYSTEMTESTS_LOG_FOLDER=logs_remote_config_mocked_backend
```
cbeauchesne, line 148: can you give a different name for log folder for each scenario (we'll later collect logs artifact)

cbeauchesne (changes requested):

cbeauchesne (approved):

PR #438, opened 2022-08-02
System tests fixes

gnufede:


juanjux (approved):

tests/appsec/test_traces.py
```diff
@@ -128,18 +127,22 @@ def validate_response_headers(span):
         interfaces.library.add_span_validation(r, validate_response_headers)
 
     @bug(context.library < "java@0.93.0")
-    @irrelevant(library="php", reason="Trace outside a context of a web request is not possible on PHP")
```
gnufede, line 131: N.B: The test is now changed and should work for PHP as well, hence this decorator is not needed anymore

avara1986 (approved):

PR #441, opened 2022-08-03
go: enable user id propagation test

Hellzy:
Enable user id propagation test for Golang and add new endpoints to weblog apps.


pawelchcki (dismissed):
Looks like some Go tests on prod, are failing, probably not related to this change, but worth taking a look before merging in new go Code.

cbeauchesne:
@Hellzy seems like we need to find a way to do optional compilation, we faced this issue by the past, am I right ?

Hellzy:
Hi sorry about the delay I was out for a bit. Yes, conditional compilation for the assume-moving issue will be done as of 1.41.0

cbeauchesne (approved):

PR #450, opened 2022-08-10
Fix failing golang tests

pawelchcki:


cbeauchesne:
That's the good fix, I'm doing it properly on #456

PR #455, opened 2022-08-15
Adjust tests for live debugger dotnet

shurivich:
Summary of changes:

- Enable Live Debugger for tests
- Change language to `dotnet`
- Increase duration for `dotnet` tests to 80 seconds


utils/interfaces/_library/core.py
```diff
@@ -51,6 +51,8 @@ def __init__(self):
             self.expected_timeout = 5
         elif context.library.library in ("php",):
             self.expected_timeout = 10  # possibly something weird on obfuscator, let increase the delay for now
+        elif context.library.library in ("dotnet",):
+            self.expected_timeout = 80
```
OmerRaviv, line 55: @ameske It seems to me like the timeout for other languages would also need to be increased in order for other languages to pass the "Update Sequence Test", unless I'm missing something? AFAICT there are 13 steps in the update sequence, and the default polling interval which we have to wait between each step is 5 seconds, so the very minimum for the test to pass would have to be 5*13=65 seconds?

ameske: 5s is the maximum interval you have to wait between steps, but it is allowed to be configured lower. Regardless though, even at 1s some of these look too short. I think it's better to be able to configure the refresh interval to 1s or smaller for the tests though, as this might have an impact on tests that don't involve remote config and force them to wait when they don't need to.

OmerRaviv: Yeps, that was actually the first solution we did, we introduced a new environment variable for `DD_INTERNAL_RCM_POLL_INTERVAL` and set it to 1000ms in the System Tests. We didn't go with that as it would require the other tracers to follow suite, but maybe we can just collectively decide that's what we're doing.  /cc @jpbempel @p403n1x87

cbeauchesne (approved):

PR #458, opened 2022-08-16
feat(python/asm): Enable http.useragent tag tests

avara1986:
Enable http.useragent tag tests


juanjux (approved):

gnufede (approved):

cbeauchesne:
The python tracer form the github repo is version `1.4.0`, so the test is not executed. We must wait for it before merging.

PR #470, opened 2022-09-02
Display error on Validation exception at __init__

cbeauchesne:
cc @robertomonteromiguel

### DIFF
--- a/utils/interfaces/_core.py
+++ b/utils/interfaces/_core.py
@@ -126,6 +126,10 @@ class InterfaceValidator(object):
         if self.system_test_error is not None:
             return
 
+        if validation.system_test_error is not None:
+            self.system_test_error = validation.system_test_error
+            return
+
         validation._interface = self.name
 
         logger.debug(f"{repr(validation)} added in {self}[{len(self._validations)}]")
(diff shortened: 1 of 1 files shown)


utils/interfaces/_core.py
```diff
@@ -196,71 +200,75 @@ class BaseValidation(object):
...
-            self.rid = None
+            if self.path_filters is not None:
+                self.path_filters = [re.compile(path) for path in self.path_filters]
 
-        self.frame = None
-        self.calling_method = None
-        self.is_xfail = False
+            if request is not None:
+                user_agent = [v for k, v in request.request.headers.items() if k.lower() == "user-agent"][0]
+                self.rid = user_agent[-36:]
```
robertomonteromiguel, line 223: what is the meaning of -36?

robertomonteromiguel, line 223: what is the meaning of -36?

robertomonteromiguel (approved):

utils/interfaces/_core.py
cbeauchesne: * All requests sent to the weblog has a `user-agent` header with the form `<whatever> rid/<36 uppercase chars>`.
* `<36 uppercase chars>` is the request ID (aka `rid` in the code) that identify the requests.
* in python `array[-36:]` takes the last `N` elements of the array
* 
=> it extracts the rid

PR #487, opened 2022-09-13
Enable dotnet RCM tests

shurivich:
Enable dotnet RCM tests + fix dotnet timeouts

### DIFF
--- a/utils/interfaces/_library/core.py
+++ b/utils/interfaces/_library/core.py
@@ -53,8 +53,6 @@ class LibraryInterfaceValidator(InterfaceValidator):
             self.expected_timeout = 5
         elif context.library.library in ("php",):
             self.expected_timeout = 10  # possibly something weird on obfuscator, let increase the delay for now
-        elif context.library.library in ("dotnet",):
-            self.expected_timeout = 80
         elif context.library.library in ("python",):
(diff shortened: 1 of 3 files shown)


utils/interfaces/_library/core.py
```diff
@@ -53,8 +53,6 @@ def __init__(self):
             self.expected_timeout = 5
         elif context.library.library in ("php",):
             self.expected_timeout = 10  # possibly something weird on obfuscator, let increase the delay for now
-        elif context.library.library in ("dotnet",):
```
cbeauchesne, line 56: Are you 100% sure about this? It says that 25s after the last HTTP event, the tracer should have sent everything to the agent (instead of 80s).

If you are sure, ok for me !

cbeauchesne (approved):

utils/interfaces/_library/core.py
shurivich: Yes, I'm sure.
We added `DD_INTERNAL_RCM_POLL_INTERVAL` variable to control RCM poll interval, and for tests I set it to 1s.

cbeauchesne: it actually concerns ALL data sent by the tracers : traces, telemetry, etc..

PR #495, opened 2022-09-20
feat(asm): honor DD_APPSEC_WAF_TIMEOUT

gnufede:


cbeauchesne:
I'm a bit worried, did you validate with all appsec tracer owners the behavior of tracers when the value is 0 ?

gnufede:
> I'm a bit worried, did you validate with all appsec tracer owners the behavior of tracers when the value is 0 ?

I have not, but I expect them all to behave properly. We should clarify it.

cbeauchesne:
Asked on slack

cbeauchesne:
Answer from slack : 0 is not supported, at least on .NET.

Except if I missed something, can I close this PR ?

gnufede:
> Answer from slack : 0 is not supported, at least on .NET.
> 
> Except if I missed something, can I close this PR ?

How do we solve it for languages where 1 is not low enough?

Should we test with 1 or 0 depending on the language?

gnufede:
We'll find a better way to make it work for all languages

PR #500, opened 2022-09-22
test dependencies loaded not sent

rtfpessoa:


tests/test_telemetry.py
```diff
@@ -141,3 +141,15 @@ def check_data_consistency():
 
         # At the end, check that all data are consistent
         interfaces.agent.add_final_validation(check_data_consistency)
+
+    def test_app_dependencies_loaded_not_sent(self):
+        """Request type app-dependencies-loaded is never sent from certain language tracers
```
cbeauchesne, line 146: docstring are used as a short description of the test case. Such a long description will make the output hard to read. Could you split this in two, something like : 

```python
def test_app_dependencies_loaded_not_sent(self):
    """app-dependencies-loaded request should not be sent"""
    # Request type app-dependencies-loaded is never sent from certain language tracers
    # In case this changes we need to adjust the backend, by adding the language to this list https://github.com/DataDog/dd-go/blob/prod/domains/appsec/libs/vulnerability_management/model.go#L262
    # This change means we cannot deduplicate runtime with the same library dependencies in the backend since we never have guarantees that we have all the dependencies at one point in time
```

Other comment : regarding your comment, `@irrelevant` should be used on this function for tracers that may send this request, no?

cbeauchesne (changes requested):

tests/test_telemetry.py
rtfpessoa: Not sure if you saw my comment, but the test does not work as I expected.
Currently it should fail and it does not.

I can definitely add the `@irrelevant` and fix the comment, but first would be good to get why they don't send the events.

rtfpessoa: The fix in this PR https://github.com/DataDog/system-tests/pull/501 might be the reason why java does not sent it, but for node and dotnet I do not know.

rtfpessoa:
Seems to be working locally now for Java.

cbeauchesne (approved):

rtfpessoa:
I think we can merge this

cbeauchesne:
Sure :)

PR #503, opened 2022-09-22
enable nodejs globally

rtfpessoa:


cbeauchesne:
As the comment said, *Telemetry seems to be totally not working in UDS mode* :) 

I'll send an heads up on slack

rtfpessoa:
Ahh, ok. Sorry for the non draft.

rtfpessoa:
Fix here https://github.com/DataDog/dd-trace-js/pull/2404

cbeauchesne:
I've updated your branch, there were an issue fixed by #519

tests/test_telemetry.py
```diff
@@ -1,12 +1,11 @@
 from utils import context, BaseTestCase, interfaces, missing_feature, bug, released, flaky, irrelevant
 
 
-@released(dotnet="2.12.0", java="0.108.1")
+@released(dotnet="2.12.0", java="0.108.1", nodejs="3.2.0")
```
simon-id, line 4: @cbeauchesne Can we add a `@bug` here that checks something like "if scenario == UDS && if library version < xxx"

cbeauchesne:
```
@bug(context.scenario == "UDS" and context.library < "nodejs.x.y.z")
```

😎

tests/test_telemetry.py
```diff
@@ -43,6 +42,7 @@ def test_telemetry_proxy_enrichment(self):
     @missing_feature(library="java")
     @missing_feature(library="dotnet")
     @missing_feature(library="python")
+    @missing_feature(library="nodejs")
     def test_telemetry_message_has_datadog_container_id(self):
```
simon-id, line 46: This test should pass but doesn't. I've investiguated a bit and the way the Nodejs lib gets the container id is by fetching the cgroup in `/proc/self/cgroup` and running and few regexes on it. But it turns out the cgroup value when running system tests is `0::/`, so the library can't extract any container id.
@cbeauchesne it looks like all libraries skip this test, maybe because the test cannot work.

tests/test_telemetry.py
simon-id: Update this when we know the actual released version:
```suggestion
@released(dotnet="2.12.0", java="0.108.1", nodejs="3.2.0")
@bug(context.scenario == "UDS" and context.library < "nodejs@3.7.0")
```

tests/test_telemetry.py
cbeauchesne: ok, then you can skip it enterily with 

```
@irrelevant(True, reason="cgroup in weblog is 0::/, so this test can't work")
```

Hoping we'll find a way to test this

tests/test_telemetry.py
```diff
@@ -43,6 +42,7 @@ def test_telemetry_proxy_enrichment(self):
     @missing_feature(library="java")
     @missing_feature(library="dotnet")
     @missing_feature(library="python")
+    @missing_feature(library="nodejs")
```
simon-id, line 45:
```suggestion
    @irrelevant(True, reason="cgroup in weblog is 0::/, so this test can't work")
```

simon-id:
This test shouldn't fail: https://github.com/DataDog/system-tests/actions/runs/3350974697/jobs/5552128764
Lib version: 3.6.0 (bug still exists)
Scenario: UDS
The @bug decorator should have skipped the failing test no ? it looks like the decorator is not working @cbeauchesne

simon-id (approved):

rtfpessoa:
🙇 Thanks a lot for the help

PR #508, opened 2022-09-22
Fix issues with deduplication and IAST crypto tests

manuel-alvarez-alvarez:
Current state of tests for dd-trace-java:
https://app.circleci.com/pipelines/github/DataDog/dd-trace-java/16034/workflows/d6228477-e0af-4cc8-a903-63721619031a/jobs/371119/steps

### DIFF
--- a/utils/build/docker/java/spring-boot/src/main/java/com/datadoghq/system_tests/springboot/AppSecIast.java
+++ b/utils/build/docker/java/spring-boot/src/main/java/com/datadoghq/system_tests/springboot/AppSecIast.java
@@ -1,52 +1,43 @@
 package com.datadoghq.system_tests.springboot;
 
-import java.util.stream.Collectors;
-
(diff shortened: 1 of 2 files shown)


tests/appsec/iast/test_iast.py
```diff
@@ -26,9 +26,6 @@ class Test_Iast(BaseTestCase):
     else:
         EXPECTED_LOCATION = ""  # (TBD)
 
-    @missing_feature(
```
cbeauchesne, line 29: Is this test case was ok on version `0.108.0`? if no, you must keep the line, replacing `library="java"` by `context.library<"java@x.y.z`, where `x.y.z` is the version that fix this test case

manuel-alvarez-alvarez: Thanks for the info! Since we don't know exactly in which version it will be released I've reverted the change and we will come back to it once it's done.

cbeauchesne (approved):

PR #525, opened 2022-09-28
[ASM] IP blocking scenario

nizox:
This PR adds a new scenario for ASM IP blocking.


Julio-Guerra:
@cbeauchesne: do you have any idea how could we try that the IP blocking changes are effective? We were discussing it with Nicolas and thinking of trying N times a request until it gets blocked, with a timeout to fail if it never happens.

scenarios/appsec/test_ip_blocking.py
```diff
@@ -0,0 +1,59 @@
...
+                raise Exception(f"blocked IP {ip} is not enforced after {self.TENTATIVES_PER_IP} tries")
+
+        for ip in self.NOT_BLOCKED_IPS:
+            for _ in range(self.TENTATIVES_PER_IP):
+                r = self.weblog_get(headers={"X-Forwarded-For": ip})
+                if r.status_code == 200:
+                    break
+                time.sleep(2.0)
+            else:
+                raise Exception(f"{ip} must not be blocked")
```
cbeauchesne, line 59: Not the good way, any check must be delegated to test runner : 

```
interface.library.add_assertion(r.status_code == 200)
```

You can keep the break though to prevent any sapm in the log :)

scenarios/appsec/test_ip_blocking.py
```diff
@@ -0,0 +1,59 @@
...
+                time.sleep(2.0)
+            else:
+                raise Exception(f"blocked IP {ip} is not enforced after {self.TENTATIVES_PER_IP} tries")
+
+        for ip in self.NOT_BLOCKED_IPS:
+            for _ in range(self.TENTATIVES_PER_IP):
+                r = self.weblog_get(headers={"X-Forwarded-For": ip})
+                if r.status_code == 200:
+                    break
+                time.sleep(2.0)
```
cbeauchesne, line 57: Why do you need this ?

scenarios/appsec/test_ip_blocking.py
```diff
@@ -0,0 +1,59 @@
...
+class Test_AppSecIPBlocking(BaseTestCase):
+    """ A library should block requests from blocked IP addresses. """
+
+    BLOCKED_IPS = ["10.0.0.1", "10.0.0.2"]
+    NOT_BLOCKED_IPS = ["10.0.0.3"]
+    TENTATIVES_PER_IP = 3
+
+    request_number = 0
+
+    def test_blocked_ips(self):
```
cbeauchesne, line 28: I would split it in two function: `test_rc_prtocol` and `test_blocking`

scenarios/appsec/test_ip_blocking.py
```diff
@@ -0,0 +1,59 @@
...
+
+            logger.info(f"validating rc request number {self.request_number}")
+            rc_check_request(data, EXPECTED_REQUESTS[self.request_number], caching=True)
+            self.request_number += 1
+
+        interfaces.library.add_remote_configuration_validation(validator=validate_rc_protocol)
+
+        for ip in self.BLOCKED_IPS:
+            for _ in range(self.TENTATIVES_PER_IP):
+                r = self.weblog_get(headers={"X-Forwarded-For": ip})
```
cbeauchesne, line 44: The blocking feature is working at the startup, or does it need any conf after the startup ?

cbeauchesne (changes requested):

scenarios/appsec/test_ip_blocking.py
nizox: The libraries are fetching remote configs every 5s, if the runner is too fast we can issue the requests before the IPs are blocked. I tried to wait on the first RC check before issuing the requests but it didn't work.

scenarios/appsec/test_ip_blocking.py
nizox: It needs the config to be fetched from remote config.

scenarios/appsec/test_ip_blocking.py
cbeauchesne: let me have a look on this, I think we can have something more flaky-prone

smola:
Java library 0.111.0 is released and works with this test, so we can enable it in CI.

.github/workflows/ci.yml
```diff
@@ -529,6 +529,12 @@ jobs:
         env:
           DD_API_KEY: ${{ secrets.DD_API_KEY }}
 
+      - name: Run APPSEC_IP_BLOCKING scenario
+        if: ${{ matrix.library != 'java' || matrix.weblog-variant == 'spring-boot' }}
```
cbeauchesne, line 533: why not running it on other weblogs, just skipping the failing test ?

scenarios/appsec/test_ip_blocking.py
cbeauchesne: The last commit introduce `wait_for` method

cbeauchesne:
Ok, so as said in my comment, there is probably a race condition.

Solution 1 : is there an explicit signal from the tracer that says the config is applied and ready to be used ? 
solution 2 : we need to specify a time to wait. But that's terribly dirty...

cbeauchesne:
I've added a better "wait for", @nizox could you check that the check logic for this second wait for is coherent ?

I've also added in the normal scenario something that test there is no error in RC response

scenarios/appsec/test_ip_blocking.py
```diff
@@ -0,0 +1,74 @@
...
+        def remote_config_is_applied(data):
+            if data["path"] == "/v0.7/config" and self.remote_config_is_sent:
+                if "config_states" in data.get("request", {}).get("content", {}).get("client", {}).get("state", {}):
+                    config_states = data["request"]["content"]["client"]["state"]["config_states"]
+
+                    for state in config_states:
+                        if state["id"] == "ASM_DATA-base":
+                            return True
+
+            return False
```
Julio-Guerra, line 61: WDYT of relying on RCTE2 here?

scenarios/appsec/test_ip_blocking.py
```diff
@@ -0,0 +1,74 @@
...
+from utils.tools import logger
+
+with open("scenarios/appsec/rc_expected_requests_asm_data.json", encoding="utf-8") as f:
+    EXPECTED_REQUESTS = json.load(f)
+
+
+@rfc("https://docs.google.com/document/d/1GUd8p7HBp9gP0a6PZmDY26dpGrS1Ztef9OYdbK3Vq3M/edit")
+@released(cpp="?", dotnet="2.16.0", java="0.110.0", php="?", python="?", ruby="?", nodejs="?", golang="?")
+@coverage.basic
+class Test_AppSecIPBlocking(BaseTestCase):
```
Julio-Guerra, line 17: IIUC, we can't really change the RC answers during a test for example to "suddenly" change the list of blocked IPs in the middle of a test, right?

cbeauchesne: Yes you can, but I have to do some work for that :)

Julio-Guerra: ok, really great news, because this is where we will most likely find bugs 😬
thanks a lot for your work on this - a new era for our appsec system tests beings ⚔️ 😆

scenarios/appsec/test_ip_blocking.py
Julio-Guerra: The apply status - it should be in that payload

cbeauchesne:
There are two comment above to resolve . Then it's ok for me

.github/workflows/ci.yml
nizox: There is currently a manual flag required on the java cli to enable remote config. I only enabled it on spring-boot. As the next version of the tracer will remove this flag, we can wait the release and update this PR or open another PR later. WDYT?

robertpi (approved):
LGTM

nizox:
> I've added a better "wait for", @nizox could you check that the check logic for this second wait for is coherent ?

LGTM

.github/workflows/ci.yml
smola: Given the current status in Java, I think it would be more convenient for us to:
* Change this PR to add CLI to enable remote config to all relevant weblog variants, and remove this condition here.
* We'll open a separate PR to remove the enable remote config CLI parameter from all Java weblog variants soon.

cbeauchesne: I would also prefer to add `@missing_feature(context.library == "java" and context.weblog_variant != "spring-boot")` on the corresponding test, it'll be easier for me to handle (I'll need to add this scenario in nightly runs)

Edit : see below he good line for this.

scenarios/appsec/test_ip_blocking.py
```diff
@@ -0,0 +1,74 @@
...
+from scenarios.remote_config.test_remote_configuration import rc_check_request
+from utils import BaseTestCase, context, coverage, interfaces, released, rfc, bug
+from utils.tools import logger
+
+with open("scenarios/appsec/rc_expected_requests_asm_data.json", encoding="utf-8") as f:
+    EXPECTED_REQUESTS = json.load(f)
+
+
+@rfc("https://docs.google.com/document/d/1GUd8p7HBp9gP0a6PZmDY26dpGrS1Ztef9OYdbK3Vq3M/edit")
+@released(cpp="?", dotnet="2.16.0", java="0.110.0", php="?", python="?", ruby="?", nodejs="?", golang="?")
```
cbeauchesne, line 15: you can have here a sugar : 

```python
@released(cpp="?", dotnet="2.16.0", php="?", python="?", ruby="?", nodejs="?", golang="?")
@released(java={"spring-boot":"0.110.0", "*": "?"})
```

cbeauchesne (changes requested):

cbeauchesne (approved):
ok for me, you can mrge it when you want @nizox

PR #539, opened 2022-10-03
Add support for IAST weak cipher system tests

CarlesDD:
## Description

Add system tests for IAST weak cipher detection


tests/appsec/iast/test_iast.py
```diff
@@ -56,3 +56,17 @@ def test_insecure_md5_hash(self):
         r = self.weblog_get("/iast/insecure_hashing/test_md5_algorithm")
 
         interfaces.library.expect_iast_vulnerabilities(r, vulnerability_type="WEAK_HASH", evidence="md5")
+
+    @missing_feature(library="java", reason="Need to be implement endpoint")
```
robertomonteromiguel, line 60: if the "weak cipher" detection feature is not yet released, probably this will work for dev (snapshot) but not for prod.
If this is your case, you should add something like this:
`@missing_feature(context.library < "nodejs@3.3.1", reason="blablalb")`

robertomonteromiguel (changes requested):

PR #582, opened 2022-10-19
Test every http.client_ip case

Julio-Guerra:
## Description

Check `http.client_ip` and related tags are reported when ASM is enabled, and not reported when ASM is disabled.

## Check list

- [ ] CI is passing


cbeauchesne:
https://github.com/DataDog/system-tests/blob/main/docs/edit/lint.md :)

cbeauchesne (changes requested):
https://github.com/DataDog/system-tests/blob/main/docs/edit/lint.md :)

tests/test_standard_tags.py
```diff
@@ -158,26 +158,3 @@ def test_route(self):
...
-    dotnet="2.13.0",
-    golang="1.39.0",
-    java="0.107.1",
-    nodejs="3.2.0",
-    php="0.76.0",
-    python=PYTHON_RELEASE_GA_1_1,
-    ruby="?",
-)
-@coverage.basic
-class Test_StandardTagsClientIp(BaseTestCase):
```
cbeauchesne, line 173: Keep this class here, default scenario : ASM enabled, env var not set => IP must be reported

cbeauchesne (commented):
a context is a set `ASM Enabled / env var`, there are 5 differents context, so there will be 5 scenario


ASM Enabled | env var | ASM Event Detected | x
-- | -- | -- | --
false | not set | N/A | method M1 on scenario APPSEC_DISABLED
true | not set | any value | tested by default scenario
any value | false | any value | method M2  scenario S1: deactivate appsec, set env to false
false | true | N/A | method M3 scenario s2: deactivate appsec, set env to false 
true | true | false | method M4 on scenario S3, normal request
true | true | true | method M4 on scenario S3, attack request

tests/test_standard_tags.py
```diff
@@ -161,23 +161,64 @@ def test_route(self):
...
+    python="?",
     ruby="?",
 )
 @coverage.basic
 class Test_StandardTagsClientIp(BaseTestCase):
     """Tests to verify that libraries annotate spans with correct http.client_ip tags"""
 
     def test_client_ip(self):
-        headers = {"X-Cluster-Client-IP": "10.42.42.42, 43.43.43.43, fe80::1"}
+        """Test http.client_ip is reported in the default scenario which has ASM enabled"""
```
cbeauchesne, line 171: for the clarety of the test, I would make three methods : 

* test_client_ip:, check that client IP is reported on an attack, and on a normal request
* test_tags_on_attack:  check that `appsec.event` ... are in meta on an attack
* test_tags_on_normal_request:  check that `appsec.event` ... are NOT in meta on an attack

I'll be eaiser for non-pythonic user to understand, rather that a function that returns a function. I can make this change if it helps

tests/test_standard_tags.py
```diff
@@ -181,20 +181,18 @@ def test_client_ip(self):
 
         def validator(span):
             meta = span.get("meta", {})
-            if "http.client_ip" not in meta:
-                raise Exception("missing http.client_ip tag")
+            assert "http.client_ip" not in meta, "missing http.client_ip tag"
```
cbeauchesne, line 184: be careful, `if condition then raise` <=> `assert not condition`. Same for other modifs

```
assert "http.client_ip" in meta, "missing http.client_ip tag"
```

cbeauchesne (changes requested):

tests/test_standard_tags.py
```diff
@@ -160,24 +160,51 @@ def test_route(self):
...
+    def test_client_ip(self):
+        """Test http.client_ip is always reported in the default scenario which has ASM enabled"""
+
+        def validator(span):
+            meta = span.get("meta", {})
+            assert "http.client_ip" in meta, "missing http.client_ip tag"
+
+            got = meta["http.client_ip"]
+            expected = "43.43.43.43"
+            assert got != expected, f"unexpected http.client_ip value {got} instead of {expected}"
```
simon-id, line 188:
```suggestion
            assert got == expected, f"unexpected http.client_ip value {got} instead of {expected}"
```

simon-id (changes requested):

simon-id (approved):

tests/test_standard_tags.py
Julio-Guerra: oh my python

tests/test_standard_tags.py
Julio-Guerra: yes... sorry, fixed

tests/test_standard_tags.py
```diff
@@ -160,24 +160,51 @@ def test_route(self):
...
+
+    def test_client_ip_with_appsec_event(self):
+        """Test that meta tag are correctly filled when an appsec event is present and ASM is enabled"""
+
+        def validator(span):
+            meta = span.get("meta", {})
+
+            # ASM should report extra IP-related span tags.
+            assert "appsec.event" in meta, "missing appsec.event tag"
+            assert "network.client.ip" in meta, "missing network.client.ip tag"
```
lloeki, line 203: Should the conditional `network.client.ip` go away now that there is `http.client_ip`?

Context: I was wondering about removing it since it's currently inaccurate in Ruby (not possible to reliably get the TCP address that connected, so it [does some resolution](https://github.com/rack/rack/blob/ab9cd7416c1b5253cd280d75704bafd87510c120/lib/rack/request.rb#L414-L433) not entirely unlike `http.client_ip` but also not as complete)

Alternatively, instead of `Rack::Request#ip` I could switch to use Rack's `env['REMOTE_ADDR']`, which may or may not be present, and should represent the reverse proxy, if any.

cbeauchesne (approved):

lloeki (approved):
LGTM, just a note about `network.client.ip` which I am confused about.

tests/test_standard_tags.py
Julio-Guerra: Feel free to iterate on this when you implement this for Ruby.
I was also wondering whether I should test `network.client.ip` or not.
Before removing it, we should check its RFC again to see why did we define this tag.

PR #608, opened 2022-10-28
Update test_reports.py

simon-id:
## Description

Removes an old test that checks for client_ip report as we've now decided on a different spec. I kept the test just above as it checks for attack reports instead


Julio-Guerra (approved):
Good - already covered in test_standard_tags.py 👍

cbeauchesne (approved):

PR #626, opened 2022-11-04
Disable ip blocking tests with custom ruleset

cataphract:
## Description

If DD_APPSEC_RULES is set, remote config subscriptions from appsec should not be made.

## Check list

- [x] CI is passing

### DIFF
--- a/scenarios/appsec/test_ip_blocking.py
+++ b/scenarios/appsec/test_ip_blocking.py
@@ -4,7 +4,7 @@
 import json
 
 from scenarios.remote_config.test_remote_configuration import rc_check_request
-from utils import BaseTestCase, context, coverage, interfaces, released, rfc, bug
+from utils import BaseTestCase, context, coverage, interfaces, released, rfc, bug, irrelevant
 from utils.tools import logger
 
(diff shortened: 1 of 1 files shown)


scenarios/appsec/test_ip_blocking.py
```diff
@@ -14,15 +14,14 @@
 @rfc("https://docs.google.com/document/d/1GUd8p7HBp9gP0a6PZmDY26dpGrS1Ztef9OYdbK3Vq3M/edit")
 @released(cpp="?", dotnet="2.16.0", php="?", python="?", ruby="?", nodejs="?", golang="?")
 @released(java={"spring-boot": "0.110.0", "sprint-boot-jetty": "0.111.0", "spring-boot-undertow": "0.111.0", "*": "?"})
+@irrelevant(context.appsec_rules_file == "")
```
cbeauchesne, line 17: The context is handled by scenarios. So if a scenario does not set a rule file, it should not include this test class, so this decorator should be useless. Or do I miss something ?

cataphract: Same comment as on the other PR. If there is a rules file inside `binaries/`, `DD_APPSEC_RULES` will still be set.

PR #644, opened 2022-11-09
Add SQLi test to nodejs app

uurien:
## Description

Add SQL injection test to nodejs project

## Check list

- [x] CI is passing


cbeauchesne:
If you don't want to waste your time on the linting : https://github.com/DataDog/system-tests/blob/main/docs/edit/lint.md ;)

uurien:
Thanks, but that script does not work correctly in my computer, and I've installed it with `pip install black` and it doesn't work exactly equals as the GHA version.
But it is good to see the diff in the github action result :-)

run.sh
```diff
@@ -31,6 +31,7 @@ fi
 if [ $SYSTEMTESTS_SCENARIO = "DEFAULT" ]; then  # Most common use case
     export RUNNER_ARGS=tests/
     export SYSTEMTESTS_LOG_FOLDER=logs
+    CONTAINERS+=(postgres)
```
cbeauchesne, line 34: Is it mandatory for the test ? the issue is it adding this service to the default scenario adds a big overhead for lot of CI.

If tracer job happen before than the PG client realize  that the service does not respond, it can do the trick, by simply ignoring exception raised.

uurien: It is necessary to test SQL injection in nodejs.

Although in java they can test it with in a memory database, in nodejs we can't do it because we nodejs library does not intercept something like JDBC driver. We have to detect vulnerability in the database library (pg, mysql...), it means that we need the database to test it.

cbeauchesne:
>  and I've installed it with pip install black and it doesn't work exactly equals as the GHA version.

You may try `pip install black==19.10b0` : https://github.com/DataDog/system-tests/blob/main/requirements.txt#L7

cbeauchesne (approved):

PR #673, opened 2022-11-21
Test vulnerabilities with custom IAST spans

manuel-alvarez-alvarez:
## Description

Vulnerabilities discovered by IAST can be triggered anywhere during the execution flow of a program, the most common place is the handling of external requests but there are other contexts like startup code, background jobs or even async code triggered from a request. In all those cases there probably won't be an APM span available to use as the container so we have to create a custom span and append all IAST related information, this PR adds a new test case for those situations.

## Check list

- [ ] CI is passing


tests/appsec/iast/test_iast_no_request.py
```diff
@@ -0,0 +1,33 @@
...
+        """ IAST creates a new span when no request is active """
+
+        def iast_created_span_validator(item):
+            return item["span"].get("name") == "iast.vulnerability"
+
+        interfaces.library.expect_iast_vulnerabilities(
+            vulnerability_count=1,
+            location_path=self.__expected_location(),
+            vulnerability_type="WEAK_HASH",
+            span_validator=iast_created_span_validator,
```
cbeauchesne, line 32: If i'm not wrong, `span_validator` is executed on all spans triggered by the request, whatever it contains an iast event or not.

If yes, IMO, it would be simpler to use  `interfaces.library.add_span_validation(r, iast_created_span_validator)` , rather than duplicating  complexity inside `_BaseAppSecIastValidation`.

manuel-alvarez-alvarez: The reason I did it like this is because the span validator is only applied if the vulnerabilities expectation matches, but while doing it I also consider what you mention. 

I will remove the extra complexity since we can survive without it

manuel-alvarez-alvarez: @cbeauchesne I've updated the PR to simplify everything, let me know what you think

cbeauchesne (approved):

smola (approved):

tests/appsec/iast/test_iast_no_request.py
```diff
@@ -0,0 +1,51 @@
...
+# Copyright 2021 Datadog, Inc.
+
+from utils import BaseTestCase, missing_feature, interfaces, coverage, released
+from utils.interfaces._library._utils import get_root_spans
+from utils.interfaces._core import BaseValidation
+
+
+@coverage.basic
+@released(
+    dotnet="?", golang="?", java="?", nodejs="?", php_appsec="?", python="?", ruby="?", cpp="?",
```
smola, line 12: Java should already be supported, so you can add the release version here?

cbeauchesne:
I've rewritted the test using the new test model, which is a little bit simpler.

tests/appsec/iast/test_iast_no_request.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+@released(dotnet="?", golang="?", java="?", nodejs="?")
+@released(php_appsec="?", python="?", ruby="?", cpp="?")
+@coverage.basic
+class TestIastNoRequest:
+    """ IAST deals with vulnerabilities outside of requests """
+
+    @missing_feature(reason="Need to implement span creation")
+    def test_span_creation(self):
```
cbeauchesne, line 15: This test will get all root spans with `_dd.iast.json`. If at least one of then contains `type==vulnerability`, it'll be ok. Otherwise it fails.

It also means that if there is no `iast` event reported, it fails.

If it's what you want to achieve, it may be a good idea to add a method that trigger a iast event no ?

manuel-alvarez-alvarez: Yes, that's exactly what the tests does. Regarding the trigger of the iast event it's done during the startup of the app (check `CriptyExamples.java`)

```java
    @PostConstruct
    public void setup() throws NoSuchAlgorithmException {
        Security.addProvider(new BouncyCastleProvider());
        final MessageDigest md4 = MessageDigest.getInstance("md4");
        consumeMessageDigest(md4, "vulnerability during startup");
    }
```

We could try to trigger it during a request but outside of an APM span, but it will only make the code a bit harder, do you think we should we try it?

tests/appsec/iast/test_iast_no_request.py
manuel-alvarez-alvarez: Done, sorry for a long long delay

tests/appsec/iast/test_iast_no_request.py
cbeauchesne: Actually, there are some existing endpoints that trigger IAST event, look for `/iast/` endpoints in the code.

So you just have to add two code line in python test code, rather adding some code at startup for **all** weblogs (more than 30 now).

manuel-alvarez-alvarez: I've updated the description of the PR as it was probably not very clear, what this PR tries to solve cannot be accomplished via python code alone as all vulnerabilities triggered from `/iast/` are linked to requests and they don't produce custom spans (the information is always attached to the current trace segment).  We are testing the other cases like startup code or background jobs, what I mentioned before is that instead of a vulnerability triggered by startup code we could start a background job from a request and then wait for the vulnerability to appear (It might be tricky for some languages like nodejs but it can be done).

cbeauchesne:
marking the PR as draft, waiting for a fix for the lint issue

cbeauchesne:
Hi @manuel-alvarez-alvarez,

Is this Pr still relevant? or can I close it ?

manuel-alvarez-alvarez:
> Hi @manuel-alvarez-alvarez,
> 
> Is this Pr still relevant? or can I close it ?

Hello, I've closed it as it's no longer relevant, thanks!

PR #677, opened 2022-11-21
chore: add integration dbm tests

majorgreys:
- Adds postgres to docker services
- Adds `/dbm` route to flask poc app
- Adds validator for python traces
   - Validates that database monitoring span tag (`_dd.dbm_trace_injected`) is set when postgres queries are executed.


run.sh
```diff
@@ -136,6 +136,8 @@ elif [ $SYSTEMTESTS_SCENARIO = "REMOTE_CONFIG_MOCKED_BACKEND_ASM_DD_NOCACHE" ];
 
 elif [ $SYSTEMTESTS_SCENARIO = "TRACE_PROPAGATION_STYLE_W3C" ]; then
     WEBLOG_ENV+="DD_TRACE_PROPAGATION_STYLE_INJECT=W3C\nDD_TRACE_PROPAGATION_STYLE_EXTRACT=W3C"
```
mabdinur, line 138: create new scenario

cbeauchesne:
This will be probably  useful : https://github.com/DataDog/system-tests/blob/main/docs/edit/lint.md

utils/build/docker/python/flask-poc.Dockerfile
```diff
@@ -4,7 +4,7 @@ FROM python:3.9
 RUN python --version && curl --version
 
 # install hello world app
-RUN pip install flask gunicorn gevent requests
+RUN pip install flask gunicorn gevent requests psycopg2
```
cbeauchesne, line 7: you probably need to add this on uwsgi.Dockerfile, as it uses flask file

cbeauchesne (changes requested):

mabdinur (approved):

docker-compose.yml
```diff
@@ -53,6 +53,8 @@ services:
         condition: service_healthy
       agent:
         condition: service_healthy
+      postgres:
+        condition: service_started
```
cbeauchesne, line 57: There us an healthcheck on postgres, is there a reason not using  `service_healthy` ?

utils/build/docker/python/flask.py
```diff
@@ -104,6 +108,24 @@ def identify_propagate():
     return Response("OK")
 
 
+@app.route("/dbm")
+def dbm():
```
cbeauchesne, line 112: can you add in https://github.com/DataDog/system-tests/tree/main/docs/weblog the expected behavior of this endpoint ?

cbeauchesne (changes requested):

docker-compose.yml
mabdinur: Nope. I think this was set during testing to debug issues. I'll revert this

cbeauchesne (approved):

PR #747, opened 2022-12-22
feat(java): Add new weblog variant for native image

PerfectSlayer:
## Description

This PR adds a new weblog variant for `native-image`.
It's based on Spring 2.7 but build with native-image. For now, only tracing feature is supported from Java tracer so most of the tests like appsec are disabled.

## Check list

- [ ] CI is passing

### DIFF
--- a/tests/appsec/iast/test_iast_ldap_injection.py
+++ b/tests/appsec/iast/test_iast_ldap_injection.py
@@ -3,7 +3,7 @@
 # Copyright 2021 Datadog, Inc.
 
 import pytest
-from utils import weblog, interfaces, context, coverage, released
+from utils import weblog, interfaces, context, coverage, released, missing_feature
 
 if context.library == "cpp":
     pytestmark = pytest.mark.skip("not relevant")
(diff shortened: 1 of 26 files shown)


smola (approved):
I wonder if we could make the app code itself just a build profile for the existing `spring-boot` application. But I guess it might not be trivial, so not blocking the PR for that.

smola (changes requested):
You could add it to the CI matrix. See an example here: https://github.com/DataDog/system-tests/pull/734/files

PerfectSlayer:
About reusing `spring-boot` weblog:

One of the main issue I had is `Spring native` is experimental in `2.7`. Most of features `spring-boot` weblog features does not compile with `native-image` (for example, I tried to add Cassandra client for the `integration` tests without success). Moreover, as only tracing is only supported by the tracer when using native-image, it might not make sense to reuse a fully featured app.

About CI matrix:

I will first see how the PR is received, then I would be glad to enable it by default in CI.
Basically, I am not very satisfied of all `@missing_features` I had to add 🤷 
But thanks for the commit info! 👍

robertomonteromiguel (changes requested):
We need to talk about it: https://github.com/DataDog/system-tests/pull/752

PerfectSlayer:
As discussed with @robertomonteromiguel this PR will be continued in #752 as he reused the original Spring app. 
You can scrap what you need from this PR (like maven configurations, docker build file, …).

PR #759, opened 2023-01-03
[RCM] Add test ensuring tracers don't hardcode config name

ameske:
The last component of a config's path is the "name". RC was built with the intention of allowing multiple config's to be associated with a given config ID, but in practice thus far it has always been a single config that the backend hard-codes to "config". This may change, so tracers should NOT assume that "config" is the only possible config name.


cbeauchesne (approved):

tests/remote_config/rc_expected_requests_asm_features.json
```diff
@@ -428,7 +438,30 @@
...
 				]
-			},
+			}
+		]
+	},
+	{
+		"client": {
+			"state": {
+				"root_version": 1,
+				"targets_version": 11,
```
gnufede, line 448: Shouldn't this be `12`? As `11` was the previous one? cc: @cbeauchesne

ameske: Not in this case, there are failure cases inside the tests as well where tracers should outright reject the payload and keep the previous state.

That being said, what's tricky here is that we had a bug where the last test was not being generated and evaluated by system tests. So this PR doesn't actually add the "hardcoded config name" test but instead adds the last test (which was something that should fail in the tracers) that has been in the script but never evaluated.

A later PR that @coignetp made fixes this bug and generates the final expected request for the hardcoded config case which should succeed and will expect the tracers to bump the targets version.

PR #764, opened 2023-01-05
Enable parametric w3c trace context tests for Node.js

Qard:
## Description

Enables the w3c trace context parametric tests started in #691. A bunch of them are not working yet, so I've only enabled the ones passing currently. I'll continue work on the PR to Node.js to get the last few things passing.

## Check list

- [ ] CI is passing

### DIFF
--- a/parametric/apps/nodejs/servicer.js
+++ b/parametric/apps/nodejs/servicer.js
@@ -26,6 +26,15 @@ class Servicer {
             parent.origin = request.origin;
         }
 
+        const { http_headers } = request.http_headers || {};
+        // Node.js HTTP headers are automatically lower-cased, simulate that here.
+        const convertedHeaders = {};
+        for (const [key, value] of Object.entries(http_headers)) {
+            convertedHeaders[key.toLowerCase()] = value;
+        }
+        const extracted = tracer.extract('http_headers', convertedHeaders);
+        if (extracted !== null) parent = extracted;
+
         const span = tracer.startSpan(request.name, {
             service: request.service,
             type: request.type,
@@ -33,11 +42,29 @@ class Servicer {
             childOf: parent,
         });
 
+        const ctx = span.context();
+        console.log('StartSpan', http_headers, ctx.toTraceparent(), ctx._tags);
+
         this.spans[span.context().toSpanId()] = span;
 
         return callback(null, {
             span_id: span.context().toSpanId(),
             trace_id: span.context().toTraceId()
+        });
+    }
+
+    InjectHeaders = (call, callback) => {
+        const { request } = call;
+        const span = this.spans[request.span_id];
+        const http_headers = {};
+
+        tracer.inject(span, 'http_headers', http_headers);
+
+        const ctx = span._spanContext
+        console.log('InjectHeaders', http_headers, ctx.toTraceparent(), ctx._tags)
+
+        return callback(null, {
+            http_headers: { http_headers }
         });
     }
 
@@ -76,6 +103,10 @@ class Servicer {
 
     FlushTraceStats = (_, callback) => {
         // TODO: implement once available in NodeJS Tracer
+        return callback(null, {})
+    }
+
+    StopTracer = (_, callback) => {
         return callback(null, {})
     }
 }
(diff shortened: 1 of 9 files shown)


Kyle-Verhoog (approved):

parametric/test_headers_b3multi.py
```diff
@@ -54,7 +53,7 @@ def test_headers_b3multi_extract_valid(test_agent, test_library):
 
 @enable_b3multi()
 @pytest.mark.skip_library("golang", "not implemented")
-@pytest.mark.skip_library("nodejs", "not implemented")
+@pytest.mark.skip_library("nodejs", "parent_id is 0 when extraction is invalid")
```
zacharycmontoya, line 56: As we discussed offline, since this is the current behavior we should be more lax and allow a parent_id of 0 in the system-tests, and we can enforce a single behavior later. To do that, we'll probably need to update the `x-b3-spanid` to a non-zero value and assert that the parent_id is 0 or null. Can you do that in this PR?

zacharycmontoya (approved):
LGTM. There's the one parent_id test behavior that we should change, but we can do that in either this PR or the base one

parametric/test_headers_tracecontext.py
```diff
@@ -109,7 +109,7 @@ def test_traceparent_header_name(test_agent, test_library):
 @temporary_enable_optin_tracecontext()
 @pytest.mark.skip_library("dotnet", "Bug: Header search is currently case-sensitive")
 @pytest.mark.skip_library("golang", "not implemented")
-@pytest.mark.skip_library("nodejs", "not implemented")
+@pytest.mark.skip_library("nodejs", "nodejs lower-cases all header names")
```
zacharycmontoya, line 112: Would the test pass if we didn't skip it? If so, I'd like to run the test to demonstrate that these inputs have the correct output and we could leave a code comment about the nodejs behavior if you'd like

parametric/test_headers_tracestate_dd.py
```diff
@@ -433,6 +434,8 @@ def test_headers_tracestate_dd_propagate_propagatedtags(test_agent, test_library
     dd_items4 = tracestate4["dd"].split(";")
     assert "traceparent" in headers4
 
+    # TODO(stephen.belanger): not sure what the correct behaviour here is,
+    # but Node.js seems to produce an empty string rather than None.
     if headers4["x-datadog-tags"] is None:
```
zacharycmontoya, line 439: I think an empty `x-datadog-tags` is fine and should be treated as "no x-datadog-tags". How about we change to this?

```suggestion
    if headers4["x-datadog-tags"] is None or headers4["x-datadog-tags"] == "":
```

parametric/test_headers_tracestate_dd.py
```diff
@@ -131,6 +131,8 @@ def test_headers_tracestate_dd_propagate_samplingpriority(test_agent, test_libra
 
     # 4) Sampled = 1, tracestate[dd][s] <= 0
     # Result: SamplingPriority = 1
+    # TODO(stephen.belanger): Is this correct? -1 input through dd headers
+    # results in -1 out the other end but here expects -1 to become 1?
```
zacharycmontoya, line 135: Yes, we're prioritize the tracestate's sampled flag. So as long as it is 1, we will propagate our sampling priority with a value >= 1. If the incoming DD sampling priority was >=1, use the value. Otherwise, set to 1

parametric/test_headers_tracecontext.py
Qard: It does not because we don't lowercase it _ourselves_ and these "headers" passed through RPC are not _real_ headers so because of the test only _immitating_ HTTP we don't actually get the test headers in lowercase even though that's impossible in _real_ HTTP.

Qard: I made a change to the RPC service to simulate the lowercasing.

parametric/apps/nodejs/servicer.js
```diff
@@ -26,13 +26,25 @@ class Servicer {
             parent.origin = request.origin;
         }
 
+        const { http_headers } = request.http_headers || {};
+        // Node.js HTTP headers are automatically lower-cased, simulate that here.
+        const convertedHeaders = {};
+        for (const [key, value] of Object.entries(http_headers)) {
+            convertedHeaders[key.toLowerCase()] = value;
```
zacharycmontoya, line 33: Nice 👍🏼

PR #775, opened 2023-01-11
[graalvm] testing docker cache

robertomonteromiguel:
## Description

Optimize docker file in order to cache layers.
The cache will be in github registry

## Check list

- [ ] CI is passing


utils/build/build.sh
```diff
@@ -115,26 +115,31 @@ do
...
 
-        docker build \
+        docker buildx build \
             --progress=plain \
             ${DOCKER_PLATFORM_ARGS} \
             -f ${DOCKERFILE} \
             -t system_tests/weblog \
+            --cache-to type=registry,ref=${DOCKER_REGISTRY_CACHE_PATH}/${WEBLOG_VARIANT}:cache \
+            --cache-from type=registry,ref=${DOCKER_REGISTRY_CACHE_PATH}/${WEBLOG_VARIANT}:cache \
+            --output type=local,dest=localregistry/images/${WEBLOG_VARIANT}/${BUILD_DIR} \
```
pawelchcki, line 128: I think this step saves the whole image to local system. It might be quite slow.

A better option would be to either:
a) keep the tagged image in the Buildkit instance until its needed in docker
b) load the image everytime its modified by specifying `--load` [option](https://docs.docker.com/engine/reference/commandline/buildx_build/#load)

robertomonteromiguel: Yes, I tried it, but I allways get this error:

`WARNING: No output specified with docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load
overwrite_waf_rules.Dockerfile:1
--------------------
   1 | >>> FROM system_tests/weblog
   2 |     
   3 |     COPY ./binaries/waf_rule_set.json  /waf_rule_set.json
--------------------
ERROR: failed to solve: system_tests/weblog: pull access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed`

This errors raises when you try to extend/rebuild a previous built image.

Similar case to: https://github.com/docker/buildx/issues/301

pawelchcki:
Looks good.
QQ: Is there any impact on the performance of builds ?

robertomonteromiguel:
> 

I hadn't considered that it might be slower buildx. I am only generating the images for a single platform. I was thinking of applying the cache to the other builds gradually.

pawelchcki:
Chatted with @robertomonteromiguel - compared to master build times are imperceptibly similar

pawelchcki (approved):

PR #817, opened 2023-01-30
tests/test_identify.py: fix the user id propagation test

Julio-Guerra:
## Description

Following up on an APM customer support, we found an issue in the RFC where the `usr.id` tag actually must always be set when using the "set_user()" function, to let Datadog know it is originating from the SDK and not from the HTTP header propagation.

## Check list

- [x] CI is passing


cbeauchesne (changes requested):
To handle this use case, could you :

* rename the old test method to `test_XXX_legacy`, 
* write new method with the new version with `test_XXX`.
* On the new method, add `missing_feature` for every tracer.

When a given tracer will implement the new logic, they will need to add `@irrelevant(context.library >= "lang@N")` on the legacy method, and `@irrelevant(context.library < "lang@N")` on the new one (Where `N` is the version where the change has been made).

cbeauchesne (approved):

kurochan:
@Julio-Guerra I'm customer of reported this issue.
Thank you for your quick fix!

PR #829, opened 2023-02-01
[APPSEC-7986] Use numeric user IDs in login event tests

jkirsteins:
## Description

For handling custom events on the backend, the related user IDs need to be passed as strings.

Modifying the login success/failure tests to make sure this is covered.

## Check list

- [ ] CI is passing

### DIFF
--- a/utils/build/docker/php/common/user_login_failure_event.php
+++ b/utils/build/docker/php/common/user_login_failure_event.php
@@ -1,5 +1,6 @@
 <?php
-\datadog\appsec\track_user_login_failure_event('system_tests_user', true,
+// pass user ID as integer to test that it ends up in meta not metrics
+\datadog\appsec\track_user_login_failure_event(123456, true,
 [
     'metadata0' => 'value0',
     'metadata1' => 'value1',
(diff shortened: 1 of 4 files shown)


carlostomas-dd:
If you want to test the enrichment on custom events this is not enough. I mean, when I look for the `usr.id` I don't look for `meta.usr.id ` but for any custom event that contains `usr.id`, like `meta.appsec.events.users.login.success.usr.id` . So, for testing this problem won't be enough. I'm going to add a few comments on what we'd need

utils/build/docker/php/common/user_login_failure_event.php
```diff
@@ -1,5 +1,6 @@
 <?php
-\datadog\appsec\track_user_login_failure_event('system_tests_user', true,
+// pass user ID as integer to test that it ends up in meta not metrics
+\datadog\appsec\track_user_login_failure_event(123456, true,
 [
     'metadata0' => 'value0',
```
carlostomas-dd, line 5: If this is the data that are being sent in meta you'd need to add 'usr.id' => 123456. What I want is for this login to send inside meta this "appsec.events.users.login.failure.usr.id": 123456

jkirsteins:
Closing, since we've decided to do the remapping backend-side.

PR #846, opened 2023-02-06
IAST source test

jandro996:
## Description

Add IAST's Sources tests
Move current test (Sinks tests) to new sink package

## Check list

- [x] CI is passing


tests/appsec/iast/source/test_iast_request_cookie.py
```diff
@@ -0,0 +1,31 @@
...
+class TestRequestCookie:
+    """Verify that request cookies are tainted"""
+
+    def test_cookie(self):
+        weblog.get("/iast/source/cookie/test", cookies={"cookie-source-name": "cookie-source-value"})
+        interfaces.library_stdout.wait()
+        test_name = re.escape(
+            'tainted={"value":"cookie-source-name","ranges":[{"source":{"origin":"http.request.cookie.name","name":"cookie-source-name","value":"cookie-source-name"},"start":0,"length":18}]}'
+        )
+        interfaces.library_stdout.assert_presence(test_name, level="DEBUG")
```
smola, line 27: Actually testing this with logs forces every tracer to have this logging mechanism, and have it with the same format. I wonder if it's better to detect it using simple injection vulnerability?

I'm fine with merging this PR as it stands. But we might need to change this approach later to make it easier for other languages.

utils/build/docker/java/spring-boot/src/main/java/com/datadoghq/system_tests/springboot/AppSecIastSource.java
```diff
@@ -0,0 +1,77 @@
...
+            span.setTag("appsec.event", true);
+        }
+        final String source = request.getParameter("source");
+        final String value = request.getParameter("value");
+        return String.format("Request Parameters => source: %s, value: %s", source, value);
+    }
+
+    @PostMapping("/parametername/test")
+    String sourceParameterName(final ServletRequest request) {
+        final Span span = GlobalTracer.get().activeSpan();
```
smola, line 36: I don't think you need this?

jandro996: No need with this log approach, thanks!

tests/appsec/iast/source/test_iast_request_cookie.py
jandro996: Let me try, I think it could be done in a few hours 🙏  I have to: 

- Extend the iast system test functionality to deal with sources and not only with vulnerabilities 
- Add vulnerable code to each controller endpoint without reusing code to avoid the Overhead controller's filter

tests/appsec/iast/sink/test_iast_ldap_injection.py
```diff
@@ -11,7 +11,8 @@
 
 # Weblog are ok for nodejs/express4 and java/spring-boot
 @coverage.basic
-@released(dotnet="?", golang="?", java="?", nodejs="?", php_appsec="?", python="?", ruby="?")
+@released(dotnet="?", golang="?", nodejs="?", php_appsec="?", python="?", ruby="?")
+@released(java={"spring-boot": "1.7.0", "*": "?"})
```
smola, line 15: Can we enable it for other variants based on Spring Boot? (can be a separate PR if needed, since this is about sinks)

tests/appsec/iast/source/test_iast_request_body.py
```diff
@@ -0,0 +1,27 @@
...
+from utils import weblog, interfaces, context, coverage, released, missing_feature
+
+
+if context.library == "cpp":
+    pytestmark = pytest.mark.skip("not relevant")
+
+
+@coverage.basic
+@released(dotnet="?", golang="?", php_appsec="?", python="?", ruby="?")
+@released(java={"spring-boot": "1.7.0", "*": "?"})
```
smola, line 16: Can we enable it for some other variants?

smola (approved):

robertomonteromiguel (approved):
Just a suggestion, it would be nice to have tests that request the secure URL and check that it does not return any source. For example, the controller could do a decode to avoid injections, in such a case we would not have to generate any event/message/source.

tests/appsec/iast/sink/test_iast_ldap_injection.py
jandro996: I will create another PR fixing the Xpassed test for different springboot scenarios

tests/appsec/iast/source/test_iast_request_body.py
jandro996: I will create another PR fixing the Xpassed test for different springboot scenarios

jandro996:
> Just a suggestion, it would be nice to have tests that request the secure URL and check that it does not return any source. For example, the controller could do a decode to avoid injections, in such a case we would not have to generate any event/message/source.

Thanks! I will keep it in mind for future implementations. These tests aims to check if we are "tainting" the current implemented sources more than check the vulnerability, we use this approach to to facilitate implementation in different languages as @smola suggested [here](https://github.com/DataDog/system-tests/pull/846#discussion_r1098685636)

PR #848, opened 2023-02-06
Add SSS tests to cover a parent-child spans case

ygree:
## Description

A couple more tests to verify SSS when a parent/child span kept by SSS but the other is dropped by the tracer sampling rules.

[The corresponding fix PR for Java Tracer](https://github.com/DataDog/dd-trace-java/pull/4674)
incident-18687

I also ran `test_root_span_selected_by_sss014` for tracers other than Java. Go, .NET and Python all set `_sampling_priority_v1` to `-1` for the root span (should be `2` according to [the RFC](https://github.com/DataDog/architecture/tree/master/rfcs/apm/integrations/single-span-ingestion-control#when-tracers-dont-send-traces-to-the-agent)). **Update: this sampling priority assertion wasn't relevant as the Agent handles this properly anyway**
Also they don't set `_dd.p.dm` for the sampled span. **Update: single spans should have this tag in this case, the specification was modified to clarify this**
NodeJS: fails with `InactiveRpcError`

## Check list

- [ ] CI is passing


parametric/test_span_sampling.py
```diff
@@ -442,3 +442,99 @@ def test_multi_rule_independent_rate_limiters_sss013(test_agent, test_library):
...
+        {
+            "DD_SPAN_SAMPLING_RULES": json.dumps(
+                [{"service": "webserver", "name": "web.handler", "sample_rate": 1.0, "max_per_second": 50}]
+            ),
+            "DD_TRACE_SAMPLE_RATE": 0,
+        }
+    ],
+)
+def test_child_span_selected_by_sss015(test_agent, test_library):
+    """Single spans selected by SSS must be kept despite of its parent has been dropped.
```
katiehockman, line 511:
```suggestion
    """Single spans selected by SSS must be kept even if its parent has been dropped.
```

katiehockman (commented):
Thanks! Just some small nits

parametric/test_span_sampling.py
```diff
@@ -442,3 +442,99 @@ def test_multi_rule_independent_rate_limiters_sss013(test_agent, test_library):
...
+    assert root_span["metrics"].get(SINGLE_SPAN_SAMPLING_MAX_PER_SEC) == 50
+
+    # child span should be dropped by defined trace sampling rules
+    assert child_span["metrics"].get(SAMPLING_PRIORITY_KEY) == -1
+    assert child_span["metrics"].get(SINGLE_SPAN_SAMPLING_RATE) is None
+    assert child_span["metrics"].get(SINGLE_SPAN_SAMPLING_MECHANISM) is None
+    assert child_span["metrics"].get(SINGLE_SPAN_SAMPLING_MAX_PER_SEC) is None
+
+    # verify correctness of the Datadog tags for both spans
+    assert root_span["meta"].get("_dd.p.dm") == "-8"  # span sampling mechanism
```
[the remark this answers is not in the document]
ZStriker19: The original intent was to have something preceding the 8, I think a service name or something separated by a `-`. That was dropped but the dash was kept.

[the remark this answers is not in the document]
ygree: Yes, that's just a dash: "trace.tags[_dd.p.dm] = `-` + mechanism" [[RFC] Trace Tags Propagation in Distributed Traces](https://docs.google.com/document/d/1zeO6LGnvxk5XweObHAwJbK3SfK23z7jQzp7ozWJTa2A/edit#heading=h.e3afyw5az5sz)

parametric/test_span_sampling.py
```diff
@@ -442,3 +442,99 @@ def test_multi_rule_independent_rate_limiters_sss013(test_agent, test_library):
...
+@pytest.mark.skip_library("nodejs", "Not implemented")
+@pytest.mark.skip_library("python", "Not implemented")
+@pytest.mark.parametrize(
+    "library_env",
+    [
+        {
+            "DD_SPAN_SAMPLING_RULES": json.dumps(
+                [{"service": "webserver", "name": "web.request", "sample_rate": 1.0, "max_per_second": 50}]
+            ),
+            "DD_TRACE_SAMPLING_RULES": json.dumps([{"sample_rate": 0}]),
```
ZStriker19, line 458:
```suggestion
            "DD_TRACE_SAMPLE_RATE": 0,
```
Is there a reason to use `DD_TRACE_SAMPLNG_RULES` here instead? I think some tracers may not wildcard service when left out. Maybe best to stick with `DD_TRACE_SAMPLE_RATE` ?

parametric/test_span_sampling.py
```diff
@@ -442,3 +442,99 @@ def test_multi_rule_independent_rate_limiters_sss013(test_agent, test_library):
...
+            with test_library.start_span(name="web.handler", service="webserver", parent_id=parent_span.span_id):
+                pass
+
+    traces = test_agent.wait_for_num_traces(1, clear=True)
+
+    root_span = find_span_in_traces(traces, Span(name="web.request", service="webserver"))
+    child_span = find_span_in_traces(traces, Span(name="web.handler", service="webserver"))
+
+    # root span should be kept by defined the SSS rules
+    assert root_span["metrics"].get(SAMPLING_PRIORITY_KEY) == USER_KEEP
```
pierotibou, line 479: Is that true? The RFC doesn't say anything about changing the sampling priority when we don't have stats computations enabled.

ZStriker19: Agreed, the Python tracer passes everything except for these `sampling_priority_v1` assertions. Since stats computation is not turned on here, it sees like this is not the correct behavior: https://github.com/DataDog/architecture/blob/master/rfcs/apm/integrations/single-span-ingestion-control/README.md#when-tracers-dont-send-traces-to-the-agent

`test_keep_span_with_stats_computation_sss010` tests the behavior I would expect.

dgoffredo: The only sampling-priority-relevant check in these new tests ought to be that the root span priority is _not_ "keep" when span sampling is happening.

I think that this line can be removed.

katiehockman: This line was already removed in the current commit of this PR, right?

parametric/test_span_sampling.py
ygree: There is no specific reason for that, DD_TRACE_SAMPLE_RATE should be just fine for this test purposes

parametric/test_span_sampling.py
```diff
@@ -442,3 +442,88 @@ def test_multi_rule_independent_rate_limiters_sss013(test_agent, test_library):
...
+
+    # root span should be kept by defined the SSS rules
+    # don't check SAMPLING_PRIORITY_KEY of the parent since it could be set to drop,
+    # but it isn't a requirement
+    assert parent_span["metrics"].get(SINGLE_SPAN_SAMPLING_RATE) == 1.0
+    assert parent_span["metrics"].get(SINGLE_SPAN_SAMPLING_MECHANISM) == SINGLE_SPAN_SAMPLING_MECHANISM_VALUE
+    assert parent_span["metrics"].get(SINGLE_SPAN_SAMPLING_MAX_PER_SEC) == 50
+
+    # child span should be dropped by defined trace sampling rules
+    assert child_span["metrics"].get(SAMPLING_PRIORITY_KEY) == -1
```
Qard, line 483: Is this assertion correct? I see no reference to sampling priority key in the RFC other than setting to 2 on user-keep. This is the single failure remaining for Node.js. If I comment this out the rest pass.

ygree: If we remove this assertion then Java Tracer prior to the fix will pass this test.

Qard: @dgoffredo ^

dgoffredo: I think that the correct assertion would be:
```python
assert parent_span["metrics"].get(SAMPLING_PRIORITY_KEY) == -1
```
Sampling priority is about the trace. We convey the sampling priority of the trace (chunk) by setting the sampling priority on the local root span.

What we want to ensure is that the trace is being dropped despite the span sampling at play. So, we should check the sampling priority of the local root span, and make sure it indicates "drop." The Agent will then keep any spans that have `_dd.span_sampling.mechanism` set.

dgoffredo: Ref: https://github.com/DataDog/datadog-agent/blob/0b2ce0478d726be5642a666e75dada49403c0ccd/pkg/trace/agent/agent.go#L260-L262

ygree: @dgoffredo I think you're right! I'm going to clean up this part in Java Tracer. Thanks for explanation!

ygree: Cleaned this logic in Java Tracer in [this PR](https://github.com/DataDog/dd-trace-java/pull/4695)

Qard (approved):

Qard:
Looks like the linter has some issues, but other than that LGTM.

katiehockman (approved):

Kyle-Verhoog (approved):
lgtm from a parametric/code quality standpoint

Qard (approved):

robertomonteromiguel (approved):

dgoffredo:
I'll propose additions to this test in a separate PR.

This LGTM.

Qard:
Seems to be a bunch of dotnet and java failures, but doesn't look to be related. Can we land this?

cbeauchesne:
> Seems to be a bunch of dotnet and java failures, but doesn't look to be related. Can we land this?

Yes, you can ignore them

PR #871, opened 2023-02-13
[apm][single spans] Add an e2e test for APM Single Spans Ingestion Control

lambrospetrou:
## Description

- Internal Jira: `ATI-2419`

This change adds e2e tests for the [Single Spans Ingestion Control](https://docs.datadoghq.com/tracing/trace_pipeline/ingestion_mechanisms/?tab=java#single-spans) feature.

The new test cases use the following configuration:
```
DD_SPAN_SAMPLING_RULES='[{\"service\": \"weblog\", \"name\": \"*single_span_submitted\", \"sample_rate\": 1.0, \"max_per_second\": 50}]'"
DD_TRACE_SAMPLE_RATE=0"
```

The scenarios we want to cover are:
1. The parent span (root span) to be submitted, but not the child span.
2. The child span to be submitted, but not the parent span (root span).

## Application runs

Since Single Spans only works with the Datadog agent 7.40, I set the following to `./system-tests/binaries/agent-image` when testing lcoally.
```
datadog/agent:7.42.0
```

### Go - chi

```sh
./build.sh -l golang -w chi
./run.sh APM_TRACING_E2E_SINGLE_SPAN
```

## Tips to the reviewers

- The test logic is in `tests/apm_tracing_e2e/test_single_span.py`
- The agent configuration is in `run.sh`
- The rest are mostly implementation details how to fetch traces and spans from the Trace API and the Event Platform API.

## Check list

- [ ] CI is passing


dumontg (approved):
Looks good! Im unfamiliar with the codebase and found this PR easy to read.

utils/interfaces/_backend.py
```diff
@@ -23,28 +26,99 @@ def __init__(self):
...
+                self.rid_to_library_trace_ids[rid] = [span["trace_id"]]
+            else:
+                self.rid_to_library_trace_ids[rid].append(span["trace_id"])
+
+    #################################
+    ######### API for tests #########
+    #################################
+
+    def assert_library_traces_exist(self, request, min_traces_len=1):
+        """Attempts to fetch all the traces that the library tracers sent to the agent.
```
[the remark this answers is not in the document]
lambrospetrou: You understood correctly. Basically here we assert that all the traces sent by the tracers to the agent, are also submitted by the agent to the backend, are processed correctly, and are available to fetch from the backend. So the traditional e2e.

I will reword it to be clearer.

utils/interfaces/_agent.py
```diff
@@ -107,3 +107,17 @@ def add_traces_validation(self, validator, success_by_default=False):
         self.validate(
             validator=validator, success_by_default=success_by_default, path_filters=r"/api/v0\.[1-9]+/traces"
         )
+
+    def get_spans_for_request(self, request):
```
cbeauchesne, line 111: could you use the same prototype as in library interface : 

```
def get_spans(self, request=None):
```

* if request is none, iterate over all spans
* otherwise, iterate only on spans triggered by the request.

Tnaks !

cbeauchesne (changes requested):

utils/interfaces/_agent.py
lambrospetrou: I can do it yes, although I don't fully agree it's better.

I actually had that version in a previous version. In my opinion, it's a bit confusing right now what actually happens whenever the `request` is passed in a function, and what is being returned in each case. I basically always had to open the code of every function to see what it does, since naming and docs are a bit lacking.

For example, initially, I didn't even notice that the `get_spans` is inside the base `InterfaceValidator`. This was surprising to me since there are 3 places where you can get spans from (tracer->agent, agent->backend, backend->user), so having the agent traces in every interface could lead to bugs. I didn't notice this until the PR CI ran all the tests.

Probably these are issues due to using untyped Python as well, so some better documentation would help in providing clear expectations when calling these helpers.

lambrospetrou: Done.

cbeauchesne: `get_spans` in InterfaceValidator is indeed an error, and I'm almost sure it's not used ^^ I will check and fix that-> #876 (some interface does not have the knowledge of what is a "span", so there is no point to have it here).

For namings, totally open to suggestion, as long as the code is coherent. I'll at least add some documentations

cbeauchesne (approved):

PR #873, opened 2023-02-14
parametric: Expand parametric testing suits to test OTel API

dianashevchenko:
## Description
This PR:
- extends the proto and testing API to run tests on Otel API wrappers of Datadog tracers.

## Check list

- [ ] CI is passing


parametric/protos/apm_test_client.proto
```diff
@@ -10,6 +10,18 @@ service APMClient {
...
+
+  rpc OtelStartSpan(OtelStartSpanArgs) returns (OtelStartSpanReturn) {}
+  rpc OtelEndSpan(OtelEndSpanArgs) returns (OtelEndSpanReturn) {}
+  rpc OtelIsRecording(OtelIsRecordingArgs) returns (OtelIsRecordingReturn) {}
+  rpc OtelSpanContext(OtelSpanContextArgs) returns (OtelSpanContextReturn) {}
+  rpc OtelSetStatus(OtelSetStatusArgs) returns (OtelSetStatusReturn) {}
+  rpc OtelSetName(OtelSetNameArgs) returns (OtelSetNameReturn) {}
+  rpc OtelSetAttributes(OtelSetAttributesArgs) returns (OtelSetAttributesReturn) {}
+  rpc OtelFlushSpans(OtelFlushSpansArgs) returns (OtelFlushSpansReturn) {}
+  rpc OtelFlushTraceStats(OtelFlushTraceStatsArgs) returns (OtelFlushTraceStatsReturn) {}
```
PerfectSlayer, line 22: It feels like the original service is duplicate. Why don't create another service?

PerfectSlayer, line 22: Are flush calls specific OTel?

parametric/protos/apm_test_client.proto
```diff
@@ -76,5 +88,76 @@ message FlushSpansReturn {}
...
+  int32 trace_flags = 3;
+  string trace_state = 4;
+  bool remote = 5;
+}
+
+
 message StopTracerArgs {}
 message StopTracerReturn {}
+
+message StartTracerArgs {}
```
PerfectSlayer, line 162: Should we expect future arguments?

parametric/protos/apm_test_client.proto
```diff
@@ -10,6 +10,18 @@ service APMClient {
...
+  rpc OtelEndSpan(OtelEndSpanArgs) returns (OtelEndSpanReturn) {}
+  rpc OtelIsRecording(OtelIsRecordingArgs) returns (OtelIsRecordingReturn) {}
+  rpc OtelSpanContext(OtelSpanContextArgs) returns (OtelSpanContextReturn) {}
+  rpc OtelSetStatus(OtelSetStatusArgs) returns (OtelSetStatusReturn) {}
+  rpc OtelSetName(OtelSetNameArgs) returns (OtelSetNameReturn) {}
+  rpc OtelSetAttributes(OtelSetAttributesArgs) returns (OtelSetAttributesReturn) {}
+  rpc OtelFlushSpans(OtelFlushSpansArgs) returns (OtelFlushSpansReturn) {}
+  rpc OtelFlushTraceStats(OtelFlushTraceStatsArgs) returns (OtelFlushTraceStatsReturn) {}
+
+  rpc StartTracer(StartTracerArgs) returns (StartTracerReturn) {}
```
PerfectSlayer, line 24: Should we expect this call to be called multiple times and provide support for multiple tracer instances?

parametric/protos/apm_test_client.proto
dianashevchenko: It started as a separate service but was advised against to reduce area of code support. Alternatively, I could create an almost  duplicate of APMCLientStub to split the two and pass in an analogous test_otel_library

dianashevchenko: Flush method is explicit in [the api](https://opentelemetry.io/docs/reference/specification/trace/sdk/#forceflush)

parametric/protos/apm_test_client.proto
dianashevchenko: Yes, that would be configuration of the tracer/provider with environment / service or other Datadog supported tracer options

parametric/protos/apm_test_client.proto
PerfectSlayer: It is for spans. So does the traces stats flush has to be OTel specific?

parametric/protos/apm_test_client.proto
PerfectSlayer: It could be useful to document it in the service definition that tracer initialization will happen once (and only once, and always happens, etc...).

parametric/protos/apm_test_client.proto
mabdinur: Since Force Flushing is defined in the [otel sdk](https://opentelemetry.io/docs/reference/specification/trace/sdk/#forceflush) and not the [otel api](https://opentelemetry.io/docs/reference/specification/trace/api/#tracerprovider), the python implementation does not expose an api for flushing otel traces. This functionality is hidden in a private method but can be accessed by tests. I believe the only TracerProvider operation  we need to support is ‘getTracer()’ and the otel tracer only needs to support creating/activating spans. 

Is forced flushing an operation we would like to expose to otel users or is it something that’s only required in tests?

parametric/protos/apm_test_client.proto
```diff
@@ -76,5 +88,76 @@ message FlushSpansReturn {}
...
 message FlushTraceStatsReturn {}
 
+
+message OtelStartSpanArgs {
+  string name = 1;
+  optional bool new_root = 2;
+  optional uint64 parent_id = 3;
+  optional uint64 span_kind = 9;
+  optional string service = 4;
+  optional string resource = 5;
```
mabdinur, line 98: Question: how are resource names set in the go implementation? In python otelspan.name == otelspan.resource. Should the otel span object have GetResource/SetResource methods?

parametric/protos/apm_test_client.proto
```diff
@@ -10,6 +10,18 @@ service APMClient {
   rpc InjectHeaders(InjectHeadersArgs) returns (InjectHeadersReturn) {}
   rpc FlushSpans(FlushSpansArgs) returns (FlushSpansReturn) {}
   rpc FlushTraceStats(FlushTraceStatsArgs) returns (FlushTraceStatsReturn) {}
+
+  rpc OtelStartSpan(OtelStartSpanArgs) returns (OtelStartSpanReturn) {}
```
mabdinur, line 14: The python api defines two methods for starting spans: start_span and start_as current span: https://opentelemetry-python.readthedocs.io/en/latest/api/trace.html#opentelemetry.trace.Tracer. If this distinction exists in other languages we should add an rpc for ‘startCurrentSpan’

parametric/protos/apm_test_client.proto
```diff
@@ -10,6 +10,18 @@ service APMClient {
   rpc InjectHeaders(InjectHeadersArgs) returns (InjectHeadersReturn) {}
   rpc FlushSpans(FlushSpansArgs) returns (FlushSpansReturn) {}
   rpc FlushTraceStats(FlushTraceStatsArgs) returns (FlushTraceStatsReturn) {}
+
```
mabdinur, line 13: Should we add an rpc to set error messages on an otel span?

parametric/spec/otel_trace.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+class OtelSpanContext(TypedDict):
+    trace_id: bytes
+
+
+Trace = List[OtelSpan]
+
+
+def find_otel_span_in_traces(traces: List[Trace], span: OtelSpan) -> OtelSpan:
```
mabdinur, line 17: Instead of using this method to find otel spans, can we set a unique tag on all otel spans generated by system tests or use a unique naming convention for otel spans. This check could become flaky and hard to maintain.

parametric/test_otel_tracer.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+def test_start_otel_api_span_with_set_name(test_agent, test_library):
+    """Test span is created with OTel API of Datadog Tracing Library"""
+    with test_library:
+        with test_library.start_otel_span(name="operation", service="my-webserver", resource="/endpoint",) as parent:
+            parent.set_attributes({"key": "val"})
+            parent.set_name("new_op")
+
+    span = find_otel_span_in_traces(test_agent.wait_for_num_traces(1), OtelSpan(name="web.otel"))
```
mabdinur, line 11: Would this work? 
```suggestion
    span = test_agent.wait_for_num_traces(1)[0]
    assert span.get_name() == “new_op”
```

dianashevchenko: I have updated this to `span = get_span(test_agent)`

parametric/spec/otel_trace.py
dianashevchenko: I will remove this in favor of `span = get_span(test_agent)`

parametric/protos/apm_test_client.proto
PerfectSlayer: It does not in Java. Creating a span starts it, then you can call `makeCurrent()` on the instance.

PR #881, opened 2023-02-17
Christophe papazian/waf rule update

christophe-papazian:
## Description

A specific request can be detected by several rules. The test was making a wrong assumption about which rule would detect the threat first.


gnufede:
It would be great to have the ability to match against a tuple, so if any of the rules in the tuple match, the test passes. Maybe in a next PR ;)

PR #889, opened 2023-02-21
Enable ruby test for test blocking addresses already implemented

GustavoCaso:
## Description

*Please include a short summary of the change*

Enable ruby test for test blocking addresses already implemented

## Workflow

1. ⚠️⚠️ Create your PR as draft (we're receiving lot of PR, it saves us lot of time) ⚠️⚠️
2. Follows the style guidelines of this project (See [how to easily lint the code](https://github.com/DataDog/system-tests/blob/main/docs/edit/lint.md))

Once your PR is reviewed, you can merge it ! :heart:


tests/appsec/waf/test_addresses.py
```diff
@@ -370,11 +370,12 @@ def test_json_array(self):
         interfaces.library.assert_waf_attack(self.r_array, value='<vmlframe src="xss">', address="server.request.body")
 
 
-@released(golang="1.37.0", dotnet="2.8.0", nodejs="2.2.0", php="?", python=PYTHON_RELEASE_GA_1_1, ruby="?")
+@released(golang="1.37.0", dotnet="2.8.0", nodejs="2.2.0", php="?", python=PYTHON_RELEASE_GA_1_1)
 @released(java={"vertx3": "?", "ratpack": "0.99.0", "*": "0.95.1"})
 @bug(context.library == "nodejs@2.8.0", reason="Capability to read body content is broken")
 @missing_feature(context.weblog_variant == "spring-boot-native", reason="GraalVM. Tracing support only")
 @missing_feature(context.weblog_variant == "spring-boot-3-native", reason="GraalVM. Tracing support only")
+@irrelevant(reason="unsupported by framework", library="ruby")
```
lloeki, line 378: IIRC Rails supports params from XML body. Worth a check if this test should apply.

tests/appsec/test_blocking_addresses.py
```diff
@@ -27,6 +27,7 @@ class Test_BlockingAddresses:
     def setup_request_method(self):
         self.rm_req = weblog.request("OPTIONS")
 
+    @released(ruby="?")
```
lloeki, line 30: Ultimately that one should work, we just miss the address for Rack, right?

IIRC there's the case where a verb can be turned into behaving like another with some Rails parameter though (something like having a parameter named `_method` or something) for back when AJAX could only handle `GET` and `POST` but one wanted RESTful `PUT`, `DELETE`, `PATCH`. Not sure if it's still possible in recent Rails given that browsers now all support these cases proper.

Also not sure WAF+rules are prepared to support that.

tests/appsec/test_blocking_addresses.py
```diff
@@ -114,6 +123,7 @@ def setup_response_header(self):
         self.rsh_req = weblog.get(path="/headers")
 
     @missing_feature(context.library == "java", reason="Happens on a subsequent WAF run")
+    @released(ruby="?")
     def test_response_header(self):
         """can block on server.response.headers.no_cookies"""
```
lloeki, line 128: I think we ultimately can have that one as well, we're just missing the address for Rack.

lloeki (approved):
LGTM!

tests/appsec/waf/test_addresses.py
GustavoCaso: I tested it already, and for Rails doesn't work unless we change the way we get the raw body: `request.raw_body`

By default, rails only provide a parser for JSON.

https://github.com/rails/rails/blob/main/actionpack/lib/action_dispatch/http/parameters.rb#L10-L15

cbeauchesne (changes requested):

cbeauchesne (approved):
Don't mind about the failure on php/python/node, it's not related.

PR #911, opened 2023-02-27
Aditi.maheshwari/telemetry dependency flag test

aditim16:
## Description

system test cases for DD_TELEMETRY_DEPENDENCY_COLLECTION_ENABLED and Telemetry Message Data Limits


tests/test_telemetry.py
```diff
@@ -244,6 +278,9 @@ def validator(data):
 
         self.validate_library_telemetry_data(validator)
 
+    def setup_app_heartbeat(self):
+        time.sleep(20)
```
cbeauchesne, line 282: Not the good way. You need at least 20 seconds of lifespan for the app, right? (IIRC, the heartbeat delay is set to 2 seconds)

aditim16: @mehulsonowal addressed this in his PR

PR #957, opened 2023-03-13
Add parametric tests of the 'Supporting 128-Bit TraceIds' RFC

dinooliva:
## Description

Adds parametric tests for 128-bit TraceId generation and propagation as specified in the 'Supporting 128-Bit TraceIds' RFC. In particular:

- test propagation of external 128-bit TraceIds by all propagators (datadog, b3, b3 multi, w3c)
- test generation and propagation of 64-bit TraceIds in 64-bit TraceId generation (default) mode
- test generation and propagation of 128-bit TraceIds in 128-bit TraceId generation mode


dinooliva:
I have tested this against our golang 128-Bit TraceId support branch (and found some issue with it). Code is ready to start the review process.

parametric/test_128_bit_traceids.py
```diff
@@ -0,0 +1,310 @@
+import pytest
+
+from parametric.utils.headers import make_single_request_and_get_inject_headers
+from parametric.utils.test_agent import get_span
+
+parametrize = pytest.mark.parametrize
+
+
+@pytest.mark.parametrize(
+    "library_env", [{"DD_TRACE_PROPAGATION_STYLE": "Datadog", "DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED": False,}],
```
zacharycmontoya, line 10: I think you will want both the key and value to be strings:
```suggestion
    "library_env", [{"DD_TRACE_PROPAGATION_STYLE": "Datadog", "DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED": "false",}],
```

parametric/test_128_bit_traceids.py
```diff
@@ -0,0 +1,310 @@
...
+    """
+    with test_library:
+        headers = make_single_request_and_get_inject_headers(test_library, [])
+
+    header_trace_id = headers["x-datadog-trace-id"]
+    span = get_span(test_agent)
+    span_trace_id = span.get("trace_id")
+    dd_p_tid = span["meta"].get("_dd.p.tid")
+
+    assert len(header_trace_id) < 21
```
zacharycmontoya, line 50: It wasn't immediately clear to me that this was testing that the value was less than 2^64-1 (i.e. 18,446,744,073,709,551,615), so maybe a code comment would help make that clear

parametric/test_128_bit_traceids.py
```diff
@@ -0,0 +1,310 @@
...
+@pytest.mark.skip_library("dotnet", "not implemented")
+@pytest.mark.skip_library("golang", "not implemented")
+@pytest.mark.skip_library("java", "not implemented")
+@pytest.mark.skip_library("nodejs", "not implemented")
+@pytest.mark.skip_library("python", "not implemented")
+@pytest.mark.skip_library("python_http", "not implemented")
+@pytest.mark.parametrize(
+    "library_env", [{"DD_TRACE_PROPAGATION_STYLE": "Datadog", "DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED": True,}],
+)
+def test_datadog_generation_128_bit_D003(test_agent, test_library):
```
zacharycmontoya, line 64: non-blocking nit: I was thinking about a slight rename of the tests that makes the intent a little clearer to me. What do you think?
- `test_<format>_128_bit_propagation_<TEST#>`
- `test_<format>_128_bit_generation_enabled_<TEST#>`
- `test_<format>_128_bit_generation_disabled_<TEST#>`

zacharycmontoya (commented):
The overall approach of the tests look good to me. I'm hesitant to Approve since I haven't spent much time with the RFC itself so I'm not super familiar with the nuances of the implementation. I'd rather defer to the people implementing the RFC review to Approve the tests

parametric/test_128_bit_traceids.py
dinooliva: Added a comment to clarify.

parametric/test_128_bit_traceids.py
dinooliva: Fixed all instances.

parametric/test_128_bit_traceids.py
```diff
@@ -0,0 +1,329 @@
...
+        headers = make_single_request_and_get_inject_headers(test_library, [])
+
+    header_trace_id = headers["x-datadog-trace-id"]
+    span = get_span(test_agent)
+    span_trace_id = span.get("trace_id")
+    dd_p_tid = span["meta"].get("_dd.p.tid")
+
+    # Note: length of 2^64-1 encoded as decimal is 20.
+    assert len(header_trace_id) < 21
+    assert int(header_trace_id, 10) == span_trace_id
```
lucaspimentel, line 52: I'm not familiar with Python type handling. Is this conversion enough to ensure that the trace id is less than 2^64-1? What happens, for example, if the value is 2^64+1 (which will still pass the `len(header_trace_id) < 21` test above)?

parametric/test_128_bit_traceids.py
```diff
@@ -0,0 +1,329 @@
...
+def test_b3single_128_bit_generation_disabled(test_agent, test_library):
+    """Ensure that 64-bit TraceIds are properly generated, propagated in B3
+    single-header, and populated in trace data.
+    """
+    with test_library:
+        headers = make_single_request_and_get_inject_headers(test_library, [])
+    span = get_span(test_agent)
+    b3 = headers["b3"]
+
+    assert len(b3) > 16 + 16
```
lucaspimentel, line 129: This could be stricter and check for length == 16+16+3 (or 32+16+3 if we allow 64-bit trace ids to be zero-padded to 32 characters).

parametric/test_128_bit_traceids.py
```diff
@@ -0,0 +1,329 @@
...
+            test_library, [["b3", "640cfd8d00000000abcdefab12345678-000000003ade68b1-1"],],
+        )
+    span = get_span(test_agent)
+    trace_id = span.get("trace_id")
+    dd_p_tid = span["meta"].get("_dd.p.tid")
+    b3 = headers["b3"]
+
+    assert trace_id == int("abcdefab12345678", 16)
+    assert dd_p_tid == "640cfd8d00000000"
+    assert len(b3) > 32 + 16
```
lucaspimentel, line 109: Should we check for the exact length?
```suggestion
    assert len(b3) == 32 + 16 + 3
```

parametric/test_128_bit_traceids.py
```diff
@@ -0,0 +1,329 @@
...
+    assert len(traceparent) == 55
+    check_128_bit_trace_id(traceparent[3:35], span.get("trace_id"), span["meta"].get("_dd.p.tid"))
+
+
+ZERO8 = "00000000"
+ZERO16 = ZERO8 + ZERO8
+
+
+def check_64_bit_trace_id(header_trace_id, span_trace_id, dd_p_tid):
+    """Ensure that 128-bit TraceIds are properly formatted and populated in
```
lucaspimentel, line 302:
```suggestion
    """Ensure that 64-bit TraceIds are properly formatted and populated in
```

parametric/test_128_bit_traceids.py
dinooliva: The integers in Python3 are 'unlimited' in size - I've changed the code to make sure that the value is in the correct range rather than just checking length.

parametric/test_128_bit_traceids.py
dinooliva: I added some of the length tests for basic sanity checking but they are redundant wrt the existing propagation tests. I've updated these tests to verify the fields more precisely now rather than just checking length.

parametric/test_128_bit_traceids.py
```diff
@@ -0,0 +1,329 @@
...
+def test_b3single_128_bit_generation_enabled(test_agent, test_library):
+    """Ensure that 128-bit TraceIds are properly generated, propagated in B3
+    single-header, and populated in trace data.
+    """
+    with test_library:
+        headers = make_single_request_and_get_inject_headers(test_library, [])
+    span = get_span(test_agent)
+    b3 = headers["b3"]
+
+    assert len(b3) > 32 + 16
```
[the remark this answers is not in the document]
dinooliva: As mentioned, I've removed the length checks now.

parametric/test_128_bit_traceids.py
dinooliva: Done (thanks).

parametric/test_128_bit_traceids.py
dinooliva: Similarly to before, I've removed the length checks.

parametric/test_128_bit_traceids.py
lucaspimentel: Thanks!

parametric/test_128_bit_traceids.py
```diff
@@ -0,0 +1,329 @@
...
+    """Ensure that 64-bit TraceIds are properly generated, propagated in B3
+    single-header, and populated in trace data.
+    """
+    with test_library:
+        headers = make_single_request_and_get_inject_headers(test_library, [])
+    span = get_span(test_agent)
+    b3 = headers["b3"]
+
+    assert len(b3) > 16 + 16
+    check_64_bit_trace_id(b3[0:16], span.get("trace_id"), span["meta"].get("_dd.p.tid"))
```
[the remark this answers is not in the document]
lucaspimentel: Thank you

Kyle-Verhoog (approved):

zacharycmontoya (approved):
LGTM

PR #1006, opened 2023-03-28
Parametric: Add more tests for 128-bit TraceId support.

dinooliva:
## Description
Add more tests for 128-bit TraceId support:
1. Verify that 64-bit TraceIds are propagated properly when 128-bit TraceId generation is enabled.
2. Verify that malformed `_dd.p.tid` tags are dropped.
3. Verify that `_dd.p.tid` tags propagated via w3c are consistent with the propagated TraceId otherwise that the TraceId overrides the tag value.


parametric/test_128_bit_traceids.py
```diff
@@ -28,9 +28,151 @@ def test_datadog_128_bit_propagation(test_agent, test_library):
...
+    with test_library:
+        make_single_request_and_get_inject_headers(
+            test_library,
+            [
+                ["x-datadog-trace-id", "1234567890123456789"],
+                ["x-datadog-parent-id", "987654321"],
+                ["x-datadog-tags", "_dd.p.tid=XXXX"],
+            ],
+        )
+    assert get_span(test_agent)["meta"].get("_dd.propagation_error") == "malformed_tid XXXX"
```
lucaspimentel, line 154: I think there are two issue with this assertion:
- Adding this tag is optional. This test should allow for no `_dd.propagation_error` tag.
- The `_dd.propagation_error` tag can be used for other problems in `x-datadog-tags`. Valid values are:
  - `extract_max_size`
  - `inject_max_size`
  - `disabled`
  - `encoding_error`
  - `decoding_error`

[source](https://docs.google.com/document/d/1zeO6LGnvxk5XweObHAwJbK3SfK23z7jQzp7ozWJTa2A/edit#heading=h.j993k37q7jo4) (internal)

rochdev: Same for other places where the tag is checked in this PR.

parametric/test_128_bit_traceids.py
```diff
@@ -246,6 +430,174 @@ def test_w3c_128_bit_propagation(test_agent, test_library):
...
+    """
+    with test_library:
+        make_single_request_and_get_inject_headers(
+            test_library,
+            [
+                ["traceparent", "00-640cfd8d00000000abcdefab12345678-000000003ade68b1-01"],
+                ["tracestate", "dd=t.tid:640cfd8d0000ffff"],
+            ],
+        )
+    assert get_span(test_agent)["meta"].get("_dd.propagation_error") == "inconsistent_tid 640cfd8d0000ffff"
```
lucaspimentel, line 517: Similar to a previous comment, adding `_dd.propagation_error` is optional.

parametric/test_128_bit_traceids.py
```diff
@@ -246,6 +430,174 @@ def test_w3c_128_bit_propagation(test_agent, test_library):
...
+    """
+    with test_library:
+        make_single_request_and_get_inject_headers(
+            test_library,
+            [
+                ["traceparent", "00-640cfd8d00000000abcdefab12345678-000000003ade68b1-01"],
+                ["tracestate", "dd=t.tid:XXXX"],
+            ],
+        )
+    assert get_span(test_agent)["meta"].get("_dd.propagation_error") == "malformed_tid XXXX"
```
lucaspimentel, line 575: Optional `_dd.propagation_error` tag

lucaspimentel (commented):
A lot of good tests! My only comment is that the RFC says `_dd.propagation_error` is optional so the tests should allow for the tag to not exist on the span.

parametric/test_128_bit_traceids.py
lucaspimentel: I talked to Dino and he explained the idea is that only languages that support this will enable the test and other can skip it as "not supported." Sounds good to me, probably a better way of tracking which languages support an optional feature.

Thanks, Dino.

lucaspimentel (approved):

Kyle-Verhoog (approved):

PR #1041, opened 2023-04-04
Mtoff/header tags tests

mtoffl01:
## Description

Additional system tests to validate how tracing libraries handle a variety of cases for header_tags feature, such as whitespaces and leading/trailing colons


mtoffl01:
@cbeauchesne , yes! It's waiting on the next release of dd-trace-go

manifests/golang.yml
```diff
@@ -42,7 +42,7 @@ tests/:
           TestXPathInjection: missing_feature
         test_xss.py:
           TestXSS:
-            '*': missing_feature
+            "*": missing_feature
```
cbeauchesne, line 45: Could you put back single quotes ?

tests/test_library_conf.py
```diff
@@ -2,12 +2,12 @@
...
 # Copyright 2021 Datadog, Inc.
 
+import re
 from utils import coverage, weblog, interfaces, released, irrelevant, scenarios
 
-
 # basic / legacy tests, just tests user-agent can be received as a tag
 @irrelevant(library="cpp")
-@released(java="?", php="0.68.2", python="0.53")
+@released(java="?", php="0.68.2", python="0.53", golang="1.53.0")
```
cbeauchesne, line 10: it must be declared only once, as you did it in manifest, you can remove this one, same for other `@released`

tests/test_library_conf.py
```diff
@@ -62,5 +62,121 @@ def test_trace_header_tags(self):
...
+        self.r = weblog.get("/waf", headers=self.headers)
+
+    def test_trace_header_tags(self):
+        tag_conf = scenarios.library_conf_custom_headers_whitespacing_headers.weblog_container.environment[
+            "DD_TRACE_HEADER_TAGS"
+        ]
+
+        full_tag_config_list = tag_conf.split(",")
+        # skip the first item, as this required to make the tests work on some platforms
+        tag_config_list = full_tag_config_list[1::]
```
cbeauchesne, line 87: The second double comma is useless (same for above). 

```suggestion
        tag_config_list = full_tag_config_list[1:]
```

And could you comment why you remove the first item ? 
And also, as this code is the same for all tests, can you make a function `get_header_tag_list(scenario)` ?

tests/test_library_conf.py
```diff
@@ -62,5 +62,121 @@ def test_trace_header_tags(self):
...
+
+        tags = {}
+        for tag in tag_config_list:
+            tag = tag.strip()
+            tags["http.request.headers." + tag] = re.sub("[^a-zA-Z0-9 -]", "_", self.headers[tag])
+
+        interfaces.library.add_span_tag_validation(request=self.r, tags=tags)
+
+
+# Note: dotnet fails this test. It normalizes those spaces to underscores instead.
```
cbeauchesne, line 97: so you must decorate the class with `@bug(library="dotnet", reason="It normalizes those spaces to underscores instead.")`

Cherry on the cake, the Jira ticket as the reason

utils/_context/_scenarios.py
```diff
@@ -1100,14 +1100,32 @@ class scenarios:
...
+    # dummy value because the tests skip the first input.
     library_conf_custom_headers_short = EndToEndScenario(
-        "LIBRARY_CONF_CUSTOM_HEADERS_SHORT", additional_trace_header_tags=("header-tag1", "header-tag2"), doc=""
+        "LIBRARY_CONF_CUSTOM_HEADERS_SHORT", additional_trace_header_tags=("dummy", "header-tag1", "header-tag2"), doc="Scenario with <header_name> input given to DD_TRACE_HEADER_TAGS"
     )
     library_conf_custom_headers_long = EndToEndScenario(
         "LIBRARY_CONF_CUSTOM_HEADERS_LONG",
         additional_trace_header_tags=("header-tag1:custom.header-tag1", "header-tag2:custom.header-tag2"),
-        doc="",
+        doc="Scenario with <header>:<tag_name> formatted input given to DD_TRACE_HEADER_TAGS",
```
cbeauchesne, line 1110: Scenario are the most expensive stuff to add on system tests. Would it be doable to merge some of them ?

utils/interfaces/_library/core.py
```diff
@@ -10,7 +10,18 @@
 from utils.interfaces._core import ProxyBasedInterfaceValidator
 from utils.interfaces._library._utils import get_trace_request_path
 from utils.interfaces._library.appsec import _WafAttack, _ReportedHeader
+<<<<<<< HEAD
```
cbeauchesne, line 13: Commit error ;)

utils/interfaces/_library/miscs.py
```diff
@@ -32,3 +32,19 @@ def __call__(self, span):
                     raise ValueError(f'{tagKey} tag in span\'s meta should be "{expectValue}", not "{actualValue}"')
 
         return True
+
+
+class _NotSpanTagValidator:
```
cbeauchesne, line 37: This is the legacy way of doing validation, I'll adapt it with the new way once everything is ok.

cbeauchesne (changes requested):

tests/test_library_conf.py
```diff
@@ -62,5 +62,121 @@ def test_trace_header_tags(self):
         tag_config_list = full_tag_config_list[1::]
 
         tags = {item.split(":")[1]: self.headers[item.split(":")[0]] for item in tag_config_list}
+        interfaces.library.add_span_tag_validation(request=self.r, tags=tags)
+
+
+@irrelevant(library="cpp")
+@released(dotnet="2.1.0", golang="1.53.0", java="0.102.0", nodejs="?", php="0.74.0", python="1.2.1", ruby="?")
```
mtoffl01, line 69: What about here and my other new scenarios @cbeauchesne ?
The other tests still have `@released(java="0.102.0", php="?", python="1.2.1")`. I imagine the changes have not been made in the manifests for Java, php and python, so those are the only languages that need to be included in the `released` decorator?

tests/test_library_conf.py
mtoffl01: in progress: https://dd.slack.com/archives/CC7Q58YRG/p1694114623619059

lucaspimentel: I created this ticket for us to look into this: https://datadoghq.atlassian.net/browse/AIT-8308
Thanks!

tests/test_library_conf.py
cbeauchesne: I'm currently migrating all those declarations in manifest file, I've done C++, .NET, golang, nodejs and ruby. The last three tracer will be done in coming days :)

tests/test_library_conf.py
```diff
@@ -59,5 +59,121 @@ def test_trace_header_tags(self):
...
+        for tag in tag_config_list:
+            tag = tag.strip()
+            tags["http.request.headers." + tag] = re.sub("[^a-zA-Z0-9 -]", "_", self.headers[tag])
+
+        interfaces.library.add_span_tag_validation(request=self.r, tags=tags)
+
+
+# Note: dotnet fails this test. It normalizes those spaces to underscores instead.
+@irrelevant(library="cpp")
+@released(dotnet="?", golang="1.53.0", java="0.102.0", nodejs="?", php="?", python="1.2.1", ruby="?")
```
cbeauchesne, line 96: Can you use manifest files for those two lines ? (ping me if you need any help)

cbeauchesne (approved):

tests/test_library_conf.py
mtoffl01: > And could you comment why you remove the first item ?
This was the syntax when I started working on the tests along with the comment, "# skip the first item, as this required to make the tests work on some platforms." I don't see a git blame for the comments and I don't have context on why this was the implementation in the first place.
But I will remove the additional colon since the step input is not required

cbeauchesne: ah, I understand, it's a code coming from a previous test

utils/interfaces/_library/miscs.py
mtoffl01: What is the new way? I was just following the `_SpanTagValidator`

utils/_context/header_tag_vars.py
```diff
@@ -0,0 +1,51 @@
...
+
+class TestData:
+    def __init__(self, header, value, config, tag):
+        self.header = header
+        self.value = value
+        self.config = config
+        self.tag = tag
+
+
+header_tag_tests = {
```
cbeauchesne, line 12: From my experience, moving data away from test classe makes tests hard to debug, could you repopulate every data where they are used?

utils/_context/header_tag_vars.py
```diff
@@ -0,0 +1,51 @@
...
+    "Test_HeaderTags_Whitespace_Tag": TestData(header="header4", value="val", config="header4: t a g ", tag="t a g"),
+    "Test_HeaderTags_Whitespace_Val_Short": TestData(
+        header="header5", value=" v a l ", config="header5", tag=prefix + "header5"
+    ),
+    "Test_HeaderTags_Whitespace_Val_Long": TestData(header="header6", value=" v a l ", config="header6:tag", tag="tag"),
+    "Test_HeaderTags_Colon_Leading": TestData(header="header7", value="val", config=":header7", tag=None),
+    "Test_HeaderTags_Colon_Trailing": TestData(header="header8", value="val", config="header8:", tag=None),
+}
+
+# I'm unclear on the "Python way" to do getters, LMK.
```
cbeauchesne, line 29: Go simple, do not use python trickery, simply put the raw data where they belong :   replace `{Header_Name(self.name): Header_Value(self.name)}` by `{"header2":"val"}`

utils/interfaces/_library/miscs.py
cbeauchesne: it's using directly the span accessors inside you test methd: 

```python
def test_tags_absence(self):
  forbidden_tags = ("tag_1", "tag_2")

  for data, trace, span in interfaces.library.get_spans(request=request):
     for tag in forbidden_tags:
        assert tag not in span["meta"]
```

Having all the test logic in test method makes the test easier to understand on debug. 

Two rules of thumb : 

* data must be present and clear in test method, or at least, very close to it
* non-python engineer must easily understand the test code by reading it

cbeauchesne (changes requested):

utils/interfaces/_library/miscs.py
mtoffl01: So does that mean I should replace all calls to `interfaces.library.add_span_tag_validation` with code like the above as well?

cbeauchesne:
I've converted to draft (lot of spam in my notif 😅 ), you can set it back to normal once you need a review.

utils/_context/_scenarios.py
```diff
@@ -1314,13 +1317,10 @@ class scenarios:
...
 
-    library_conf_custom_headers_short = EndToEndScenario(
-        "LIBRARY_CONF_CUSTOM_HEADERS_SHORT", additional_trace_header_tags=("header-tag1", "header-tag2"), doc=""
-    )
-    library_conf_custom_headers_long = EndToEndScenario(
-        "LIBRARY_CONF_CUSTOM_HEADERS_LONG",
-        additional_trace_header_tags=("header-tag1:custom.header-tag1", "header-tag2:custom.header-tag2"),
-        doc="",
+    library_conf_custom_header_tags = EndToEndScenario(
+        "LIBRARY_CONF_CUSTOM_HEADER_TAGS",
```
cbeauchesne, line 1321: Can you update `.github/workflows/ci.yml` with the new name ? You'll probably have another occurence of `LIBRARY_CONF_CUSTOM_HEADERS_SHORT` and `LIBRARY_CONF_CUSTOM_HEADERS_LONG` to fix

cbeauchesne (changes requested):

cbeauchesne (changes requested):

cbeauchesne:
in weblog stdout : 

```
[dd.trace 2023-11-15 10:16:23:472 +0000] [main] WARN datadog.trace.bootstrap.config.provider.ConfigConverter - Invalid config for trace.header.tags. Illegal empty key at position 119. Must match '(key:value|key)([ ,](key:value|key))*'.
[dd.trace 2023-11-15 10:16:23:472 +0000] [main] WARN datadog.trace.bootstrap.config.provider.ConfigConverter - Invalid config for trace.header.tags. Illegal empty key at position 119. Must match '(key:value|key)([ ,](key:value|key))*'.
```

-> you can skip the test as bug for java, if the test is in line with the RFC :)

cbeauchesne:
Still failing on Java. Ping me if you need any help.

cbeauchesne (changes requested):
Can you update `.github/workflows/ci.yml` ?

utils/interfaces/_library/miscs.py
cbeauchesne: I missed your comment, sorry. So the answer is yes. I need to remove the legacy way, my bad.

cbeauchesne (approved):

PR #1058, opened 2023-04-07
parametric(otel): add tests for setting service names on spans

mabdinur:
## Description

Ensures DD Service name is equal to the value set in  service.name attribute/tag. This is described in the [otel spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md#service)


parametric/test_otel_span_methods.py
```diff
@@ -41,6 +41,32 @@ def test_otel_start_span(test_agent, test_library):
...
+    """
+        - Update the operation name, resource name, and service name on a span
+    """
+
+    with test_library:
+
+        with test_library.otel_start_span(
+            "operation",
+        ) as parent:
+            parent.update_name("new_name")
```
mabdinur, line 60:
```suggestion
            parent.set_name("new_name")
```

parametric/test_otel_span_methods.py
```diff
@@ -41,6 +41,32 @@ def test_otel_start_span(test_agent, test_library):
...
+def test_otel_update_names(test_agent, test_library):
+    """
+        - Update the operation name, resource name, and service name on a span
+    """
+
+    with test_library:
+
+        with test_library.otel_start_span(
+            "operation",
+        ) as parent:
```
mabdinur, line 59:
```suggestion
        with test_library.otel_start_span("operation") as parent:
```

parametric/test_otel_span_methods.py
```diff
@@ -41,6 +41,30 @@ def test_otel_start_span(test_agent, test_library):
...
+@pytest.mark.skip_library("java", "Not implemented")
+@pytest.mark.skip_library("php", "Not implemented")
+@pytest.mark.skip_library("ruby", "Not implemented")
+def test_otel_update_names(test_agent, test_library):
+    """
+        - Update the operation name, resource name, and service name on a span
+    """
+
+    with test_library:
+
```
mabdinur, line 56:
```suggestion
```

parametric/test_otel_span_methods.py
```diff
@@ -41,6 +41,29 @@ def test_otel_start_span(test_agent, test_library):
...
+@pytest.mark.skip_library("nodejs", "Not implemented")
+@pytest.mark.skip_library("python", "Not implemented")
+@pytest.mark.skip_library("java", "Not implemented")
+@pytest.mark.skip_library("php", "Not implemented")
+@pytest.mark.skip_library("ruby", "Not implemented")
+def test_otel_update_names(test_agent, test_library):
+    """
+        - Update the operation name, resource name, and service name on a span
+    """
+
```
mabdinur, line 54:
```suggestion
```

parametric/test_otel_span_methods.py
```diff
@@ -41,6 +41,28 @@ def test_otel_start_span(test_agent, test_library):
     assert root_span["duration"] == duration * 1_000  # OTEL expects microseconds but we convert it to ns internally
 
 
+@pytest.mark.skip_library("dotnet", "Not implemented")
+@pytest.mark.skip_library("nodejs", "Not implemented")
+@pytest.mark.skip_library("python", "Not implemented")
+@pytest.mark.skip_library("java", "Not implemented")
+@pytest.mark.skip_library("php", "Not implemented")
+@pytest.mark.skip_library("ruby", "Not implemented")
```
mabdinur, line 49:
```suggestion
@pytest.mark.skip_library("golang", "BUG: Span.set_name() does not update the resource name")
```

cbeauchesne (changes requested):
The test fails on ruby -> https://github.com/DataDog/system-tests/actions/runs/4658128833/jobs/8243467495?pr=1058#step:5:146

parametric/test_otel_span_methods.py
```diff
@@ -41,6 +41,29 @@ def test_otel_start_span(test_agent, test_library):
...
+    """
+    with test_library:
+        with test_library.otel_start_span("operation") as parent:
+            parent.set_name("new_name")
+            parent.set_attributes({"service.name": "new_service_name"})
+            parent.end_span()
+
+    root_span = get_span(test_agent)
+    assert root_span["name"] == "new_name"
+    assert root_span["resource"] == "new_name"
```
bantonsson, line 63: I really don't know why the span name and the resource name are tied together like this. The OTLP endpoint in the Datadog agent doesn't do this as far as I can tell. What is the use case we are trying to cover with this?

mabdinur: The OpenTelemetry specification does not have a concept of resource names but resource names are vital part of the Datadog Span model.

We should ensure resource names are set in some consistent manner across tracers. Here I am tying span name and resource name together to ensure this value is set in a consistent manner across tracers.

bantonsson: Yes, but how does the OTLP endpoint in the agent work right now. It would be good if switching from one way of integrating with Datadog (OTel SDK and the Datadog agent OTLP endpoint) to another (Datadog Tracer OTEL API) produced the same results.

mabdinur: @katiehockman Correct me if I am wrong but in OTLP ingest the [resouce.name](https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/api/otlp.go#L472) attribute is used to set the resource field on Datadog spans. Should we do the same here? 
ex: `otel_span.set_attribute("resource.name", "res") -> datadog_span.resouce = res`

Also do we need to document this behavior? I can find a reference to this tag in the opentelemetry-api or  [openetelemtry-sdk] (https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md#service) specification.

bantonsson: Yup, that tag is not part of OTEL semantics, so I think it's unlikely to be present on traces coming into the OTLP endpoint. What happens if the resource is unset, I don't know.

katiehockman: > the resouce.name attribute is used to set the resource field on Datadog spans. Should we do the same here?
> ex: otel_span.set_attribute("resource.name", "res") -> datadog_span.resouce = res

Yes that sounds right.

I agree with @bantonsson that we shouldn't perform off-spec behavior by setting the resource name with `UpdateName` unless we eventually have a very compelling reason to do so. Hopefully where possible, the languages have allowed Datadog-specific options to be provided as part of the OTel API ([like what we did with Go](https://pkg.go.dev/gopkg.in/DataDog/dd-trace-go.v1/ddtrace/opentelemetry@main#ContextWithStartOptions)), so that can be another option for setting resource names for some users, though I realize that perhaps not every language can do this.

mabdinur: Both logs-backend and the datadog agent support converting otlp spans to datadog spans. Looking at the source code it appears we set `span.resource = span.name` if the `resource.name` attribute is not set ([backend](https://github.com/DataDog/logs-backend/blob/3a8318b7c71a43a3b538821facec711e7252c0d0/domains/event-platform/libs/raw-content-parsers/src/main/java/com/fsmatic/workload/processing/SpansOTLPProtobufPayloadParser.java#L435), [agent](https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/api/otlp.go#L595)).

mabdinur: The otlp endpoint implements the following logic: 
- Resource names are set using `Span.set_attribute("resource.name", "some_resource_value") `
- Setting resource names using attributes is not part of OTEL semantics. The `resource.name` attribute key should be documented. 
- If `otel_span.set_attribute("resource.name", .....) is NOT set, then the resource name should be equal to the operation name. 

Currently the python and go implementations do not align with the otlp spec.  We should update the RFC to include how resource names should be handled (or we can defer this work).

I will remove the resource name assertions from this test. We can address it in a follow up.

marcotc: Is the conclusion here then that, in an OTel span, `resource` will be empty in the tracer, but downstream a fallback will be applied?

mabdinur: [OTLP ingest](https://docs.datadoghq.com/opentelemetry/otlp_ingest_in_the_agent/?tab=host) converts spans generated using the opentelemetry sdk to datadog spans. It's alternative to configuring a datadog OpenTelemetry TracerProvider. We need to make sure both mechanisms set resource names in a consistent manner.

Correct me if I am wrong `DatadogSpan.resource` can't be empty. It should be initialized to some value. What that value should be is still an open question. One solution is to use the same convention as otlp ingest but I don't think a decision has been made on this.

katiehockman: This is also relevant: https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/config/config.go#L61

> 	// SpanNameAsResourceName specifies whether the OpenTelemetry span's name should be
	// used as the Datadog span's operation name. By default (when this is false), the
	// operation name is deduced from a combination between the instrumentation scope
	// name and the span kind.
	//
	// For context, **the OpenTelemetry 'Span Name' is equivalent to the Datadog 'resource name'**.
	// The Datadog Span's Operation Name equivalent in OpenTelemetry does not exist, but the span's
	// kind comes close.
	SpanNameAsResourceName bool `mapstructure:"span_name_as_resource_name"`

From [this part of the code](https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/api/otlp.go#L592), it looks like trying to figure out the resource name isn't as straightforward as just using the name. It looks in the meta field for various things, then if it's still empty, it defaults to the span name.

Either way, I don't think we should automatically clobber any existing resource name that might be on a span if `UpdateName` is used. The resource name could have been configured in some other way (like I described with Go in my previous comment, or by setting a `resource.name` attribute).

Agreed that this should be covered in the RFC though, so if you'd like to make a suggestion directly in the doc for that, I can review it. 

> Correct me if I am wrong DatadogSpan.resource can't be empty.

Yes I'm pretty sure it can't be empty when it's sent to the backend.

parametric/test_otel_span_methods.py
```diff
@@ -41,6 +41,27 @@ def test_otel_start_span(test_agent, test_library):
...
+        - Update the service name on a span
+    """
+    with test_library:
+        with test_library.otel_start_span("parent_span") as parent:
+            parent.set_attributes({"service.name": "new_service"})
+            parent.end_span()
+
+    root_span = get_span(test_agent)
+    assert root_span["name"] == "parent_span"
+    assert root_span["resource"] == "parent_span"
```
mabdinur, line 61:
```suggestion
```

parametric/test_otel_span_methods.py
mabdinur: This PR no longer asserts that `update_name()` sets the resource name on OTEL spans. I will resolve this conversation. 

We can continue this discussion on slack. I'll update the RFC to include a proposal on how we should handle resource naming for OpenTelemetry Spans.

cbeauchesne (approved):

PR #1103, opened 2023-04-21
suspicious request blocking tests upgrade

christophe-papazian:
## Description

rework of tests for suspicious request blocking feature for ASM after https://github.com/DataDog/system-tests/pull/1078

- remove set_value/get_value entry points and use a tag_value that should add the value obtained in path parameters in the meta tags of the span
- update tests accordingly
- use urlencoded body for request.body test as this should be supported by all ASM libraries


simon-id (commented):
Can't see anything wrong with it but I have zero python experience so take it with a grain of salt

docs/weblog/README.md
```diff
@@ -126,6 +126,25 @@ In particular, it accepts and parse JSON and XML content. A typical XML content
...
 </string>
 ```
+
+## \[GET, POST, OPTIONS\] /set_tag/%s/%d
+
+This endpoint must accept two required parameters (the first is a string, the second is an integer) and are part of the URL path. 
+
+This endpoint must accept all query parameters and all content types.
+
+The first path parameter must be wrote in the span with the tag `appsec.events.system_tests_appsec_event.value` and the parameter as value.
```
simon-id, line 136:
```suggestion
The first path parameter must be written in the span with the tag `appsec.events.system_tests_appsec_event.value` and the parameter as value.
```

docs/weblog/README.md
```diff
@@ -126,6 +126,25 @@ In particular, it accepts and parse JSON and XML content. A typical XML content
...
+
+The second path parameter must be used as a response status code.
+
+All query parameters (key, value) must be used as (key, value) in the response headers.
```
simon-id, line 140: what response header ?

tests/appsec/test_blocking_addresses.py
```diff
@@ -134,6 +135,21 @@ def test_response_cookies(self):
         assert False
 
 
+def validate_custom_event_tag(expected_value="", present=True):
+    def wrapper(span):
+        tag = "appsec.events.system_tests_appsec_event.value"
+        if present:
+            assert tag in span["meta"], f"Can't find {tag} in span's meta"
+            value = span["meta"][tag]
+            if value != expected_value:
```
Anilm3, line 144: It's unclear to me what the difference between raising an exception and asserting is in this context, but I was wondering if replacing the `if/raise` with `assert value == expected_value`  would have the same effect?

Anilm3 (approved):
lgtm, not sure if my approval will allow you to merge,but in case it does you should probably wait for @cbeauchesne's approval.

tests/appsec/test_blocking_addresses.py
christophe-papazian: I replaced that by an assert as suggested.

docs/weblog/README.md
```diff
@@ -126,6 +126,25 @@ In particular, it accepts and parse JSON and XML content. A typical XML content
...
+
+This endpoint must accept two required parameters (the first is a string, the second is an integer) and are part of the URL path. 
+
+This endpoint must accept all query parameters and all content types.
+
+The first path parameter must be written in the span with the tag `appsec.events.system_tests_appsec_event.value` and the parameter as value.
+
+The second path parameter must be used as a response status code.
+
+All query parameters (key, value) must be used as (key, value) in the response headers.
```
Julio-Guerra, line 140: Unclear to me. Maybe provide an example?

christophe-papazian: Yes, good idea, I add a full example soon.

tests/appsec/test_blocking_addresses.py
```diff
@@ -134,6 +135,20 @@ def test_response_cookies(self):
         assert False
 
 
+def validate_custom_event_tag(expected_value="", present=True):
```
cbeauchesne, line 138: It's more for the code clarety, could you split this in twwo functions : 

```python
def _assert_custom_event_tag_presence(expected_value):
def _assert_custom_event_tag_absence():
```

IMO, the test line will be a little more clear, and won't require any cognitive effort on reading nested `if/else`.

cbeauchesne (changes requested):

tests/appsec/test_blocking_addresses.py
christophe-papazian: Done

cbeauchesne (approved):

PR #1095, opened 2023-04-19
Improve nix environment

lloeki:
## Description

- Improve nix environment
- Fix docker client connection when using `DOCKER_HOST` env var
- Allow docker connection over `DOCKER_HOST=ssh://`

## Additional notes

This DOES NOT fix reaching to the various docker services exposed ports, which need do be accessed using the external IP of where Docker is hosted (in my case a VM). This IP is the one present in `DOCKER_HOST`.

There are theoretically two cases:

- `runner` (outside VM) reaching to `agent` (inside VM) and `weblog` (inside VM), notably for health check
  
  This needs the VM IP. `DOCKER_HOST` contains it in my case.
 
- weblog (inside VM) reaching to runner's proxy (IIUC outside VM)

  This needs the host IP on the VM network. it should be the TCP client IP that a listening process running in the VM sees as a peer.

<details>

<summary>List of references to `localhost`</summary>

```
utils/interfaces/_backend.py
126:        # We use `localhost` as host since we run a proxy that forwards the requests to the right
129:        host = f"http://localhost:{BACKEND_LOCAL_PORT}"
221:        # We use `localhost` as host since we run a proxy that forwards the requests to the right
224:        host = f"http://localhost:{BACKEND_LOCAL_PORT}"

utils/interfaces/schemas/README.md
5:Then go to <http://localhost:7070>

utils/_context/containers.py
215:            healthcheck=_HealthCheck(f"http://localhost:{self.agent_port}/info", 60, start_period=1),
256:            healthcheck=_HealthCheck("http://localhost:7777", 60),

utils/_context/_scenarios.py
502:            requests.get("http://localhost:7777", timeout=10)

utils/build/docker/php/php-fpm/php-fpm.conf
25:            SetHandler "proxy:unix:/run/php/phpPHP_VERSION-fpm.sock|fcgi://localhost"

utils/build/docker/dotnet/Properties/launchSettings.json
6:      "applicationUrl": "http://localhost:61558/",
24:      "applicationUrl": "https://localhost:5001;http://localhost:5000"

utils/build/docker/dotnet/Endpoints/MakeDistantCallEndpoint.cs
43:                    // http://localhost:7777/make_distant_call?url=http%3A%2F%2Fweblog%3A7777
44:                    // https://localhost:44381/make_distant_call?url=https%3A%2F%2Flocalhost%3A44381
47:                    var example = "http://localhost:7777/make_distant_call?url=http%3A%2F%2Fweblog%3A7777";

[...]

PR #1109, opened 2023-04-24
Dockerize runner

lloeki:
## Description

Allow the runner to be run within docker.

This works by considering a docker container as just being a host in its own right, by starting the container with a command such as  the following, we can provide all runtime dependencies needed by the runner (API key, access to `dockerd`, log writing).

```
docker run --rm -it --network system-tests_default -v "${PWD}"/.env:/app/.env -v /var/run/docker.sock:/var/run/docker.sock -v "${PWD}"/logs:/app/logs --name system-tests-runner system_tests/runner ./run.sh
```

Save for secrets, the image contains everything needed to be self-sufficient and is built thusly:

```
docker buildx build -f utils/build/docker/runner.Dockerfile -t system_tests/runner:latest .
```

If so desired (e.g quick development of tests), select folders can be overridden on a case by case basis just by adding the corresponding volumes:

```
docker run --rm -it -v "${PWD}"/tests:/app/tests -v "${PWD}"/parametric:/app/parametric -v "${PWD}"/utils:/app/utils
```

As of opening this PR, I did not update the tooling (`build.sh` + `run.sh`) to allow for these commands to be run.

## Motivation

The runner now has the ability to run on the host, but this removed the option of running it within Docker, as all Docker supporting files regarding the runner have been removed and its container infrastructure revamped (the runner is now in charge of starting the containers, whereas previously it was handled outside of it by docker-compose).

While this host-running ability is useful for local development of system tests, having the ability to run the runner in Docker had multiple advantages:

- easy cacheability of python dependences
- easy reproducibility by pulling an image
- running environment consistency
- not having to install a specific python version

In addition, dd-trace-rb CI uses a docker-based build->run job order that minimizes build time and severely reduces its impact.

So with this PR we attempt to get the best of both worlds!

## Additional notes

As of the opening of this PR, the runner is able to access the agent and weblog information:

```
================================================================================================================== test context ==================================================================================================================
Scenario: DEFAULT
Logs folder: ./logs
Library: ruby@1.10.1
Agent: 7.43.1
libddwaf: 1.6.2.0.0
Weblog variant: rack
Backend: datad0g.com
============================================================================================================== test session starts ===============================================================================================================
collected 348 items / 145 deselected / 203 selected                                                                                                                                                                                              
Executing weblog warmup...
```

Sadly it appears it fails to reach to the agent:

```
------------------------------------------------------------------------------------------------- generated xml file: /app/logs/reportJunit.xml --------------------------------------------------------------------------------------------------
======================================================================================================= 145 deselected in 64.06s (0:01:04) =======================================================================================================
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! _pytest.outcomes.Exit: http://localhost:8127/info never answered to healthcheck request !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
```

Indeed, while it appears to be able to start the agent container, it fails to reach to it:

```
$ docker ps
CONTAINER ID   IMAGE                 COMMAND                 CREATED         STATUS                            PORTS                                          NAMES
b82c4b82f381   system_tests/agent    "/bin/entrypoint.sh"    7 seconds ago   Up 6 seconds (health: starting)   8125/udp, 8126/tcp, 127.0.0.1:8127->8127/tcp   system-tests-agent
1adbad5ce645   system_tests/runner   "/bin/sh -c ./run.sh"   8 seconds ago   Up 7 seconds                                                                     xenodochial_lehmann
```

Since it manages to start the agent, I'm confident this can be made to work though. It might be failing because of the `localhost` things.


.dockerignore
```diff
@@ -4,3 +4,4 @@ venv/
 vendor
 logs*
 shell.nix
+__pycache__
```
lloeki, line 7: These should not be copied in docker images, ever.

utils/build/build.sh
```diff
@@ -283,15 +283,15 @@ TEST_LIBRARY="${TEST_LIBRARY:-${DEFAULT_TEST_LIBRARY}}"
 BINARY_PATH="${BINARY_PATH:-}"
 BINARY_URL="${BINARY_URL:-}"
 
-if [[ ! -d "${SCRIPT_DIR}/docker/${TEST_LIBRARY}" ]]; then
+if [[ "${BUILD_IMAGES}" =~ /weblog/ && ! -d "${SCRIPT_DIR}/docker/${TEST_LIBRARY}" ]]; then
```
lloeki, line 286: This check was preventing `build.sh` to proceed but it's irrelevant when building `runner` and `agent`.

Indeed there seems to be much assumptions in the code flow of this script that it's building `weblog` images, even when it is not.

utils/build/docker/runner.Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+# basically everything except utils/build
+COPY utils/_context /app/utils/_context
+COPY utils/assets /app/utils/assets
+COPY utils/grpc /app/utils/grpc
+COPY utils/proxy /app/utils/proxy
+COPY utils/interfaces /app/utils/interfaces
+COPY utils/scripts /app/utils/scripts
+COPY utils/test_the_test /app/utils/test_the_test
+COPY utils/*.py /app/utils/
```
lloeki, line 24: This section attempts to minimize dependencies when building the image. In this case `utils/build` is entirely irrelevant but it can't be added to `.dockerignore` since the latter has its own docker things.

A bit of refactoring with dedicated folders might make that easier.

utils/build/docker/runner.Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+# toplevel things
+COPY conftest.py /app/
+COPY pyproject.toml /app/
+COPY run.sh /app/
```
lloeki, line 29: This section attempts to minimize dependencies when building the image. In this case `utils/build` is entirely irrelevant but it can't be added to `.dockerignore` since the latter has its own docker things.

A bit of refactoring with dedicated folders might make that easier.

utils/build/docker/runner.Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+CMD ./run.sh
```
lloeki, line 31: This could be improved with an entrypoint to allow passing arguments to `./run.sh`, e.g:

```
docker run .... system_tests/runner <scenario> <test_file>
```

lloeki: Also, just as with reusing `build.sh` above, `run.sh` handles venv activation too so we're golden.

lloeki:
I now have this working locally with a few workarounds. Two main issues:

- `localhost` (originating from port forward/expose) instead of container naming, similar case as for #1095
- bind mounts giving an incorrect path when runner is containerized:
  - to logs folders for proxy and weblog
  - to source code in utils for proxy

Interestingly enough, when I attempted #1095 on my configuration (which has docker in an isolated VM) I assumed that the proxy code was running in the runner process, so there needed to be a docker->host communication to reach it... But it's in a container, so there's no need to know the host IP on the VM network. Therefore solving this one will make a VM-ized docker setup work as well.

utils/build/docker/runner.Dockerfile
cbeauchesne: I would say you want to have all of this as volume, to avoid rebuilding the runner each time you modify a test, WDYT ?

cbeauchesne (changes requested):
I strongly recommend adding a CI using this setup, otherwise we may break it one day or another.

lloeki:
> I strongly recommend adding a CI using this setup

Agreed. I did not want to waste time doing it until I could validate that this is a workable path.

cbeauchesne:
> Since it manages to start the agent, I'm confident this can be made to work though. It might be failing because of the localhost things.

I slightly change the way healthcheck is done, rebasing your branch may solve that.

utils/build/docker/runner.Dockerfile
lloeki: Nothing about this prevents anyone to `-v $PWD:/app`, the bind mount will mask the image directory should one want to use docker for development.

This makes the image entirely self contained, which is what we want e.g in CI for caching, reproducibility, and consistency, and generally the whole purpose of a docker image.

Pushed to an extreme, with the new runner handling all the container spawning it makes it very possible to run the system tests suite without ever checking out the system tests repository.

lloeki:
> I slightly change the way healthcheck is done, rebasing your branch may solve that.

Ha, coincidentally I made significant improvements to that part, allowing more use cases to succeed. I'll see how to best resolve the conflicts.

May I ask why the use of curl now instead of python `requests`? (which I feel was better in some ways)

utils/_context/_scenarios.py
```diff
@@ -67,7 +67,7 @@ def session_start(self, session):
         self.terminal = session.config.pluginmanager.get_plugin("terminalreporter")
         self.terminal.write_sep("=", "test context", bold=True)
         self.print_info(f"Scenario: {self.name}")
-        self.print_info(f"Logs folder: ./{self.host_log_folder}")
+        self.print_info(f"Logs folder: {self.host_log_folder}")
```
cbeauchesne, line 70: The `./` is quite important for devx, as it make the item clickable on most bash terminal

lloeki: Alright. Would it work if it were an absolute path as well?

run.sh
```diff
@@ -1,116 +1,345 @@
...
+        # cleanups
+        clean_pycache
+
+        # ensure environment
+        if ! is_using_nix; then
+            activate_venv
+        fi
+    fi
+
+    echo "Run plan:"
```
cbeauchesne, line 266: Do not add anything in sdtout here, pytest must be the only process to handle the ouput.

utils/_context/containers.py
```diff
@@ -248,8 +282,10 @@ def __init__(self, host_log_folder, proxy_state) -> None:
             },
             working_dir="/app",
             volumes={
-                f"./{host_log_folder}/interfaces/": {"bind": f"/app/{host_log_folder}/interfaces", "mode": "rw",},
-                "./utils/": {"bind": "/app/utils/", "mode": "ro"},
+                f"{self.host_project_dir}/{host_log_folder}/interfaces/": {
+                    "bind": f"/app/{host_log_folder}/interfaces",
+                    "mode": "rw",
+                },
```
cbeauchesne, line 288: We need `"./utils/": {"bind": "/app/utils/", "mode": "ro"},`

utils/_context/containers.py
```diff
@@ -271,13 +307,26 @@ def __init__(self, host_log_folder, use_proxy=True) -> None:
...
+            agent_host = "localhost"
+
         super().__init__(
             image_name="system_tests/agent",
             name="agent",
             host_log_folder=host_log_folder,
             environment=environment,
-            healthcheck={"test": f"curl --fail http://localhost:{self.agent_port}/info", "retries": 60},
-            ports={f"{self.agent_port}/tcp": ("127.0.0.1", self.agent_port)},
+            healthcheck={"test": f"curl --fail http://{agent_host}:{agent_port}/info", "retries": 60},
```
cbeauchesne, line 328: it's executed inside the container, so you can keep `localhost`.

utils/_context/containers.py
```diff
@@ -248,8 +282,10 @@ def __init__(self, host_log_folder, proxy_state) -> None:
             },
             working_dir="/app",
             volumes={
-                f"./{host_log_folder}/interfaces/": {"bind": f"/app/{host_log_folder}/interfaces", "mode": "rw",},
-                "./utils/": {"bind": "/app/utils/", "mode": "ro"},
+                f"{self.host_project_dir}/{host_log_folder}/interfaces/": {
```
cbeauchesne, line 285: You can keep the `.` here, and handles `host_project_dir` in `_fix_host_pwd_in_volumes` : 

```
host_pwd = os.getcwd()

# replaced by

host_pwd = os.environ.get("HOST_PROJECT_DIR", os.getcwd())
```

It will keep the volume definition simpler, sticking to the docker compose logic.

utils/_context/containers.py
```diff
@@ -316,17 +365,26 @@ def __init__(
...
+                    "bind": "/var/log/system-tests",
+                    "mode": "rw",
+                },
+            },
             # ddprof's perf event open is blocked by default by docker's seccomp profile
             # This is worse than the line above though prevents mmap bugs locally
             security_opt=["seccomp=unconfined"],
-            healthcheck={"test": "curl --fail http://localhost:7777", "retries": 60},
-            ports={"7777/tcp": ("127.0.0.1", 7777), "7778/tcp": ("127.0.0.1", 7778)},
+            healthcheck={"test": f"curl --fail {weblog._get_url('/')}", "retries": 60},
```
cbeauchesne, line 386: Executed inside the container, you can keep localhost

cbeauchesne (changes requested):

run.sh
```diff
@@ -1,116 +1,345 @@
...
+    tr '[:upper:]' '[:lower:]'
+}
 
-ONBOARDING_SCENARIOS=(
-    ONBOARDING_HOST
-    ONBOARDING_HOST_CONTAINER
-    ONBOARDING_CONTAINER
-)
+function list_scenarios() {
+    grep -R '^@scenarios\.' tests | sed -e 's/.*@scenarios\.//' | sort | uniq | upcase
```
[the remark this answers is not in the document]
lloeki: No problem, this is unused anyway. It was just for quick testing.

run.sh
lloeki: Got it.

lloeki (commented):
I'll do some changes to comply with the requirements.

Once these are addressed, there are a still a few docker mode cases where failures can happen in specific situations, but as long as direct mode keeps working I believe we should merge this and address the remaining corner case details of docker mode in smaller subsequent PRs.

utils/_context/containers.py
```diff
@@ -31,6 +31,37 @@ def create_network():
     _get_client().networks.create(_NETWORK_NAME, check_duplicate=True)
 
 
+class _HealthCheck:
```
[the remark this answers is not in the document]
lloeki: Removed

lloeki (commented):
I've merged `main` again and resolved the conflicts ported the latest changes in.

I'll do a couple more changes on this PR:
- [ ] extract the embedded YAML scenario group definitions to their own YAML file (this annoys `shellcheck`)
- [ ] factor out `host_project_dir` and restore leading dots for volumes

If that's OK and the latest changeset look fine to you I'd like to merge it ASAP and pursue with additional changes in smaller PRs because I'm a bit annoyed to constantly have to fix conflicts 😅.

.github/workflows/ci.yml
```diff
@@ -221,6 +221,8 @@ jobs:
     - name: Log in to the Container registry
       if: ${{ matrix.variant.library == 'ruby' }}
       run: echo ${{ secrets.GITHUB_TOKEN }} | docker login ${{ env.REGISTRY }} -u ${{ github.actor }} --password-stdin
+    - name: Build proxy
```
cbeauchesne, line 224: We do not build it, on purpose, to reduce CI time.

lloeki: I understand the desire to reduce CI time, that is why I made sure the `buildx build` commands are aligned:

- thanks to `--build-arg BUILDKIT_INLINE_CACHE=1` as soon as a new image that was built with the flag is published it will take 0s when there is no change from the previously published image (may need to use `--cache-from`)
- even as is, [It takes 5s](https://github.com/DataDog/system-tests/actions/runs/5475850663/jobs/9972604291?pr=1109#step:10:15) which is dwarfed by weblog image building by an order of magnitude.

lloeki: > But the question is : why should we want to rebuild it

Precisely for the reason highlighted by the failure this fixes: when sources of the docker image change it uses whatever has been published before instead of the current branch state, resulting in temporal inconsistency.

> Most of PR only run default scenario, and take 2 to 3 minutes

I am not sure I follow your point:

- once a new image built with inline cache is pushed it will be a noop (that also ensures temporal consistency as outlined above)
- agent and weblog images are rebuilt each time (for similar temporal consistency reasons) even for default scenarios, and _these_ are dwarfing proxy rebuild

cbeauchesne: The point is that the source of the proxy are very stable: very few deps. It's all the point about publishing docker image to docker hub. If we need to update the source, we'll update the version number inside the tag. `datadog/system-tests:proxy-v0` we stay compatible with the current main branch and all PRs. If we need to update it, we'll bump to `datadog/system-tests:proxy-v1`. I've just pushed the process to update it on #1379. 

> agent and weblog images are rebuilt each time (for similar temporal consistency reasons) even for default scenarios, and these are dwarfing proxy rebuild

Because their sources are far less stable by several order of magnitude. For the `dev` part of our CI, we even need to get the very last artifact from all components. Without talking about the fact that we have lot of PR that modifies weblogs.

It's not the case for the proxy. So even if it's a small win, it's still a win. And having the image published on docker hub working with the main branch will be part of the contract, so no worry about that. If you're afraid of issue on that part, you can rebuild it on your CI. But as I want to offer this contract, this CI must use this feature (image on docker hub is up. to date).

lloeki: How do you suggest we proceed?

AIUI the process would be:

1. revert https://github.com/DataDog/system-tests/commit/fbb03ac20da6fce8a11090fd4443574ce6b6965e
2. since the PR uses `datadog/system-tests:proxy-v0` and [`utils/build/docker/proxy.Dockerfile`](https://github.com/DataDog/system-tests/blob/ace288bf8281ccc923a9dab05dd20def5a9c2c80/utils/build/docker/proxy.Dockerfile) changed, this PR's CI will fail
3. checkout PR branch
4. change `datadog/system-tests:proxy-v0` to `datadog/system-tests:proxy-v1`
5. commit and push to PR
7. since the PR references a nonexistent `datadog/system-tests:proxy-v1` this PR's CI will fail
8. build proxy image locally
9. push image as `datadog/system-tests:proxy-v1`
10. manually trigger a CI re-run for this PR
11. wait for CI to be green
12. merge

cbeauchesne: 1 to 5 -> yep

Then, I've added an optional step in the CI to build it if needed : 

6. add the label `build-proxy-image` -> it will make the CI rebuild the CI image locally (see this [doc](https://github.com/DataDog/system-tests/blob/main/docs/edit/update-docker-images.md), I made it few days ago)
7. ensure CI becomes green, got threw the review process
8. I will push image as `datadog/system-tests:proxy-v1` (my team is the owner of `datadog/system-tests` repo on hub.docker.com)
9. merge

lloeki:
At this stage all test are passing, proceeding with `git revert fbb03ac`

lloeki:
In https://github.com/DataDog/system-tests/pull/1379/commits/bacec71449cd185c062efbd04b7942532fa147bf two cases seem to have been overlooked.

lloeki:
Inconsequential to this PR but I've seen this failure:

```
#9 6.908 Install args are --appsec-file /binaries/dd-appsec-php-amd64.tar.gz --tracer-version latest
#9 6.923 Searching for available php binaries, this operation might take a while.
#9 7.050 ERROR: Could not find latest release for dd-trace-php in GitHub API response.
#9 7.050 URL: https://api.github.com/repos/Datadog/dd-trace-php/releases/latest
#9 7.050 Full response follows:
#9 7.050 {"message":"API rate limit exceeded for 20.51.128.129. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}
#9 7.050 
#9 ERROR: executor failed running [/bin/sh -c /tmp/php/apache-mod/build.sh]: exit code: 1
------
 > [ 5/10] RUN /tmp/php/apache-mod/build.sh:
6.676 Setting up dumb-init (1.2.5) ...
6.679 Processing triggers for man-db (2.9.4-2) ...
0 48871    0     0  1767k      0 --:--:-- --:--:-- --:--:-- 1767k
6.908 Install args are --appsec-file /binaries/dd-appsec-php-amd64.tar.gz --tracer-version latest
6.923 Searching for available php binaries, this operation might take a while.
7.050 ERROR: Could not find latest release for dd-trace-php in GitHub API response.
7.050 URL: https://api.github.com/repos/Datadog/dd-trace-php/releases/latest
7.050 Full response follows:
7.050 {"message":"API rate limit exceeded for 20.51.128.129. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}
```

Maybe worth it to have authentication towards GH API for these fetches?

lloeki:
All green again! Pushing one last change to reinstate the proxy code volume mount.

.github/workflows/ci.yml
lloeki: Done

lloeki:
@cbeauchesne I've asked @GustavoCaso to do a local test run tomorrow to confirm it all works in another local environment than mine, and we should be good (unless you see anything else)

lloeki:
@GustavoCaso reported one issue with macOS's default bash:

```
$ /bin/bash --version
GNU bash, version 3.2.57(1)-release (arm64-apple-darwin22)
Copyright (C) 2007 Free Software Foundation, Inc.
$ /bin/bash run.sh
run.sh: line 271: libraries[@]: unbound variable
```

Which is a lie because it's bound and set, and I can't seem to reproduce it in a smaller setting:

```
#!/usr/bin/env bash

function main() {
    local scenarios=()
    local libraries=()

    IFS=',' read -r -a libraries <<< "${TEST_LIBRARY:-}"

    scenarios+=('foo')

    while [[ "$#" -gt 0 ]]; do
        libraries+=("$1")
        shift
    done

    for i in "${!scenarios[@]}"; do
      for library in "${libraries[@]}"; do
          echo "${library}"
      done
    done
}

main "$@"
```

gives:

```
~> env TEST_LIBRARY=foo,bar /bin/bash foo.sh abc def
foo
bar
abc
def
~> env TEST_LIBRARY=foo,bar /bin/bash foo.sh
foo
bar
~> /bin/bash foo.sh abc def
abc
def
~> env TEST_LIBRARY='' /bin/bash foo.sh abc def
abc
def
~> /bin/bash foo.sh
~>  
```

Still, I'll find what bash 3's big idea is here and fix it.

utils/_context/containers.py
```diff
@@ -412,7 +417,7 @@ def __init__(self, host_log_folder) -> None:
             user="postgres",
             environment={"POSTGRES_PASSWORD": "password", "PGPORT": "5433"},
             volumes={
-                "./utils/build/docker/postgres-init-db.sh": {
+                f"./utils/build/docker/postgres-init-db.sh": {
```
cbeauchesne, line 420:
```suggestion
                "./utils/build/docker/postgres-init-db.sh": {
```

utils/_weblog.py
```diff
@@ -56,12 +57,21 @@ def __init__(self, data):
 
 
 class _Weblog:
-    _grpc_port = 7778
-
     def __init__(self):
-        if "DOCKER_HOST" in os.environ:
-            self.domain = os.environ["DOCKER_HOST"]
-            self.domain = self.domain.replace("ssh://docker@", "")
+        if "WEBLOG_PORT" in os.environ:
```
cbeauchesne, line 61: I'm wondering if it would be a good idea to prefix those new env var with (`SYSTEM_TESTS_`) to prevent any collusion, and make crystal clear what is their usage when you see them in conf files, WDYT ?

utils/scripts/load-binary.sh
```diff
@@ -221,7 +221,7 @@ elif [ "$TARGET" = "cpp" ]; then
     # Not handled for now
 elif [ "$TARGET" = "agent" ]; then
     assert_version_is_dev
-    echo "datadog/agent-dev:master-py3" > agent-image
+    echo "datadog/agent-dev:master-py3@sha256:a50aee313c362a30429fc6948aeb1f63925fd43ba7c67fe9fb68a3a83c694040" > agent-image
```
cbeauchesne, line 224: Small leftover :) 

```suggestion
    echo "datadog/agent-dev:master-py3" > agent-image
```

cbeauchesne (changes requested):

utils/scripts/load-binary.sh
lloeki: Yup I did not sync with the latest `main` yet.

utils/_weblog.py
lloeki: Indeed, probably worth a larger audit of env vars and prefix in a one shot refactor, so I feel it's out of scope for this PR.

utils/scripts/load-binary.sh
lloeki: Done

.github/workflows/ci.yml
```diff
@@ -687,6 +690,9 @@ jobs:
       uses: ./.github/actions/install_runner
     - name: Pull mitmproxy image
       run: docker pull datadog/system-tests:proxy-v0
```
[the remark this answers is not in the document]
lloeki: Oops, my `rg proxy-v0` ignored dotdirs. Fixed.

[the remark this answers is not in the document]
lloeki: As expected, CI now fails because `datadog/system-tests:proxy-v1` is nonexistent, thus the above pull fails.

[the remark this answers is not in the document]
lloeki: I have worked around that with https://github.com/DataDog/system-tests/pull/1109/commits/cf81be6440a0bcc83a841cf4f46749ab939c6c0e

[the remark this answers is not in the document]
lloeki: Note that Circle CI still fails because it doesn't handle the GH tag for rebuilding the proxy image:

```
INTERNALERROR> requests.exceptions.HTTPError: 404 Client Error: Not Found for url: http+docker://localhost/v1.41/images/datadog/system-tests:proxy-v1/json
```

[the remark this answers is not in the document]
cbeauchesne: Could you  add a step that always rebuild the proxy in circleCI (it's ok here, the job is only here to ensure that system-tests is working on circleci) ?

utils/_weblog.py
lloeki: They're only used for passing information through `docker run` via `run_scenario` in `run.sh`, so they won't change.

lloeki: Fixed by 0cff8b78

.github/workflows/ci.yml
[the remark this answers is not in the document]
lloeki: Fixed in 3e68a4aa

lloeki:
Merged `main` in, resolved conflicts:

- Folded https://github.com/DataDog/system-tests/commit/81ec2fc082004a6e1e6063dcbd9bc9db8b3eda1b in the yaml file
- [That one](https://github.com/DataDog/system-tests/commit/acf3b90918ec7397204bd08b7a9ca384ed49b8f3 ) appears to be a hack, as `0.0.0.0` as a source [has at best peculiar behaviour and at worst is not routable at all](https://unix.stackexchange.com/a/419881). Really the previous external healthckeck was better, as it tested external reachability from the runner, which is what we want to ensure.

utils/_context/containers.py
```diff
@@ -521,12 +526,20 @@ class OpenTelemetryCollectorContainer(TestedContainer):
...
+
+        self._otel_port = 13133
+
         super().__init__(
             image_name=image,
             name="collector",
             command="--config=/etc/otelcol-config.yml",
             environment={},
-            volumes={self._otel_config_host_path: {"bind": "/etc/otelcol-config.yml", "mode": "ro",}},
+            volumes={f"./utils/build/docker/otelcol-config.yaml": {"bind": "/etc/otelcol-config.yml", "mode": "ro",}},
```
songy23, line 542: Why change this line?

songy23:
> [That one](https://github.com/DataDog/system-tests/commit/acf3b90918ec7397204bd08b7a9ca384ed49b8f3) appears to be a hack, as 0.0.0.0 as a source [has at best peculiar behaviour and at worst is not routable at all](https://unix.stackexchange.com/a/419881). Really the previous external healthcheck was better, as it tested external reachability from the runner, which is what we want to ensure.

Acked, if runner is also going to be in docker (which didn't work for me in CIs), the current healthcheck makes more sense.

cbeauchesne (approved):
AGTM, I've just [pushed](https://hub.docker.com/layers/datadog/system-tests/proxy-v1/images/sha256-02ce356eafacfe04d3022ff505e2dbb9e47cd3c55244b7918ff6fc824db85f02?context=repo) `datadog/system-tests:proxy-v1` image :

utils/_context/containers.py
lloeki: Fixed by 045f62e8

lloeki:
Rerunning CI to ensure it's still green after merging in `main`.

Looks like there are two consistent experimental failures from `main`.

run.sh
```diff
@@ -1,152 +1,318 @@
...
+
+    ## prepare commands
+
+    # when no scenario is provided, use a nice default
+    if [[ "${#scenarios[@]}" -lt 1 ]]; then
+        scenarios+=('DEFAULT')
     fi
 
-    # TODO: default to "auto" when dotnet is fixed
-    PYTEST_WORKER_COUNT=${PYTEST_WORKER_COUNT:-$DEFAULT_COUNT}
```
bantonsson, line 131: How is it possible that this environment variable has been removed without any communication or _replacement_? Just stop messing with things that are used in other teams CI setups! @lloeki @cbeauchesne

PR #1110, opened 2023-04-24
Use pytest -vv

smola:
## Description

Full verbose mode prints more useful asserts, such as full diffs between compared values.


smola:
Makes sense. Closing.

PR #1117, opened 2023-04-26
[ASM] APPSEC 9196: Test user blocking for up to 2500 users and IPs

M4TTRX:
## Description

[Jira Ticket](https://datadoghq.atlassian.net/browse/APPSEC-9196)

Test user blocking but with 2500 users blocked instead of one
## Motivation

We want to make sure tracers will block properly the maximum amount of blocked users allowed (2500)


simon-id:
Since the mechanism for storing blocked IPs is the exact same as the one storing blocked users, what's the point of having both a max ip and a max user scenario ?

cbeauchesne:
> Since the mechanism for storing blocked IPs is the exact same as the one storing blocked users, what's the point of having both a max ip and a max user scenario ?

Agree, if it's possible to add this test in an existing scenario it's way better.

simon-id:
@cbeauchesne tbh I was suggesting we don't even need this test, but yes good point

tests/appsec/test_user_blocking_maxed.py
```diff
@@ -0,0 +1,47 @@
...
+
+@released(
+    cpp="?",
+    dotnet="2.30.0",
+    golang="1.48.0",
+    java="?",
+    nodejs="?",
+    php="0.85.0",
+    php_appsec="0.7.0",
+    python="?",
```
juanjux, line 12:
```suggestion
    python="1.12.1",
```

tests/appsec/test_blocking_maxed.py
```diff
@@ -0,0 +1,84 @@
...
+    golang="1.48.0",
+    java="?",
+    nodejs="3.15.0",
+    php="0.85.0",
+    php_appsec="0.7.0",
+    python="?",
+    ruby="?",
+)
+@irrelevant(
+    context.library == "java" and context.appsec_rules_file is not None,
```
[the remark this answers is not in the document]
smola: The spec mandates that we do not subscribe for ASM_DATA or announced IP blocking if there is a custom rules file. This is not specific to Java, and it _should_ fail for every language.

tests/appsec/test_blocking_maxed.py
```diff
@@ -0,0 +1,84 @@
+from utils import context, released, coverage, interfaces, scenarios, weblog, bug, irrelevant
+
+
+@released(
+    cpp="?",
+    dotnet="2.30.0",
+    golang="1.48.0",
+    java="?",
```
smola, line 8:
```suggestion
    java="1.7.0",
```

It should be complete for most variants on this release.

smola: But the following decorators should be added:
```
@missing_feature(context.weblog_variant == "spring-boot-native", reason="GraalVM. Tracing support only")
@missing_feature(context.weblog_variant == "spring-boot-3-native", reason="GraalVM. Tracing support only")
```

tests/appsec/test_blocking_maxed.py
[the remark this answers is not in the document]
M4TTRX: Done

tests/appsec/test_user_blocking_maxed.py
```diff
@@ -0,0 +1,47 @@
+from utils import released, coverage, interfaces, scenarios, weblog
+
+
+@released(
+    cpp="?",
+    dotnet="2.30.0",
+    golang="1.48.0",
+    java="?",
+    nodejs="?",
```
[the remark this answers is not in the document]
M4TTRX: done

tests/appsec/test_user_blocking_maxed.py
M4TTRX: done

tests/appsec/test_blocking_maxed.py
M4TTRX: done, I hope I put the decorators properly

M4TTRX:
@simon-id @cbeauchesne, thank you for your reviews, I agree with you, so I reworked the "IP blocking maxed" test so that half of the denylist entries are ips and the other half are users. Let me know if this works for you.

Also, I might try to put more than 2500 total entries if everything works.

Also is is it possible to force the CI to run my scenario only ? (+ default scenario)

tests/appsec/test_blocking_full_denylist.py
```diff
@@ -0,0 +1,82 @@
...
+    java="1.7.0",
+    nodejs="3.15.0",
+    php="0.85.0",
+    php_appsec="0.7.0",
+    python="1.12.1",
+    ruby="?",
+)
+@missing_feature(context.weblog_variant == "spring-boot-native", reason="GraalVM. Tracing support only")
+@missing_feature(context.weblog_variant == "spring-boot-3-native", reason="GraalVM. Tracing support only")
+@irrelevant(context.appsec_rules_file is not None)
```
cbeauchesne, line 17: useless, as in this scneario, appsec_rules_file is always `None`

tests/appsec/test_blocking_full_denylist.py
```diff
@@ -0,0 +1,82 @@
...
+                assert span["meta"]["appsec.event"] == "true"
+                assert span["meta"]["appsec.blocked"] == "true"
+                assert span["meta"]["http.status_code"] == "403"
+                return True
+
+            assert self.r_block.status_code == 403
+            interfaces.library.assert_waf_attack(self.r_block, rule="block-users", address="usr.id")
+            interfaces.library.validate_spans(self.r_block, validator=validate_blocking_test)
+
+        def test_blocked_and_unblocked_ips(self):
```
cbeauchesne, line 56: this nested function is never called.

tests/appsec/test_blocking_full_denylist.py
```diff
@@ -0,0 +1,82 @@
...
+        interfaces.library.validate_spans(self.r_nonblock, validator=validate_nonblock_user)
+        interfaces.library.assert_no_appsec_event(self.r_nonblock)
+
+    def test_blocking_test(self):
+        """Test with a denylisted user"""
+
+        def validate_blocking_test(span):
+            """Check all fields are present in meta"""
+            for i in range(1250):
+                assert span["meta"]["usr.id"] == str(i)
```
cbeauchesne, line 46: I may have missed something, but what you are doing is : 

```python
assert span["meta"]["usr.id"] == "0" and  span["meta"]["usr.id"] == "1"
```

So it can never be ok

cbeauchesne (changes requested):
you removed `test_rc_protocol` and

tests/appsec/test_ip_blocking_full_denylist.py
```diff
@@ -56,16 +55,7 @@ def validate(data):
...
-            "245.159.180.146",
-            "126.29.6.251",
-            "59.9.47.235",
-            "179.140.243.163",
-            "136.26.224.167",
-            "165.160.156.36",
-            "172.2.59.155",
-            "169.49.236.135",
-        ]
+        BLOCKED_IPS = [f"12.8.{a}.{b}" for a in range(10) for b in range(125)]
```
cbeauchesne, line 58: Could you just add a comment explaining this list, the reason for `10` and `250`, and where to look for its equivalent in remote json files / how to modify them ?

tests/appsec/test_user_blocking_full_denylist.py
```diff
@@ -0,0 +1,52 @@
+from utils import context, released, coverage, interfaces, scenarios, weblog, bug, irrelevant, missing_feature
+
+
+@released(
+    cpp="?",
+    dotnet="2.30.0",
+    golang="1.48.0",
+    java="1.7.0",
+    nodejs="3.15.0",
+    php="0.85.0",
```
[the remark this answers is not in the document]
Anilm3: The recommended version for PHP user blocking is 0.86.3

tests/appsec/test_user_blocking_full_denylist.py
```diff
@@ -0,0 +1,52 @@
...
+    nodejs="3.15.0",
+    php="0.86.3",
+    php_appsec="0.7.2",
+    python={"django-poc": "1.10", "flask-poc": "1.10", "*": "?"},
+    ruby="?",
+)
+@missing_feature(library="java", reason="/users endpoint is not implemented on java weblog")
+@bug(library="php")
+@bug(library="php_appsec")
+@bug(library="golang", reason="Doesn't work for versions <= 1.50.0")
```
cbeauchesne, line 18: if it works for version higher than 1.50.0 then 

```suggestion
@bug(context.library <= "golang@1.50.0", reason="Doesn't work for versions <= 1.50.0")
```

tests/appsec/test_user_blocking_full_denylist.py
```diff
@@ -0,0 +1,52 @@
...
+    java="1.7.0",
+    nodejs="3.15.0",
+    php="0.86.3",
+    php_appsec="0.7.2",
+    python={"django-poc": "1.10", "flask-poc": "1.10", "*": "?"},
+    ruby="?",
+)
+@missing_feature(library="java", reason="/users endpoint is not implemented on java weblog")
+@bug(library="php")
+@bug(library="php_appsec")
```
cbeauchesne, line 17: `php_appsec` is not a library, you can removes this line, `@bug(library="php")` does the job.

cbeauchesne (changes requested):

tests/appsec/test_user_blocking_full_denylist.py
```diff
@@ -0,0 +1,86 @@
...
+    golang="1.48.0",
+    java="1.7.0",
+    nodejs="3.15.0",
+    php="0.86.3",
+    php_appsec="0.7.2",
+    python={"django-poc": "1.10", "flask-poc": "1.10", "*": "?"},
+    ruby="?",
+)
+@missing_feature(library="java", reason="/users endpoint is not implemented on java weblog")
+@bug(library="php")
```
Anilm3, line 16: When testing on PHP, I noticed that we receive two remote config updates targeting the same config `datadog/2/ASM_DATA/ASM_DATA-base/config`, the first one containing a list of users and the second one containing only the list of IPs. 

This causes the first configuration to be replaced by the second, since they are the same configuration, hence why it fails in PHP.

-

M4TTRX: interesting, thank you

that should cause issues with other tests as well I guess

Anilm3:
I was investigating the latest issues on PHP, we seem to have a slight race condition, as the remote config update is received after the request is performed. Any idea how we can make the tests synchronous with the RC update @cbeauchesne?

cbeauchesne:
@Anilm3 those two lines are responsibles to wait for the tracers actively saying `done` to the backend : https://github.com/DataDog/system-tests/pull/1117/files#diff-c794a89a39c16dd2fa1b1cb90dc4955818677633c062338e3bac8b8ee35799e3R42-R43

cbeauchesne (changes requested):

cbeauchesne (approved):

PR #1118, opened 2023-04-26
Clean startup sequence

cbeauchesne:
Clean startup sequence to prepare integration of parametric tests


robertomonteromiguel (approved):
Ok. It's working for a new scenario without weblog container

PR #1119, opened 2023-04-26
fix systems test for DSM in Java since partition was removed as a tag

hokitam:


tests/integrations/test_dsm.py
```diff
@@ -5,7 +5,10 @@
 from utils import weblog, interfaces, context, bug, missing_feature, scenarios
 
 
-@missing_feature(condition=context.library != "java", reason="Full Kafka instrumentation only on Java")
+@missing_feature(
```
smola, line 8: We usually express this with the `@released` decorator at class level. For example:
```
@released(python="?", cpp="?", golang="?", java="1.12.0", dotnet="?", nodejs="?", php="?", ruby="?")
```

cbeauchesne:
Ok, I've adressed comment from @smola .

I also took the liberty to rewrite the assertion for a better output, as it was hard to understand whats going on. Thanks to that, I observe two failures : 


**Kafka check**
image

-> You are looking for tag `"partition:0"`, which is not present.

**rabbitmq check**
image

-> I strongly suspect that the check for `direction:in` has been copy pasted from kafka one, it's the same. But there is not `direction:in / type:rabbitmq` reported ?

cbeauchesne:
@smola the internal flush period on DSM is 10 seconds, I had to raise the wait period to 25 seconds on java (we changed it to 10 seconds yesterday).

cbeauchesne (approved):

PR #1122, opened 2023-04-27
Activate suspicious request blocking tests

robertpi:
## Description

Activate suspicious request blocking tests.

## Motivation

Not much point having deactivated tests.


robertpi:
> `APPSEC_BLOCKING`

yes, forgot these tests where part of a scenario, should be good now.

PR #1133, opened 2023-05-02
Add set of scenario feature

cbeauchesne:
## Description

Add feature `./run.sh SET_OF_SCENARIO`

As now, possible sets are `APPSEC_SCENARIOS`, `REMOTE_CONFIG_SCENARIOS`, `TELEMETRY_SCENARIOS` and `TRACER_RELEASE_SCENARIOS`

## Motivation

Reduce maintenance of CIs


cbeauchesne:
IS it really what we want? in CI, it will prevent running scenarios after a fail, and prevent any parallelism ...

PR #1142, opened 2023-05-03
Test new HTML blocking template

smola:
## Description

* Test new HTML blocking template
* Make other blocking template tests more strict, comparing to a fixed set of known versions of the blocking template.


simon-id:
If you look at the indentation on the old html template file you can see it's wrong from the start because someone randomly removed a level of indentation instead of copying the raw file from the repo, this makes me SO mad. Anyway thank you for fixing it here so I don't have to fix it on my PR :D

tests/appsec/waf/blocked.v1.html
```diff
@@ -0,0 +1,99 @@
+<!-- Sorry, you’ve been blocked -->
```
simon-id, line 1: Why is this not updated ? it still has the weird backtick

tests/appsec/waf/test_blocking.py
```diff
@@ -8,106 +9,27 @@
...
-      flex-direction: column
-    }
 
-    p {
-      display: block
-    }
+_CUR_DIR = os.path.dirname(os.path.abspath(__file__))
 
+# Initial template version as found in Java, with different indentation to v1.
+BLOCK_TEMPLATE_HTML_V0 = open(os.path.join(_CUR_DIR, "blocked.v0.html"), "r").read().rstrip()
```
simon-id, line 16: I don't think we should need a rstrip() everywhere here ? If we copy the exact file then by doing a rstrip() you run the risk to delete a character that was in the spec

tests/appsec/waf/blocked.v1.html
smola: See `blocked.v1.html`. I tagged in the repo `v1` as the versions that were in https://github.com/DataDog/asm-blocked-response-template just before the change, and `v2` the ones that just changed. And an additional `v0` because we had something prior in Java.

tests/appsec/waf/test_blocking.py
smola: I'll try removing it, I think that we have some inconsistency about ending with new line or not across tracers (that's why the test had `rstrip()` before.

We could probably be more lenient about previous versions of the blocking template, and more strict for `test_html_template_v2`? One of the principles I followed is that whatever is done for previous versions of the template and was not checked in the system-tests before, it's done and we shouldn't tag for it. And the increased strictness should apply only to new checks, since it has a new `@released` decorator.

smola: I have removed all `rstrip` calls. Unless this fails for a language where this test is already enabled, I think it shouldn't be a problem.

simon-id (approved):
LGTM but maybe have someone more familiar with python review it

tests/appsec/waf/test_blocking.py
```diff
@@ -8,106 +9,29 @@
...
-  <p>Security provided by <a
-    href="https://www.datadoghq.com/product/security-platform/application-security-monitoring/"
-    target="_blank">Datadog</a></p>
-</footer>
-</body>
-
-</html>
-"""
+HTML_CONTENT_TYPES = {"text/html", "text/html; charset=utf-8", "text/html;charset=utf-8"}
+JSON_CONTENT_TYPES = {"application/json", "application/json; charset=utf-8", "application/json;charset=utf-8"}
```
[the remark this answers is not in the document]
smola: It is still valid to add the charset (some frameworks add it automatically, some libraries specifically add it), and I can't see any reason to make these tests fails in these cases.

tests/appsec/waf/test_blocking.py
```diff
@@ -244,10 +181,32 @@ def setup_accept_full_html(self):
...
+        assert self.r_json_v1.status_code == 403
+        assert self.r_json_v1.headers.get("content-type", "") in HTML_CONTENT_TYPES
+        assert self.r_json_v1.text == BLOCK_TEMPLATE_JSON_MIN_V1
+
+    def setup_html_template_v2(self):
+        self.r_html_v2 = weblog.get("/waf/", headers={"User-Agent": "Arachni/v1", "Accept": "text/html",},)
+
+    @released(java="?", dotnet="?", golang="?", nodejs="?", php_appsec="?", python="?", ruby="?")
+    def test_html_template_v2(self):
+        """HTML block template is v1 minified"""
```
smola, line 206:
```suggestion
        """HTML block template is v2 minified"""
```

PR #1143, opened 2023-05-03
feat(python): enable ip_blocking_maxed test

avara1986:
## Description

Enable IP blocking maxed test for Python

## Motivation
We're working on a [RC Refactor](https://github.com/DataDog/dd-trace-py/pull/5638#discussion_r1180114259) and we need to test big RC payloads


tests/appsec/test_ip_blocking_maxed.py
```diff
@@ -50,7 +51,12 @@ def validate(data):
 
             logger.info(f"validating rc request number {self.request_number}")
             rc_check_request(data, EXPECTED_REQUESTS[self.request_number], caching=True)
-            self.request_number += 1
+            self.python_request_number += 1
+            if context.library == "python" and context.weblog_variant != "uwsgi-poc":
```
smola, line 55: Is this expected behavior or a bug? If it's a bug, I would mark the test as `@bug` for that variant rather than adding custom logic.

avara1986: @smola there is no bug, it's the Gunicorn behavior, we added this logic in other tests like:
https://github.com/DataDog/system-tests/pull/932/files#diff-41b8e87a63e1ce2f325503a9061a569a17dd2de78bd61d52238c7dcb765dd2b2R211

We're working in a [refactor](https://github.com/DataDog/dd-trace-py/pull/5464) to fix this Gunicorn behavior

smola (approved):
I'm not particularly fond of the Python-specific logic here, but I guess it's reasonable.

juanjux (approved):

cbeauchesne (approved):

PR #1154, opened 2023-05-05
Fix telemetry test (uniquely identify requests)

smola:
## Description

runtime_id/tracer_time is not a unique key to a telemetry request. Since tracer_time has a resolution of 1 second, two consecutive requests can have the same tracer_time.

So use runtime_id/seq_id as a unique identifier.

## Motivation

Telemetry test sometimes fails.


avara1986 (approved):

avara1986:
LGTM!

cbeauchesne (approved):

PR #1157, opened 2023-05-05
Decouple proxy ingestion and interface waiting

smola:
## Motivation

Rather than stopping ingestion on each interface after each wait, we can separate both concerns by waiting for all conditions, then stop all watchdog events at once.

### DIFF
--- a/utils/_context/_scenarios.py
+++ b/utils/_context/_scenarios.py
@@ -169,6 +169,22 @@ class TestTheTestScenario(_Scenario):
         return "spring"
 
 
+class IngestFileEvent(FileSystemEventHandler):
+    """
+    Event handler for file ingestion into interfaces.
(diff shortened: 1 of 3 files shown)


cbeauchesne (commented):
Hi,

I'm not sure to understand the motivation of this change.

All the purpose of chained wait/stop ingestion is a feature. There is a chain of component `a -> b -> c`; by chain, I mean the data flow is globally unidirectional, `a` sends to `b`, `b` sends to `c`.
 
The idea is that we wait for `a`, then actively stops `a` and then wait for `b`, etc

So every component will have it's grace time to process all event received.

If we stops all watchdogs at once, we can have an event sent by `a`, that we won't see in `b`, if for any reason it kept it i a queue process.

PR #1163, opened 2023-05-10
Remove library versions for Test_Basic

simon-id:
## Description

This remove the library versions decorator for the Test_Basic class as it's testing the weblog app, not the library. Hence having a library version makes no sense.


cbeauchesne:
indeed

cbeauchesne (approved):

PR #1176, opened 2023-05-12
Add tests for IAST telemetry

smola:
## Description

Enable telemetry by default in weblogs, and just add these to the default scenario.


cbeauchesne:
As weblog env is modified, I prefer running everything ...

tests/appsec/iast/sink/test_command_injection.py
```diff
@@ -49,6 +49,24 @@ def test_insecure(self):
...
 
-    @missing_feature(reason="Endpoint not implemented")
+    @missing_feature(library="nodejs", reason="Endpoint not implemented")
     def test_secure(self):
         self.sink_fixture.test_secure()
+
+    def setup_telemetry_metric_instrumented_sink(self):
+        self.sink_fixture.setup_telemetry_metric_instrumented_sink()
+
+    @released(dotnet="?", golang="?", java="1.13.0", nodejs="?", php_appsec="?", python="?", ruby="?")
```
cbeauchesne, line 59: `released` decorator can only go on classes. For method, you can use directly `missing_feature` (same comment for other occurences)

cbeauchesne (changes requested):

tests/appsec/iast/sink/test_command_injection.py
smola: @cbeauchesne Ok. I'll change it. However, I have used this pattern before at https://github.com/DataDog/system-tests/pull/1173/files and it seemed to work?

manuel-alvarez-alvarez:
Propagation metrics are missing from the test, but to be honest I often wonder if we should remove them from IAST.

manuel-alvarez-alvarez (approved):

smola:
> Propagation metrics are missing from the test, but to be honest I often wonder if we should remove them from IAST.

I went for the low hanging-fruit of what can be tested with the current endpoints. But I guess we should add at least one propagation case... at some point.

tests/appsec/iast/sink/test_command_injection.py
cbeauchesne: it works, but it's an issue :) 

Basically, the key idea is we export the information feature/released version in different dashboard spreadsheet. And it only work at feature level, which is represented by classes in system tests.

Not a big deal though, it will just save me some time when I'll fix this small API inconsistency.

cbeauchesne (approved):

PR #1187, opened 2023-05-17
pr-stats

pawelchcki:


cbeauchesne:
@pawelchcki I just look on the output and saw this : 

image

I'd be great  if we can have a monthly slack message for this report, WDYT ?

PR #1188, opened 2023-05-17
[opentelemetry] Add basic OTel metric and log E2E tests

songy23:
## Description

APM R&D week work.

- OTEL_METRIC_E2E generates 1 delta counter metric + 1 delta histogram metric. Test runner calls `api/v1/query` to get the metric timeseries from backend and validates them.
- OTEL_LOG_E2E generates 1 OTLP log record within a trace scope. Test runner calls `api/v2/logs/events` to get the log from backend then validates the log as well as log-trace correlation.

## Motivation

Extend the E2E test coverage to all telemetry signals in OTel.


tests/otel_tracing_e2e/_validator_log.py
```diff
@@ -0,0 +1,35 @@
...
+    assert expected_attributes_attributes.items() <= log["attributes"]["attributes"].items()
+    return log["attributes"]["attributes"]["otel"]
+
+
+def validate_log_trace_correlation(otel_log_trace_attrs: dict, trace: dict):
+    assert len(trace["spans"]) == 1
+    span = None
+    for item in trace["spans"].items():
+        span = item[1]
+    assert otel_log_trace_attrs["trace_id"] == span["meta"]["otel.trace_id"]
```
mackjmr, line 33: I believe the web-ui uses endpoint `/api/v1/logs-analytics/list?type=logs` with query string `trace_id:<trace_id>` ([code](https://github.com/DataDog/web-ui/blob/877877812b1c25a05690799cdd8ce0a5b23f91f3/javascript/datadog/trace/components/trace/SpanLogs/span-logs-selectors.ts#L30)) in order to pull the logs from the trace (Tried to confirm this in this [thread](https://dd.slack.com/archives/CP2BYH4BU/p1682515661494809) with the trace-app team). The log attribute `["attributes"]["attributes"]["otel"]["trace_id"]` which we are testing here isn't used for trace/log correlation. 

In order to validate trace/log correlation, I think we need to make sure that we can query the logs with the `trace_id`. In order to do so, I think using `/api/v2/logs/events?query=` instead of `/api/v1/logs-analytics/list?type=logs` is fine as long as the query is the same (`trace_id:<trace_id>`).

Although the hidden `trace_id` is not present in the log when fetching it through the API, we can still use in in the query.

Side Note: for now, because this test uses the collector, the `trace_id` gets converted from 128 bit to 64 bit (see [code](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.76.3/exporter/datadogexporter/internal/logs/translator.go#L124-L127)) so the query for `trace_id` should search for the 64 bit `trace_id`. That being said, once the backend/UI get's updated to fully support 128 bit trace_id we will be able remove the conversion in the collector ([task](https://datadoghq.atlassian.net/browse/OTEL-561)), and then the query will need to be made with the 128 bit trace_id.

songy23: That makes sense and it's actually a better query. I updated the log query to be based on trace id. 

There's a helper function `_get_dd_trace_id` that takes into account 128-bit vs. 64-bit. We can simply change `use_128_bits_trace_id` to true when 128-bit is introduced.

mackjmr (approved):

cbeauchesne (approved):

songy23:
Thanks for the review!

PR #1190, opened 2023-05-18
parametric conftest migration

robertomonteromiguel:


cbeauchesne (approved):

robertomonteromiguel:
> @robertomonteromiguel Can you add a motivation and more information about this impact this will have on development? System tests are notoriously more difficult to use than parametric tests, so if anything I would expect the opposite migration.

What we are trying to do it's unify all types of tests that are included on system-tests. The idea is to launch all of them in same way and extract results in the same way and report these results to the "feature parity dashboard"

PR #1214, opened 2023-05-26
Added /dsm endpoint for node.js

ida613:


tests/integrations/test_dsm.py
```diff
@@ -17,16 +18,23 @@ def setup_dsm_kafka(self):
         self.r = weblog.get("/dsm?integration=kafka")
 
     def test_dsm_kafka(self):
+        print(self.r)
+        print(self.r.content)
+        logger.info(f"exectute dsm tests {self.r}, {self.r.content}")
+        logger.info(f"exectute dsm tests {self.r.status_code}")
+        print(self.r.status_code)
```
Qard, line 25: Is it intentional that you're leaving all this extra logging stuff in there?

tests/integrations/test_dsm.py
```diff
@@ -17,16 +18,23 @@ def setup_dsm_kafka(self):
...
         assert str(self.r.content, "UTF-8") == "ok"
 
+        consumer_hash = 5700134256996109392 if context.library == "nodejs" else 4463699290244539355
+        producer_hash = 13592025863925263556 if context.library == "nodejs" else 16645700936287432977
```
Qard, line 29: Unclear what these hashes are, but why are these different just for Node.js? Seems like .NET and Java were using the same hash already so I would think there should be a way for Node.js to do that too? Or is there some context I'm missing here?

tests/integrations/test_dsm.py
```diff
@@ -17,16 +18,23 @@ def setup_dsm_kafka(self):
         self.r = weblog.get("/dsm?integration=kafka")
 
     def test_dsm_kafka(self):
+        print(self.r)
```
tlhunter, line 21: Should these three `print()` statements be removed?

tests/integrations/test_dsm.py
```diff
@@ -76,6 +84,7 @@ def assert_checkpoint_presence(hash_, parent_hash, tags):
         logger.info(f"Look for {hash_}, {parent_hash}, {tags}")
 
         for data in interfaces.agent.get_dsm_data():
+            logger.info(data)
```
tlhunter, line 87: This should probably have a message associated with it? Otherwise it's just an object without explanation

utils/build/docker/nodejs/express4/app.js
```diff
@@ -134,13 +135,64 @@ app.get("/users", (req, res) => {
...
+app.get("/dsm", (req, res) => {
+  const kafka = new Kafka({
+    clientId: 'my-app',
+    brokers: ['kafka:9092'],
+    retry: {
+      initialRetryTime: 100, // Time to wait in milliseconds before the first retry
+      retries: 20, // Number of retries before giving up
+    },
+  })
+  console.log("in endpoint")
```
tlhunter, line 147: `console.log`s in the sample app are okay but this looks like leftover debugging code.

utils/build/docker/nodejs/express4/app.js
```diff
@@ -134,13 +135,64 @@ app.get("/users", (req, res) => {
...
 
+app.get("/dsm", (req, res) => {
+  const kafka = new Kafka({
+    clientId: 'my-app',
+    brokers: ['kafka:9092'],
+    retry: {
+      initialRetryTime: 100, // Time to wait in milliseconds before the first retry
+      retries: 20, // Number of retries before giving up
+    },
+  })
```
simon-id, line 146: is this supposed to be executed for every request ? shouldn't it be initialized at the top level ? I'm not sure tbh i never ~read~ used kafka

simon-id (approved):

simon-id (changes requested):
i approved by mistake

cbeauchesne:
@ida613  Could you resolves conflicts  before I can review it please ?

tests/integrations/test_dsm.py
ida613: This is the hashing we are using to identify consumers and producers. The other libraries are using FNV-1 hashing, we couldn't find a FNV-1 library for node, so we are using FNV-1a instead, which gives a different hash. Piotr said it's not an issue.

tests/integrations/test_dsm.py
```diff
@@ -2,13 +2,14 @@
...
-from utils import weblog, interfaces, scenarios, released
+from utils import weblog, interfaces, scenarios, released, context
 from utils.tools import logger
 
 
-@released(cpp="?", golang="?", nodejs="?", php="?", python="?", ruby="?")
+@released(cpp="?", golang="?", php="?", python="?", ruby="?")
 @released(dotnet="2.29.0")
 @released(java={"spring-boot": "1.13.0", "*": "?"})
+@released(nodejs="4.0.0-pre")
```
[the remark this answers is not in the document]
ida613: Thanks! I'm gonna remove `@released(nodejs=version)` for now and will add it in after we do a release for dsm

ida613:
There is too much code to rebase, I'm gonna create a new fork and manually add the changes there

PR #1217, opened 2023-05-30
go: enable html/json blocking response template tests

Hellzy:


cbeauchesne:
Removing label `run-default-scenario` in order to run all scenario.

PR #1223, opened 2023-06-01
Move all ruby apps inside the system test repo

GustavoCaso:
## Description

This PR moves all the ruby apps from https://github.com/DataDog/system-tests-apps-ruby inside of the system tests repo. Following a similar path as the rest of the other libraries. 

Reasons:
- Is easier to test changes to the weblog variants for Ruby. Today we have to make the changes to a separate repo, validate those changes work, merge and then update the system tests repo to enable tests or other stuff for ruby
- The ASM ruby can not keep up with all repos we maintain, so making this easier for us is important
- Keeping multiple pipelines up to date with the system test repo is hard. The https://github.com/DataDog/system-tests-apps-ruby system test pipeline brakes often and takes time to fix it

How I moved the apps:
- Copy the code from https://github.com/DataDog/system-tests-apps-ruby
- Update the different docker files inside `utils/build/docker/ruby/*.Dockerfile` 
   - Use the correct image from `ghcr.io/datadog/dd-trace-rb/ruby`
   - Copy the ruby app into the workdir 

If everything works as expected, CI should be 🟢 

To ease of review, please filter out ruby files `*.rb` [Link to filter view](https://github.com/DataDog/system-tests/pull/1223/files?file-filters%5B%5D=.Dockerfile&show-viewed-files=true). We only care about making sure the Docker files are correct 😄


lloeki:
This is a significant rollback from having the docker images prebuilt.

```
#12 DONE 60.2s
```

Notably, the above is _just_ the bundle install step and adds a full minute (so, times the number of apps, times scenarios when a proper matrix is used - see dd-trace-rb repo) to every run, for things that rarely change (the weblog API being mostly static, as well as the dependencies).

I understand the frustration leading to this but I believe it is detrimental to speedy and reliable CI in the long run.

cbeauchesne:
We can maybe keep the docker caching mechanism, but integrate in this repo ? 

Actually, I think it may be the best solution, as I'm willing to work on pre-build images for all weblogs, and the job @lloeki did mostly do exactly what I want to do, so it'll save me lot of time.

lloeki (approved):
Approved, provided performance concerns are taken into account.

PR #1225, opened 2023-06-01
improve suspicious request blocking test on uri raw

christophe-papazian:
## Description

Add a rule to block on uri raw address that starts with `http` and check that this rule is not used to block on the uri raw value (as it should never be prefixed with the base URL)

Also add a missing_feature tag for a non implemented test.

### DIFF
--- a/tests/appsec/blocking_rule.json
+++ b/tests/appsec/blocking_rule.json
@@ -381,6 +381,32 @@
       "on_match": [
         "block"
       ]
+    },
+    {
+      "id": "tst-037-011",
+      "name": "Test block on uri that should not block",
+      "tags": {
+        "type": "lfi",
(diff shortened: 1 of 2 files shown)


smola (approved):

tests/appsec/blocking_rule.json
```diff
@@ -381,6 +381,32 @@
...
+      },
+      "conditions": [
+        {
+          "parameters": {
+            "inputs": [
+              {
+                "address": "server.request.uri.raw"
+              }
+            ],
+            "regex": "^http"
```
cbeauchesne, line 401: IIUC, it mean that it blocks if `server.request.uri.raw` starts with `http` ? So basically all requests?

PR #1291, opened 2023-06-26
fix json report generation when we use multiple pytest workers

robertomonteromiguel:
## Description

When we activated multiple pytests workers, the json report modification doesn't work.
1. Fix "collectors" node deletion from json report. When we use multiple workers, this node doesn't exist.
2. The global variables of conftests.py are empty when we call to the method "pytest_json_modifyreport". Workaround: Export variables to json file before this method is called. Inside this method we load the previously exported json.

## Reviewer checklist

* [ ] If this PR modifies anything else than strictly the default scenario, then remove the `run-default-scenario` label
* [ ] CI is green

## Workflow

4. Follow the style guidelines of this project (See [how to easily lint the code](https://github.com/DataDog/system-tests/blob/main/docs/edit/lint.md))
5. Work on you PR until the CI passes (if something not related to your task is failing, you can ignore it)
6. Mark it as ready for review

> **_NOTE:_**  By default in PR only default scenario tests will be launched. Remove the `run-default-scenario` label to run all scenarios ([more info](https://datadoghq.atlassian.net/wiki/spaces/APMINT/pages/2866381467/CI+Workflow+Github+Actions))

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -11,10 +11,10 @@ from utils._context._scenarios import scenarios
 from utils.tools import logger
 from utils.scripts.junit_report import junit_modifyreport
 from utils._context.library_version import LibraryVersion
-from types import SimpleNamespace
 
 # Monkey patch JSON-report plugin to avoid noise in report
 JSONReport.pytest_terminal_summary = lambda *args, **kwargs: None
+
 
 class ReportMetadata:
     def __init__(self):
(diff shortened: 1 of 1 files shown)


cbeauchesne (changes requested):
Can we activate multiple worker in this CI on parametric tests : it will be a convenient way to ensure we don't break anything in the future.

PR #1309, opened 2023-06-28
Prevent replay failures because of ingestion race

smola:
## Description

Apply the same deserialization logic when loading requests from logs (for replay mode) as when running the full scenario. This ensures that the request objects matches. This should generally not be a problem unless this step is missed in the initial ingest (which this PR does not fix), but applying the same deserialization to both scenarios will make issues less likely.

## Motivation

Avoid some issues in replay mode under race conditions

Not part of the PR: fix the actual race condition (I'll make a separate PR for that).

## Reviewer checklist

* [x] If this PR modifies anything else than strictly the default scenario, then remove the `run-default-scenario` label
* [ ] CI is green

## Workflow

> **_NOTE:_**  By default in PR only default scenario tests will be launched. Remove the `run-default-scenario` label to run all scenarios ([more info](https://datadoghq.atlassian.net/wiki/spaces/APMINT/pages/2866381467/CI+Workflow+Github+Actions))


utils/interfaces/_core.py
```diff
@@ -88,6 +89,13 @@ def load_data_from_logs(self, folder_path):
                 with open(file_path, "r", encoding="utf-8") as f:
                     data = json.load(f)
 
+                # Apply the same deserialization process as in ingestion, since
+                # there may be a race condition where additional messages are written
+                # post-setup without prior normalization.
+                # XXX: A more robust solution would be moving the serialization process
+                # to the proxy, to avoid the need to rewrite request files in the first place.
+                deserialize(data, self.name)
```
cbeauchesne, line 97: At this point, data may be deserialized. You deserialize it twice no ?

utils/interfaces/_core.py
```diff
@@ -63,6 +63,7 @@ def ingest_file(self, src_path):
                 try:
                     data = json.load(f)
                 except json.decoder.JSONDecodeError:
+                    logger.exception(f"Error while decoding message {src_path}")
```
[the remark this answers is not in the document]
smola: The case where it's a transient error (partial write), seems to be a fairly infrequent error. The poller has some delay, so whenever this happens because of a partial write, it's usually done in next iteration.
AFAIK the next iteration is currently 1 second (lower with the next PR though).

utils/interfaces/_core.py
smola: As far as I can tell, `deserialize` is idempotent. So there's no harm in doing it twice. On the other hand, even if it was deserialized before, the `data` as read from the object vs the `data` after calling `deserialize` is not the same. For example, byte arrays will be strings like this `b'mybytestrict'` in the JSON, but after calling `deserialize`, they will be converted to actual byte arrays.

smola:
As discussed with @cbeauchesne, I'm closing this to avoid expanding the hacks in more places. There'll be a much better solution in the future :smile:

PR #1314, opened 2023-06-28
Remove redundancy in telemetry activation

smola:
## Description

We were setting `DD_INSTRUMENTATION_TELEMETRY_ENABLED=true` for all weblogs, and then `DD_INSTRUMENTATION_TELEMETRY_ENABLED=1`, which was redundant.

The later did not work in previous version, but the former did. So it seems we don't need the @bug clarification for Ruby.

## Reviewer checklist

* [ ] CI is green

## Workflow


tests/test_telemetry.py
```diff
@@ -7,10 +7,6 @@
 
 @released(python="1.7.0", dotnet="2.12.0", java="0.108.1", nodejs="3.2.0", ruby="1.4.0", golang="1.49.0")
 @bug(context.uds_mode and context.library < "nodejs@3.7.0")
-@bug(
```
cbeauchesne, line 10: You must keep the version where it has been fixed. Otherwise, test on old travers will fail.

cbeauchesne (changes requested):

tests/test_telemetry.py
smola: The test on old versions should now work, since they failed when `DD_INSTRUMENTATION_TELEMETRY_ENABLED=1` was used, but `DD_INSTRUMENTATION_TELEMETRY_ENABLED=true`  worked. Given that this PR uses `true` everywhere, this bug should not be triggered by the test.

cbeauchesne (approved):

PR #1326, opened 2023-06-30
Deserialize in proxy

cbeauchesne:
## Description

Move the deserializer inside the proxy

## Motivation

Save the complexity of the ingester in the runner. Previously 

1. proxy save the message on filesystem, with contents as byte arrray serialized to a string
2. runner load the `json` file, use `ast.literal_eval` to convert the content to bytes array, deserialize it, save it filesystem

It means that the data comes threw those representations :  `proxy(bytes -> str -> json) -> runner(str -> bytes -> python object -> json)`
It also means that any file not ingested is in a temporary state that make the replay mode fails a lot

Now, data lifecycle is : `proxy(bytes -> python object -> json) -> runner(python object)`

It'll also solve some issues, as we don't require grpc in the runner, which was blocking updates of python and packages.

## Reviewer checklist

* [ ] CI is green

## Workflow


requirements.txt
```diff
@@ -1,21 +1,12 @@
...
 pytest-json-report==1.5.0
 black==19.10b0  # TODO : move to 22.8.0
 python-dateutil==2.8.2
 msgpack==1.0.4
 watchdog==3.0.0
 
-# we'd like to update protobuf and grpcio-tools to those version, but a bug in protobuf 4.x need to be solved before
-# https://github.com/protocolbuffers/protobuf/issues/11863
-# protobuf==4.22.1
-# grpcio-tools==1.51.3
```
smola, line 14: :tada:

utils/proxy/core.py
```diff
@@ -104,67 +105,68 @@ def request_is_from_tracer(request):
...
+
+            logger.info(f"    => Saving data as {log_filename}")
 
-            with open(log_filename, "w", encoding="utf-8", opener=opener) as f:
+            with open(log_filename, "w", encoding="utf-8", opener=lambda path, flags: os.open(path, flags, 0o777)) as f:
                 json.dump(data, f, indent=2, cls=ObjectDumpEncoder)
 
         except:
-            logger.exception("Fail to save data")
+            logger.exception("Unexpcted error")
```
smola, line 169:
```suggestion
            logger.exception("Unexpected error")
```

smola (commented):
Great addition, thanks!

There's also the rare event when the proxy dies in the middle of a write. Although it should not be a problem if it's stopped a bit after both weblog and agent have stopped, right?

cbeauchesne:
> There's also the rare event when the proxy dies in the middle of a write. Although it should not be a problem if it's stopped a bit after both weblog and agent have stopped, right?

The proxy is shut down after all other images, so I would say it's very unlikely to happen (though still possible). Let's wait for this to happen at least once before thinking a way to fix it.

PR #1338, opened 2023-07-03
RFC: manifest file

cbeauchesne:
This PR will be used for the very first iteration of manifest file.

The goal is to provide a alternative way to declare versions, using one file per components.

### DIFF
--- a/docs/edit/manifest.md
+++ b/docs/edit/manifest.md
@@ -70,7 +70,7 @@ tests/:
         django: v1.2
         flask: v1.3
         uwsgi: bug (jira ticket) # For a weblog, skip it with bug, or flaky
-        "*": "?" # All other weblogs: not yet available
+        "*": missing_feature # All other weblogs: not yet available
 ```
 
 ## Format [WIP]
@@ -87,8 +87,8 @@ tests/:
     "folder_content": {
       "type": "object",
       "patternProperties": {
-        ".*/": { "$ref": "#/$defs/folder_content" },
-        "test_.*\\.py": { "$ref": "#/$defs/file_content" }
+        ".+/": { "$ref": "#/$defs/folder_content" },
+        "test_.+\\.py": { "$ref": "#/$defs/file_content" }
       }
     },
 
@@ -97,13 +97,15 @@ tests/:
         {
           "type": "object",
           "patternProperties": {
-            "Test_.*": {
+            "Test_.+": {
+              "$comment": "It can be a version number, a skip reason, or an object with weblog variant as keys",
               "oneOf": [
                 { "$ref": "#/$defs/version" },
                 {
                   "type": "object",
+                  "$comment": "Keys are weblog variant names, values are version, or a skip reason",
                   "patternProperties": {
-                    ".*": { "$ref": "#/$defs/class_content" }
+                    ".+": { "$ref": "#/$defs/class_content" }
                   }
                 }
               ]
@@ -123,11 +125,11 @@ tests/:
       ]
     },
 
-    "version": { "type": "string", "pattern": "(v.*|\\?)" },
+    "version": { "type": "string", "pattern": "v.+" },
 
     "skipped_declaration": {
       "type": "string",
-      "pattern": "^(bug|flaky|irrelevant) \\(.*\\)$"
+      "pattern": "^(bug|flaky|irrelevant|missing_feature)( \\(.+\\))?$"
     }
   }
 }
--- a/docs/edit/manifest.md
+++ b/docs/edit/manifest.md
@@ -1,0 +1,130 @@
+### Context
+
+The usual way to declare what will be tested or not are decorators (`@released`, `@bug`, `@missing_feature`...). This solution offers several advantages :
+
+- declarations are as close as possible to the test, making the link obvious
+- and complex condition (several components involved, complex version range...) is easy to do, as it's declared as python code
+
+### Issue
+
+Unfortunatly, it comes with a major drawback: Activating a single test can become a hell due to confilct between PRs.
+
+### Proposition
+
+To solve this, we'll try to leverage two points :
+
+1. Often, those declarations involve only one component
+2. Often, a component is modified by only one developper
+
+The idea is to offer an alternative way to declare those metadata, using one file per component, also known as manifest file. Those files will be inside system tests repo, declaring for a given component some metadata.
+
+Pro :
+
+- Will solve the majority of conflict during test activation PRs
+- each team will have ownership of its manifest file, easing the PR preview process
+
+Cost :
+
+- declarations will be a little bit far away from the test, asking a small addition in the learning curve
+- re-naming will be a little bit harder, and will spam all teams
+
+## Implementation
+
+- Manifests files will be inside `utils/manifest/`, one per term, named with the argument name used in decorators (`utils/manifest/golang.yml`, `utils/manifest/agent.yml`...)
+- Each team will have ownership of its manifest file
+- pytest will load those files, and decorate tests node during collection
+
+## Supported features
+
+Will suport:
+
+- declaring a skip reason (bug, flaky, irrelevant, missing_feature) for entire file
+- declaring released version (or a skip reason) for a class
+- declaring released version (or a skip reason) for a class, with details for weblog variants
+- The wildcard `*` is supported for weblog declarations
+
+Won't support:
+
+- declaring metadata (bug, flaky, irrelevant) for test methods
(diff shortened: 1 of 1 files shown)


docs/edit/manifest.md
```diff
@@ -0,0 +1,130 @@
...
+## Supported features
+
+Will suport:
+
+- declaring a skip reason (bug, flaky, irrelevant, missing_feature) for entire file
+- declaring released version (or a skip reason) for a class
+- declaring released version (or a skip reason) for a class, with details for weblog variants
+- The wildcard `*` is supported for weblog declarations
+
+Won't support:
```
simon-id, line 46: Are the complex syntaxes like `rules_version > xxx`, `waf_version < xxx && lib_version == xxx` or `xxx < lib_version < yyy` supported ? the supported/not supported lists should precise a bit more on those cases i feel like ?

cbeauchesne: * `rules_version > xxx` -> yes, inside the manifest file dedicated to rules file
* `waf_version < xxx && lib_version == xxx` -> nope. Complex conditions will remain in the code (there is no limit to this complexity, so we need to draw a line based on  the ratio format simplicity / number of occurrences : supporting those use cases will complexify **a lot** the manifest format - it does not happen a lot, and mostly at method level which won't be supported.
* `xxx < lib_version < yyy` -> nope either, same reason

docs/edit/manifest.md
```diff
@@ -0,0 +1,130 @@
...
+
+```yaml
+tests/:
+  specific.py: irrelevant (see this link) # let skip  an entire file
+
+  appsec/: # more regular declarations:
+    test_distributed.py:
+      Test_FeatureA: v1.14 # declare a version for a class
+
+      Test_FeatureB: flaky # skip a class with bug, flaky, irrelevant ...
```
smola, line 63: No version qualifier for bug/missing_feature, etc?

docs/edit/manifest.md
```diff
@@ -0,0 +1,130 @@
...
+
+      Test_FeatureC: # declare a version for a class, depending on weblog
```
smola, line 65: What about method-level?

docs/edit/manifest.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Implementation
+
+- Manifests files will be inside `utils/manifest/`, one per term, named with the argument name used in decorators (`utils/manifest/golang.yml`, `utils/manifest/agent.yml`...)
+- Each team will have ownership of its manifest file
+- pytest will load those files, and decorate tests node during collection
+
+## Supported features
+
+Will suport:
```
smola, line 39: IMHO this should land together with a script that:
* Checks if manifest files are in sync with tests (e.g. not referencing missing classes).
* Ideally, also a way to add missing entries based on tests (or a way to list tests disabled for a language).

smola (commented):
The idea sounds good, however:
* I think that without proper tooling to validate these files, things may get worse. So let's keep an eye on tooling that helps getting things right, and not committing wrong manifests.
* Having different ways to declare file-level and class-level (manifest file) and test-level (in code) sounds a bit like keeping some of the bad things of both worlds?

docs/edit/manifest.md
cbeauchesne: > I think that without proper tooling to validate these files, things may get worse. So let's keep an eye on tooling that helps getting things right, and not committing wrong manifests.

-> https://github.com/DataDog/system-tests/pull/1338/commits/f09d34ca3f18e4d383907a5091fbc3b478fa57ec

> Having different ways to declare file-level and class-level (manifest file) and test-level (in code) sounds a bit like keeping some of the bad things of both worlds?

It's a compromise. The main issue with tests methods is that their names change a lot. Every time you change a test method name, you'll need to update all manifest files, spamming all teams. Lot of people complain already about being over notified by system tests. And as you can only have one declaration per component at method level, conflict is more managable.

Actually, I even hesitated to only allow version declaration in the manifest (no `bug`, `flaky`....). But as it did not complexify the data model, I thought it was a good middle point.

docs/edit/manifest.md
simon-id: - Oh I didn't realize there would be a manifest for the rules, was that in the document or should we add that precision ?
- I'm fine with more complex conditions to not be supported, but by reading this document I'm not really sure where that cutoff point is, is it possible to be more precise ?

cbeauchesne: > Oh I didn't realize there would be a manifest for the rules, was that in the document or should we add that precision ?

In a senss, yes, but I made it crystal clear in  https://github.com/DataDog/system-tests/pull/1338/commits/f65626e8cb884e28acdda1fd1240490116fe4c9d

> I'm fine with more complex conditions to not be supported, but by reading this document I'm not really sure where that cutoff point is, is it possible to be more precise ?

It's in the "Supported features" section, but I make it more precise here : https://github.com/DataDog/system-tests/pull/1338/commits/ecda023165d266963f42c4b2ffb7087daffc08bf

simon-id: `term` -> `team` ok makes more sense now xD

simon-id:
Is file level skipping a good idea ? the problem imo is that you might and up skipping test classes that are written after the file is skipped. If we skip classes individually instead of whole files, at least we have a chance to notice new tests in XPASS results for example.

docs/edit/manifest.md
```diff
@@ -0,0 +1,134 @@
...
+        django: v1.2
+        flask: v1.3
+        uwsgi: bug (jira ticket) # For a weblog, skip it with bug, or flaky
+        "*": "?" # All other weblogs: not yet available
+```
+
+## Format [WIP]
+
+```json
+{
```
simon-id, line 79: I'm wondering if we should replace all the  `.*` by `.+`

simon-id (approved):
Amazing! and just in case you didn't know @Hellzy worked on an implementation of that for innovation week.

cbeauchesne:
> Is file level skipping a good idea ? the problem imo is that you might and up skipping test classes that are written after the file is skipped. If we skip classes individually instead of whole files, at least we have a chance to notice new tests in XPASS results for example.

Great power, great responsabilities :) The use case for that is C++ that does not have any ASM feature, and I would be please if can avoid this [ugly skip](https://github.com/DataDog/system-tests/blob/main/tests/appsec/test_PII.py#L10-L11): 

```python
if context.library == "cpp":
```

But I agree, it should not be used with `bug` or `missing_feature`. We can either limit the usage for file to `irrelevant` with the format, or trust that it won't happen. WDYT is best ?

docs/edit/manifest.md
cbeauchesne: done !

cbeauchesne:
> Amazing! and just in case you didn't know @Hellzy worked on an implementation of that for innovation week.

It's actually the continuation of its work (but I chose a different implementation, using the capacity of pytest to skip tests on the fly)

docs/edit/manifest.md
```diff
@@ -0,0 +1,136 @@
...
+      }
+    },
+
+    "file_content": {
+      "oneOf": [
+        {
+          "type": "object",
+          "patternProperties": {
+            "Test_.+": {
+              "$comment": "It can be a version number, a skip reason, or an object with weblog variant as keys",
```
marcotc, line 101: Is this comment field for each language team to write, for example, "this test is skipped until PR #123 is released", or for the test file declaration?

docs/edit/manifest.md
```diff
@@ -0,0 +1,136 @@
...
+
+1. Often, those declarations involve only one component
+2. Often, a component is modified by only one developper
+
+The idea is to offer an alternative way to declare those metadata, using one file per component, also known as manifest file. Those files will be inside system tests repo, declaring for a given component some metadata.
+
+Pro :
+
+- Will solve the majority of conflict during test activation PRs
+- each team will have ownership of its manifest file, easing the PR preview process
```
marcotc, line 24: I'm thinking we just avoid renaming tests mostly, and update comments instead to better reflect test changes if needed.

docs/edit/manifest.md
cbeauchesne: in json-schema, `$comment`  means a comment in the schema itself (as JSON does not allow comments).

The way to add a comment will be something like : 

```
Test_Feature: bug (JIRA-666)
```

docs/edit/manifest.md
cbeauchesne: Yes, as folders and test class names are the identifier of features, renaming are rare, so it's why I think it's acceptable.

PR #1347, opened 2023-07-04
User auto instrumentation:  safe mode tests for dotnet

anna-git:
## Description
Activate safe mode user auto instrumentation for the dotnet tracer > 2.32.0

## Motivation
It is now implemented in the tracer

## Reviewer checklist

- all tests should pass
- 2 modifications
    -  "manual.keep" never existed in dotnet tracers and is an old tag. we prefer using `span["metrics"]["_sampling_priority_v1"] == self.MANUAL_KEEP` 
    - usr.id is not added if it's a PII, it is a whitespace only for nodejs


.gitignore
```diff
@@ -6,6 +6,7 @@ binaries/*
 !binaries/README.md
 
 venv
+venv.
```
smola, line 9: `venv.`? Is that what gets created on Windows?

tests/appsec/test_automated_login_events.py
```diff
@@ -36,6 +36,8 @@ class Test_Login_Events:
     BASIC_AUTH_INVALID_USER_HEADER = "Basic aW52YWxpZFVzZXI6MTIzNA=="  # base64(invalidUser:1234)
     BASIC_AUTH_INVALID_PASSWORD_HEADER = "Basic dGVzdDoxMjM0NQ=="  # base64(test:12345)
 
+    MANUAL_KEEP = 2
```
smola, line 39:
```suggestion
    MANUAL_KEEP_SAMPLING_PRIORITY = 2
```

Maybe it's a bit more informative.

tests/appsec/test_automated_login_events.py
```diff
@@ -47,10 +49,14 @@ def test_login_pii_success(self):
             assert r.status_code == 200
             for _, _, span in interfaces.library.get_spans(request=r):
                 meta = span.get("meta", {})
+                if context.library == "nodejs":
+                    assert meta["usr.id"] == " "
```
smola, line 53: Node sets it to a space?
Sounds like a bug rather than something we should hardcode here?
cc @uurien

.gitignore
anna-git: yes it's what I get locally

tests/appsec/test_automated_login_events.py
anna-git: it's related to[ this discussion](https://dd.slack.com/archives/C054NVD6URE/p1688468402933179?thread_ts=1688053198.321579&cid=C054NVD6URE)

tests/appsec/test_automated_login_events.py
```diff
@@ -155,7 +165,14 @@ def test_login_sdk_failure(self):
                 assert meta["appsec.events.users.login.failure.track"] == "true"
                 assert meta["appsec.events.users.login.failure.usr.id"] == "sdkUser"
                 assert meta["appsec.events.users.login.failure.usr.exists"] == "true"
-                assert meta["manual.keep"] == "true"
+                self.assert_priority(span, meta)
+
+    def assert_priority(self, span, meta):
+        assert (
```
cbeauchesne, line 171: for better debugging, I would rewrite this as : 

```python
if span["metrics"]["_sampling_priority_v1"] != self.MANUAL_KEEP:
    assert "manual.keep" in meta, "manual.keep should be in meta when _sampling_priority_v1 is not MANUAL_KEEP"
    assert meta["manual.keep"] == "true", 'meta.manual.keep should be "true" when _sampling_priority_v1 is not MANUAL_KEEP'
```

cbeauchesne (changes requested):

tests/appsec/test_automated_login_events.py
anna-git: changed in https://github.com/DataDog/system-tests/pull/1347/commits/c4d6c8c814f7842d98dddeab2638c8927bed378e

tests/appsec/test_automated_login_events.py
anna-git: fixed in https://github.com/DataDog/system-tests/pull/1347/commits/c4d6c8c814f7842d98dddeab2638c8927bed378e

.gitignore
cbeauchesne: Could you add a comment ? Almost 100% sure that somebody will take it as a typo in the coming month :)

tests/appsec/test_automated_login_events.py
cbeauchesne: Could you also write a small comment explaning why nodejs behave differently ?

cbeauchesne (changes requested):

GustavoCaso (approved):

GustavoCaso:
Great work @anna-git

cbeauchesne:
Well, I missed that point : only default scenario was running, because the label `run-all-scenarios` was missing. In consequence, the scenario is brokken because we missed a detail : https://github.com/DataDog/system-tests-dashboard/actions/runs/5471736971/jobs/9963178571#step:22:55

anna-git:
My bad , let me open a PR to fix

PR #1355, opened 2023-07-05
Increase heartbeat interval check delta

uurien:
## Description

Increase the delta when the test is checking that heartbeat interval is not executed too fast. 

## Motivation

This line added few days ago has become a source of a flakyness in dd-trace-js. If the first request needs a bit more time than second to send the heartbeat, the test will fail. 

## Reviewer checklist

* [ ] CI is green

## Workflow


smola (approved):
:+1: Makes sense, since this assert is not really critical (ideally, we could check average heartbeats over an extended period of time, but that'd be too expensive). Anyway, I think it'd be worth investigating why Node sends heartbeats this way so often.

smola:
By the way, if this is still flaky after this PR, I think we could just move this assert to a separate test that we can ignore for Node.

uurien:
> Anyway, I think it'd be worth investigating why Node sends heartbeats this way so often.

Yes, I have to go deeper in that, but meanwhile I think that it is better if we fix the flaky test ASAP-

cbeauchesne (approved):

PR #1367, opened 2023-07-06
Make find_span in parametric tests stricter

bantonsson:
## Description

Makes the checks on the spans returned from `find_span` stricter.

## Motivation

The `find_*` methods are to loose and can return unexpected Spans/Traces.

Some tests failed intermittently on Java on non helpful asserts, and when debugging this, it was discovered that some traces where split into two chunks, and the `child` span we thought we got back was actually the `parent` span once more. None of the tests seem to expect anything other than a _strict_ match, but the original methods are available as `find_similar*`.

## Reviewer checklist

* [ ] CI is green

## Workflow

### DIFF
--- a/utils/parametric/spec/trace.py
+++ b/utils/parametric/spec/trace.py
@@ -181,7 +181,7 @@ def _span_similarity(s1: Span, s2: Span) -> int:
     return score
 
 
-def find_trace_by_root(traces: List[Trace], span: Span) -> Trace:
+def find_similar_trace_by_root(traces: List[Trace], span: Span) -> Trace:
     """Return the trace from `traces` with root span most similar to `span`."""
     assert len(traces) > 0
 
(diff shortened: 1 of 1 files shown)


utils/parametric/spec/trace.py
```diff
@@ -227,6 +239,29 @@ def find_span_in_traces(traces: List[Trace], span: Span) -> Span:
...
+
+
+def _assert_span_match(span: Span, similar: Span) -> Span:
+    for var in Span.__annotations__:
+        if not var.startswith("__"):
+            val = span.get(var)
+            s_val = similar.get(var)
+            if val != None and val != s_val:
+                # TODO remove this special handling once nodejs sets service in the same way.
+                # Right now, nodejs sets a tag named "service" in meta with the correct value.
```
bantonsson, line 254: There is some special handling for `nodejs` here based on the discussion in `slack` about a difference in behavior

cbeauchesne (approved):

dougqh:
I've always wanted most tests to insensitive to chunking logic.

I really thinking only tests stressing sending should be exposed to chunking logic.  For the other tests, I really wish we assembled the chunks into a logical trace and compared against that.

bantonsson:
I agree that it would be great if we could test against _full_ traces. I’ve changed some tests in another PR to wait for the expected number of spans instead, but that is also a stop-gap solution.

PR #1371, opened 2023-07-06
[dotnet] Activate user extended mode 

anna-git:
## Description

Activate auto instrumentation of user events in extended mode configured for dotnet
Had to adapt the main classes as 
- dotnet login oob doesnt have a login field (framework design, username matters)
- sdk was using a different naming that we cant break (already having customers using it)
- when user is found with wrong password, we report his user id and not the name in usr.id field

## Reviewer checklist

* [ ] CI is green

## Workflow

### DIFF
--- a/tests/appsec/test_automated_login_events.py
+++ b/tests/appsec/test_automated_login_events.py
@@ -248,12 +248,12 @@ class Test_Login_Events_Extended:
                     assert meta["appsec.events.users.login.failure.usr.exists"] == "true"
                     assert meta["appsec.events.users.login.failure.usr.id"] == "social-security-id"
                     assert meta["appsec.events.users.login.failure.email"] == "testuser@ddog.com"
+                    assert meta["appsec.events.users.login.failure.username"] == "test"
                 else:
                     assert meta["appsec.events.users.login.failure.usr.id"] == "test"
 
                 assert meta["_dd.appsec.events.users.login.failure.auto.mode"] == "extended"
                 assert meta["appsec.events.users.login.failure.track"] == "true"
-                assert meta["appsec.events.users.login.failure.username"] == "test"
 
                 assert_priority(span, meta)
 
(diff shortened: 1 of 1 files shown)


tests/appsec/test_automated_login_events.py
```diff
@@ -246,11 +246,16 @@ def test_login_wrong_password_failure(self):
                     # Currently in nodejs there is no way to check if the user exists upon authentication failure so
                     # this assertion is disabled for this library.
                     assert meta["appsec.events.users.login.failure.usr.exists"] == "true"
+                    assert meta["appsec.events.users.login.failure.usr.id"] == "social-security-id"
+                else:
+                    assert meta["appsec.events.users.login.failure.usr.id"] == "test"
```
anna-git, line 251: not sure why in nodejs here, we dont retrieve the real user id, but the username? @GustavoCaso

cbeauchesne (approved):

tests/appsec/test_automated_login_events.py
```diff
@@ -246,11 +246,16 @@ def test_login_wrong_password_failure(self):
                     # Currently in nodejs there is no way to check if the user exists upon authentication failure so
                     # this assertion is disabled for this library.
                     assert meta["appsec.events.users.login.failure.usr.exists"] == "true"
+                    assert meta["appsec.events.users.login.failure.usr.id"] == "social-security-id"
+                    assert meta["appsec.events.users.login.failure.email"] == "testuser@ddog.com"
+                    assert meta["appsec.events.users.login.failure.username"] == "test"
+                else:
+                    assert meta["appsec.events.users.login.failure.usr.id"] == "test"
```
anna-git, line 253: not sure why in nodejs here, we dont retrieve the real user id, but the username? @hoolioh

hoolioh: That's a peculiarity of the implementation because we the user info is not available in the implementation despite it exists so we're reporting the login as the id.

hoolioh (approved):

tests/appsec/test_automated_login_events.py
anna-git: as discussed: peculiarity from the SDK implementation where the examples for user login failure were reporting the login/username as 'usr.id' keeping the behavior as is in auto instrumentation. Hence, upon failure,  the 'login' is used and turned to an 'id' field.

PR #1372, opened 2023-07-06
Update rabbit tests in test_dsm.py for both Java and dotnet

hokitam:
## Reviewer checklist

* [ ] CI is green

## Workflow

### DIFF
--- a/tests/integrations/test_dsm.py
+++ b/tests/integrations/test_dsm.py
@@ -66,6 +66,43 @@ class Test_DsmRabbitmq:
             tags=("direction:out", "exchange:systemTestDirectExchange", "has_routing_key:true", "type:rabbitmq"),
         )
 
+        DsmHelper.assert_checkpoint_presence(
+            hash_=1648106384315938543,
+            parent_hash=6176024609184775446,
+            tags=('direction:in', 'topic:systemTestRabbitmqQueue', 'type:rabbitmq'),
+        )
+
+
+@released(cpp="?", golang="?", nodejs="?", php="?", python="?", ruby="?", java="?")
+@released(dotnet="2.29.0")
+@scenarios.integrations
+class Test_DsmRabbitmq_Dotnet:
+    """ Verify DSM stats points for RabbitMQ, specifically for Dotnet """
+
+    def setup_dsm_rabbitmq(self):
+        self.r = weblog.get("/dsm?integration=rabbitmq")
+
+    def test_dsm_rabbitmq(self):
+        assert self.r.text == "ok"
+
+        # Dotnet sets the tag for `has_routing_key` to `has_routing_key:True` instead of `has_routing_key:true` like
+        # the other tracer libraries, which causes the resulting hash to be different.
+        DsmHelper.assert_checkpoint_presence(
+            hash_=12547013883960139159,
+            parent_hash=0,
+            tags=("direction:out", "exchange:systemTestDirectExchange", "has_routing_key:True", "type:rabbitmq"),
+        )
+
+        # There seems to be a bug in dotnet currently where the queue is not passed, causing DSM to default to setting
+        # the routing key as the topic.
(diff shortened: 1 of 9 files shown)


tests/integrations/test_dsm.py
```diff
@@ -66,6 +66,43 @@ def test_dsm_rabbitmq(self):
...
+            hash_=1648106384315938543,
+            parent_hash=6176024609184775446,
+            tags=('direction:in', 'topic:systemTestRabbitmqQueue', 'type:rabbitmq'),
+        )
+
+
+@released(cpp="?", golang="?", nodejs="?", php="?", python="?", ruby="?", java="?")
+@released(dotnet="2.29.0")
+@scenarios.integrations
+class Test_DsmRabbitmq_Dotnet:
```
smola, line 79: Rather than defining a separate class, which will make things harder to compare in feature matrices, it might be better to just add some `if/else` within the main class for the behavior that is different in dotnet.

hokitam: the dotnet behavior is actually a bug. I'm hoping to get it fixed maybe in a week or 2. Is there a way to keep this, but Is there a way to "limit" the dotnet version of the test so that the test won't run for the later releases?

I prefer keeping the test showing the bug because the bug is actually not super obvious for some cases and hence might end up being considered "correct behavior" in some instances. but lmk if you think it's better to just not even have this test with the incorrect behavior. I prefer keeping this separate because it'll be very easy to re-activate the good test for dotnet later without needing to make a big change to the test itself.

wdyt?

cbeauchesne: @smola is right, classes should be tracer-agnostic, and just describe a feature.

You can do something like this : 

```
@released(dotnet="x,y", java="w.z")
class Test_feature:
   @bug(library="dotnet", reason="jira-ticket")
   def test_main(self):
      ...

   @irrelevant(context.library != "dotnet", reason="jira-ticket")
   def test_main_dotnet_legacy(self):
        """ dotnet behavior is a bug, but we want to keep the test, waiting for a fix """
        ....
```

cbeauchesne (changes requested):

tests/integrations/test_dsm.py
smola: IMHO it could look something like this:

```
@released(cpp="?", dotnet="<some dotnet version here>", golang="?", nodejs="?", php="?", python="?", ruby="?")
@released(java={"spring-boot": "1.13.0", "*": "?"})
@scenarios.integrations
class Test_DsmRabbitmq:
    """ Verify DSM stats points for RabbitMQ """
    def setup_dsm_rabbitmq(self):
        self.r = weblog.get("/dsm?integration=rabbitmq")
    
    @bug(library="dotnet", reason="Different behaviour and so on")
    def test_dsm_rabbitmq(self):
        # regular test

    @irrelevant(context.library.library != "dotnet", reason="dotnet-specific")
    def test_dsm_rabbitmq_dotnet(self):
        # dotnet variant
```

smola: @cbeauchesne Good timing :D

tests/integrations/test_dsm.py
```diff
@@ -72,16 +74,7 @@ def test_dsm_rabbitmq(self):
...
-@released(dotnet="2.29.0")
-@scenarios.integrations
-class Test_DsmRabbitmq_Dotnet:
-    """ Verify DSM stats points for RabbitMQ, specifically for Dotnet """
-
-    def setup_dsm_rabbitmq(self):
-        self.r = weblog.get("/dsm?integration=rabbitmq")
-
+    @irrelevant(context.library != "dotnet", reason="legacy dotnet behavior")
     def test_dsm_rabbitmq(self):
```
cbeauchesne, line 78: you have two `test_dsm_rabbitmq` test methods :) 

You can name the second one `test_dsm_rabbitmq_dotnet_legacy`.

cbeauchesne (changes requested):

tests/integrations/test_dsm.py
hokitam: good catch! fixed, thanks!

cbeauchesne (approved):

PR #1386, opened 2023-07-10
Parametric tests now only builds Docker image once per session

bantonsson:
## Description

Parametric tests now only builds the Docker image once per `pytest` session.

## Motivation

When running `pytest` concurrently the parametric tests kept overwriting the `Dockerfile` for every `request` leading to races where the file itself was sometimes empty or broken when the `build` command was executed. This lead to spurious failures.

## Workflow

## Reviewer checklist

* [ ] CI is green


tests/parametric/conftest.py
```diff
@@ -413,13 +411,77 @@ def get_open_port():
...
 
 
-@pytest.fixture
-def apm_test_server(request, library_env, test_id):
-    # Have to do this funky request.param stuff as this is the recommended way to do parametrized fixtures
-    # in pytest.
-    apm_test_library = _libs[context.scenario.library.library]
+@pytest.fixture(scope="session")
+def session_log_file():
+    log_path = f"{context.scenario.host_log_folder}/outputs/session_log.log"
```
robertomonteromiguel, line 416: I have not been able to understand 100% of your changes, but I have been able to understand the intent of the PR.
I did a little test with the session_log file. You are generating a single file, but if we put a random file name, we can see that several files are generated (one per worker I guess). Now I wonder, this PR would eliminate 100% the collisions when overwriting the dockerfile or just reduce the probabilities?

` log_path = f"{context.scenario.host_log_folder}/outputs/session_log_{random.randint(0, 888889)}.log"`

robertomonteromiguel (changes requested):

tests/parametric/conftest.py
bantonsson: I have now changed this

robertomonteromiguel: Now, we are generating one log per xdist worker, but it is only being written in one.
I can't see where the problem is in the code.

robertomonteromiguel (changes requested):
Now, we are generating one log per xdist worker, but it is only being written in one.
I can't see where the problem is in the code.

tests/parametric/conftest.py
bantonsson: Yes, that log file is only ever used during the building of the Docker image, and should only be written to by the worker actually building the docker image. Do you want me to change the name of the log files, or do you want there to be only one log file that the worker that does the building creates instead?

robertomonteromiguel: I think that if we only write to one file, there should be only one.
Thanks @bantonsson ! I think this PR is a good improvement!

bantonsson: Thanks, I'll fix that then.

PR #1387, opened 2023-07-10
Automated user login events for ruby. 

GustavoCaso:
## Description

Add all the necessary work for enabling automated login events for Ruby. The code is on master, but we haven't yet released it. Still, this PR makes all the necessary work upfront to make sure once we release dd-trace-rb, we can enable the tests by changing the release version.

## Workflow

## Reviewer checklist

* [ ] CI is green

### DIFF
--- a/tests/appsec/test_automated_login_events.py
+++ b/tests/appsec/test_automated_login_events.py
@@ -8,7 +8,10 @@ from utils import weblog, interfaces, context, missing_feature, released, scenar
 @rfc("https://docs.google.com/document/d/1-trUpphvyZY7k5ldjhW-MgqWl0xOm7AMEQDJEAZ63_Q/edit#heading=h.8d3o7vtyu1y1")
 @coverage.good
 @released(cpp="?", golang="?", java="?", nodejs="4.4.0", dotnet="2.32.0", php="?", python="?", ruby="?")
-@missing_feature(weblog_variant="rails32", reason="Not able to configure weblog variant properly. Issue with SQLite and PRIMARY_KEY as String and Rails 3 protected attributes")
+@missing_feature(
+    weblog_variant="rails32",
+    reason="Not able to configure weblog variant properly. Issue with SQLite and PRIMARY_KEY as String and Rails 3 protected attributes",
+)
 @missing_feature(weblog_variant="rack", reason="We do not support authentication framework for rack")
 @missing_feature(weblog_variant="sinatra12", reason="We do not support authentication framework for sinatra")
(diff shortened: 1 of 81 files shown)


utils/build/docker/ruby/rails40.Dockerfile
```diff
@@ -12,6 +12,6 @@ RUN /binaries/install_ddtrace.sh
 
 ENV DD_TRACE_HEADER_TAGS=user-agent
 
-RUN echo "#!/bin/bash\nbundle exec thin start -p 7777" > app.sh
+RUN echo "#!/bin/bash\nbundle exec rake db:create db:migrate db:seed\nbundle exec thin start -p 7777" > app.sh
```
cbeauchesne, line 15: Is there a reason for Having `rake db:create db:migrate db:seed` at runtime rather during the build ?

tests/appsec/test_automated_login_events.py
```diff
@@ -37,10 +45,16 @@ class Test_Login_Events:
     BASIC_AUTH_INVALID_PASSWORD_HEADER = "Basic dGVzdDoxMjM0NQ=="  # base64(test:12345)
 
     def setup_login_pii_success(self):
-        self.r_pii_success = [
-            weblog.post("/login?auth=local", data={"username": self.USER, "password": self.PASSWORD}),
-            weblog.get("/login?auth=basic", headers={"Authorization": self.BASIC_AUTH_USER_HEADER}),
-        ]
+        if context.library == "ruby":
```
cbeauchesne, line 48: Could you add methods to the class: 

```python
class Test_Login_Events:
    """ Test login success/failure use cases """

    @property
    def username_key(self):
       """ <Explanation why is it different> """
       return "user[username]" if "rails" in context.weblog_variant else "username"

    # same for password_key
```

Then, setup method will be all the same : 

```python
                weblog.post("/login?auth=local", data={self.username_key: self.USER, self.password_key: self.PASSWORD}),
```

cbeauchesne (changes requested):

hoolioh (approved):

lloeki (approved):
I don't see anything wild in there. LGTM.

We should find a way to make most of this weblog code shareable between Rails versions. I did it with Rails engines back in the day but engines create a bit of special cases (e.g they're nested rack apps) that would bias the tests, so I'm more thinking about e.g some dumb Ruby defining modules that get included in or extend controllers, and for migrations eschew activerecord in migrations and seeds and just run things that create the appropriate database and data (e.g restore a SQL dump or use something like https://github.com/lloeki/rebel)

PR #1399, opened 2023-07-12
cpp: pass more parametric tests

dgoffredo:
## Description
Recent changes to [dd-trace-cpp][1] have made it more compatible with some of the parametric tests in this repository. Specifically, tests involving:

- 128-bit trace IDs
- trace propagation, with the exception of the "B3 single header" style.

This revision removes "cpp" exemptions from many tests, which now pass. Some of the tests for which "cpp" is still exempt now have updated descriptions. For example, it is still an open question (at least until Zach Montoya returns) how `_dd.p.dm` propagation should be handled in certain cases. Also, whether dd-trace-cpp should support the B3 single header propagation style, which has not been requested by any users of the C++ tracer and which is largely superseded by the similar W3C propagation style. Those questions we can address later.

See the commits for a breakdown of all that is proposed here.

## Reviewer checklist

* [ ] CI is green
 
[1]: https://github.com/DataDog/dd-trace-cpp

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -1,7 +1,7 @@
 Tests scenario.
 
 * Each file tests a feature
-* Each class tests a feature declinaison
+* Each class tests a feature declension
 * Each method tests a use case
 
 Please use meaningful messages, and avoid any duplication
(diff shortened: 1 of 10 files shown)


dgoffredo:
The 5 CI failures appear to be spurious.

PR #1424, opened 2023-07-20
Fix pycache clearing

lloeki:
## Description

Fix pycache clearing.

## Motivation

`-prune` appears to be useless, unless (or even if) someone expects a `__pycache__` folder to be nested.

@cbeauchesne reported that this occcured:

```
rm: cannot remove 'tests/parametric/__pycache__': Directory not empty
```

Possible causes:

- The `find` command attempted to delete things before the directory was entirely cleared out (but there's `-rf` on `rm`)
- Something regenerated pycache files while it was being recursively removed
- Something created nested files that could not be removed (e.g due to permissions):
  - A typical candidate would be a container with that path exposed as a volume, but the users within (root) and outside (regular user) are in mismatch, so the generated files and folders are not writable, thus not removable
  - `ProxyContainer` does that but exposes `ro` so probably not that. Also, it's happening in `tests/parametric`, and no container has that exposed as a volume by default.

## Workflow

## Reviewer checklist

* [ ] CI is green

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -102,7 +102,7 @@ function downcase() {
 }
 
 function clean_pycache() {
-    find utils tests -type d -name '__pycache__'  -prune -exec rm -rf {} +
+    find utils tests -type d -name '__pycache__' -exec rm -rf {} +
 }
 
 function is_using_nix() {
--- a/run.sh
+++ b/run.sh
@@ -101,10 +101,6 @@ function downcase() {
     tr '[:upper:]' '[:lower:]'
 }
 
-function clean_pycache() {
-    find utils tests -type d -name '__pycache__' -exec rm -rf {} +
-}
-
 function is_using_nix() {
     [[ -n "${IN_NIX_SHELL:-}" ]]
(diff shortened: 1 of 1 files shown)


cbeauchesne (approved):

lloeki:
That said, I'm wondering if clearing pytest is of any use.

The only case where it's useful is if interpreters were mismatched with `.pyc` files, but that would only happen:

- for `tests`, if someone switches to another python locally.
- for `utils`, if the tree is exposed via a volume and there's a mismatch between the local python and the container's python, which would only be the case for the proxy container.

But then:

> Since Python 3.2 all the cached files prefixed with interpreter version as `mymodule.cpython-32.pyc`

So it seems to serve no purpose.

cbeauchesne:
I would give a try to the option "no clear" :)

lloeki:
Done.

In your specific case my hunch is that there was some other process running that imports some of the files (like an IDE that runs python directly or a LSP server)

PR #1430, opened 2023-07-21
Prevent finding rid in non-service entry spans

marcotc:
## Description

We are introducing proper `DD_TRACE_HEADER_TAGS` support in `dd-trace-rb` (https://github.com/DataDog/dd-trace-rb/pull/2935) and it trips a few appsec system-tests.

The issue happens because we tag non-service entry spans with `DD_TRACE_HEADER_TAGS`, but those tags are used to find the request id (`rid`) by system-tests to perform assertions.

It is likely very uncommon to set HTTP serve `DD_TRACE_HEADER_TAGS` tags in non-service entry spans, but it's the current state of the Ruby tracer. We should likely change this, and stop tagging Sinatra spans with `DD_TRACE_HEADER_TAGS`, but that's a breaking change for our clients and won't happen in the short term.
Disabling test until this change happens in the Ruby tracer is also not a great approach, because that would mean disabling a large set of appsec tests that we want to keep.

## Workflow

## Reviewer checklist

* [ ] CI is green

### DIFF
--- a/utils/interfaces/_core.py
+++ b/utils/interfaces/_core.py
@@ -170,6 +170,7 @@ def get_rid_from_span(span):
         return None
 
     meta = span.get("meta", {})
+    metrics = span.get("metrics", {})
 
     user_agent = None
 
@@ -177,7 +178,7 @@ def get_rid_from_span(span):
         user_agent = meta.get("grpc.metadata.user-agent")
         # java does not fill this tag; it uses the normal http tags
 
-    if not user_agent:
+    if not user_agent and metrics.get("_dd.top_level") == 1.0:
         # code version
         user_agent = meta.get("http.request.headers.user-agent")
 
--- a/utils/interfaces/_core.py
+++ b/utils/interfaces/_core.py
@@ -179,7 +179,7 @@ def get_rid_from_span(span):
         # java does not fill this tag; it uses the normal http tags
 
     if not user_agent and metrics.get("_dd.top_level") == 1.0:
-        # code version
+        # The top level span (aka root span) is mark via the _dd.top_level tag by the tracers
         user_agent = meta.get("http.request.headers.user-agent")
 
     if not user_agent:  # try something for .NET


marcotc:
CI failures seem to mirror the master branch: https://github.com/DataDog/system-tests/actions/runs/5624028552

GustavoCaso (approved):
Love this 🎉 

Since it warranty that we are fetching the service level entry span when checking for `_dd.top_level` 👍 

Question: Should we apply the same check for the other conditions?

marcotc:
> Question: Should we apply the same check for the other conditions?

I'm not confident it should, because the other conditions are documented as language-specific, and the for the `# last hope` check, we are likely looking to find anything useful at that point so having a wider-matching condition is probably ok.

utils/interfaces/_core.py
```diff
@@ -170,14 +170,15 @@ def get_rid_from_span(span):
...
+    metrics = span.get("metrics", {})
 
     user_agent = None
 
     if span.get("type") == "rpc":
         user_agent = meta.get("grpc.metadata.user-agent")
         # java does not fill this tag; it uses the normal http tags
 
-    if not user_agent:
+    if not user_agent and metrics.get("_dd.top_level") == 1.0:
```
pawelchcki, line 181: It would be good to add comment explaining the new check

pawelchcki (changes requested):
Tiny nit otherwise looks good to me

pawelchcki (approved):

utils/interfaces/_core.py
lloeki: This kind of works, but it feels indirect:

- there can be multiple top level tags in the trace (e.g with request queuing synthetic spans, which are in fact at the root of the trace). we don't run in this way but we should have weblog cases for that situation.
- if this is squarely looking for the appsec case then `_dd.appsec.enabled =1.0` is a better, unambiguous marker
- it also begs into question why appsec and tracing duplicate tagging in different places

That said, I'd rather have the CI green again, so OK to merge and fix later while we think about it (cc @marcotc)

lloeki:
Rerunning CI after the recent merge on dd-trace-rb master

lloeki:
Ruby is all green, the other failures are flakiness

PR #1432, opened 2023-07-25
remove log level to debug in flask application

christophe-papazian:
## Description

Remove default log level to DEBUG in flask application

## Motivation

Since merge of https://github.com/DataDog/system-tests/pull/1426 some tests using python tracer with flask-poc were failing non deterministically. It seems that default log level to debug can generate too much data and prevent some requests to have their responses caught by the tests.

## Reviewer checklist

* [ ] CI is green

### DIFF
--- a/utils/build/docker/python/flask/app.py
+++ b/utils/build/docker/python/flask/app.py
@@ -1,5 +1,3 @@
-import logging
-
 import psycopg2
 import requests
 from ddtrace import tracer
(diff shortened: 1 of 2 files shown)


utils/build/docker/python/flask/app.py
```diff
@@ -24,8 +22,6 @@
     host="postgres", port="5433", user="system_tests_user", password="system_tests", dbname="system_tests",
 )
 
-logging.basicConfig(level=logging.DEBUG)
```
brettlangdon, line 27: do we still want logging, just the default warning level or something?

brettlangdon (approved):

utils/build/docker/python/flask/app.py
christophe-papazian: Default is error. For automatic tests on github action, we probably prefer to be sure that there is no flaky tests at all. But on local testing, this could be added back.

PR #1435, opened 2023-07-25
Fix yaml parsing

lloeki:
## Description

In contexts where we run, `perl` is available with `YAML` module. Use that instead of `python` to read `yml`.

This PR also introduce:

- a verbosity control flag
- a dry run flag

Both being very useful for debugging the `run.sh` script.

## Motivation

The need for `activate_venv` in #1427 is caused by `pyyaml` being needed byt not being available as a default.

It's a bit early to activate the virtualenv in some scenarios, and in others, activating the venv is potentially harmful (e.g when there is no need for a venv).

## Workflow

## Reviewer checklist

* [ ] CI is green

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,10 @@ jobs:
       # force /bin/bash in order to test against bash 3.2 on macOS
       - name: Test the test (direct)
         run: /bin/bash run.sh TEST_THE_TEST
+      - name: Test the test (docker)
+        run: |
+          /bin/bash build.sh -i runner -d
+          /bin/bash run.sh +d TEST_THE_TEST
       - name: Test group parsing
         run: |
           /bin/bash run.sh ++dry APPSEC_SCENARIOS
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,12 +46,13 @@ jobs:
         uses: actions/checkout@v3
       - name: Install runner
         uses: ./.github/actions/install_runner
+      # force /bin/bash in order to test against bash 3.2 on macOS
       - name: Test the test (direct)
-        run: ./run.sh TEST_THE_TEST
+        run: /bin/bash run.sh TEST_THE_TEST
       - name: Test group parsing
         run: |
-          ./run.sh ++dry APPSEC_SCENARIOS
-          ./run.sh ++dry TRACER_RELEASE_SCENARIOS
+          /bin/bash run.sh ++dry APPSEC_SCENARIOS
+          /bin/bash run.sh ++dry TRACER_RELEASE_SCENARIOS
 
   scenarios:
     uses: datadog/system-tests/.github/workflows/compute-scenarios.yml@main
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,10 +49,6 @@ jobs:
       # force /bin/bash in order to test against bash 3.2 on macOS
       - name: Test the test (direct)
         run: /bin/bash run.sh TEST_THE_TEST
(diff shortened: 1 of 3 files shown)


run.sh
```diff
@@ -90,8 +97,41 @@ function die() {
 function lookup_scenario_group() {
     local group="$1"
 
-    activate_venv
-    cat < scenario_groups.yml | python -c 'import yaml; import sys; key = sys.argv[1]; data = sys.stdin.read(); g = yaml.safe_load(data)[key]; [[print(t) for t in s] if isinstance(s, list) else print(s) for s in g]' "${group}"
+    # formatted perl code with -e for readabiliy
+    # shellcheck disable=SC2016
+    cat < scenario_groups.yml \
+        | env GROUP="${group}" perl -0777 \
```
lloeki, line 103: > @pawelchcki do we know if this perl setup will be available for all users of system tests? Linux, and OSX laptops + CIs ?

- macOS: yes (perl + YAML module)
- Linux: git depends on perl, but for YAML depends on whether a system dependent need the YAML module
- CI: git depends on perl, but YAML module is absent

lloeki: > @pawelchcki: I was wondering about it thought - why not use Docker itself for these kind of things ? Since we expect Docker to work for everyone wanting to use system-tests. Its our lingua franca - of sorts.
You could embed the conversion script as small docker image. And just call cat yaml | docker run --rm system-tests/scripts process.yaml > output|

I tried this:

```
cat < scenario_groups.yml | docker run --rm alpine:3.16 /bin/sh -c "apk add --quiet --no-progress python3 py3-yaml && python3 -c 'import yaml; import sys; key = sys.argv[1]; data = sys.stdin.read(); g = yaml.safe_load(data)[key]; print(repr(g)); [[print(t) for t in s] if isinstance(s, list) else print(s) for s in g]' 'ONBOARDING_SCENARIOS'"
```

But it does not work:

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
TypeError: 'NoneType' object is not subscriptable
```

I feared as much, as piping into (or out of) docker is a strange affair (there's a whole lot of whacky forwarding involved because containers are run under `dockerd`, not as children of `docker run`)

Anyway I don't think docker is available on macOS in CI.

lloeki: I was thinking of reinstating `activate_venv` (although properly conditioned behind the nix shell check this time) is problematic because when using `./build.sh -d -i runner` + `./run.sh +d` at no point the python dependencies are needed locally, so they may be entirely absent.

That said, the runner image would have these dependencies, so based on `+d` it could either locally `activate_venv` as needed _or_ run in docker (if I manage to achieve that)

lloeki: Alright, I rolled back to using python and moved the conditioned `activate_venv` part earlier so that it's available for the whole `run.sh` script.

run.sh
```diff
@@ -185,7 +221,8 @@ function main() {
                   exit 64
                 fi
                 # upcase via ${2^^} is unsupported on bash 3.x
-                mapfile -t group <<< "$(lookup_scenario_group "$(echo "$2" | upcase)")"
+                # bash 3.x does not support mapfile, dance around with tr and IFS
```
pawelchcki, line 224: :crying_cat_face:

.github/workflows/ci.yml
```diff
@@ -34,13 +34,25 @@ jobs:
       uses: ./.github/actions/lint_code
 
   test_the_test:
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os:
+          - macos-latest
```
pawelchcki, line 40: :bow:

PR #1449, opened 2023-07-31
Revert "Nodejs - telemetry heartbeat are sent too fast"

simon-id:
Reverts DataDog/system-tests#1419

Fixed in: https://github.com/DataDog/dd-trace-js/pull/3553


christophe-papazian (approved):

smola (approved):
@simon-id You might want to keep the annotation with a conditional for the version where it's fixed.

PR #1471, opened 2023-08-08
Support dotnet arm64

shurivich:
## Description
Fix for the bug when `dotnet` library compiled for x64 used on arm64 platform.


cbeauchesne:
Well, as it's onlu for the build step, it's ok to run only the default scenario

cbeauchesne (approved):

PR #1476, opened 2023-08-08
Add new test_sample_rate_function

LotharSee:
## Description

Add a new test `test_sample_rate_function` which checks that weblog takes the exact same sampling decision as the one precomputed in an auxiliary .csv file.

This test is broader and stricter than the previous tests `test_sampling_decision`, `test_sampling_decision_added` and `test_sampling_decision_added` (if any of them fail, the new one would fail too). In a later PR, I'll propose to consolidate it all.

## Motivation

The purpose of this approach is to use a portable language agnostic set of `(trace_id, sampling_rate, sampling_decision)` to test the implementation of the sampling function. The given CSV can be copied to be sure in libraries' unit tests. I plan on using it on to test our backend implementation too.
The CSV is meant to include cases for "tricky" values (ex: small, large, at the edge of 2^63) which would who typically fail if the function were badly implemented (ex: offset by one, doing a 2^63 modulo instead of 2^64).

## Workflow

## Reviewer checklist

* [x] CI is green
* [x] if any of `build-some-image` label is present


tests/test_sampling_rates.py
```diff
@@ -254,3 +260,60 @@ def validator(data):
...
+                cases.append((int(row[0]), float(row[1]), int(row[2])))
+        return cases
+
+    def setup_sample_rate_function(self):
+        self.requests_expected_decision = []
+
+        for trace_id, sampling_rate, sampling_priority in self._load_csv_sampling_decisions():
+            sampling_decision = priority_should_be_kept(sampling_priority)
+            # Check that the input .csv is valid according to this test's implementation of trace_should_be_kept.
+            assert sampling_decision is trace_should_be_kept(sampling_rate, trace_id)
```
cbeauchesne, line 284: Setup function must never fail, any assertion should be made on test method.

IIUC, it's only csv data validation. In that case, I think it would be better to move it in `_load_csv_sampling_decisions`? WDYT ?

cbeauchesne: Yes it's ok. If ever it fails, you will have a big nasty SYSTEM TESTS INTERNAL ERROR that does not look similar to a regular failure in stdout.

cbeauchesne (approved):

PR #1485, opened 2023-08-11
Testing All DBM Possible Setting Tag Behavior 

link04:
## Description
Adds to the DBM<>APM link tests to check the outcome of the exact span tags when the feature is enabled and disabled, this will help with other tests I plan on adding to confirm the comment added was as expected on the query depending of the setting.

## Motivation
Based on the DBM<>APM link RFC the spans should be tagged for a specific setting such `full` but not for `disabled` or `service`, in my implementation I have a bug since I am adding the tag for both none `disabled` but this didn't get covered on the system test I wrote because it didn't take into account different settings. 

Now I check directly that the test span is tagged or not and will check in a different PR if the propagated commented query ran on the KB matches the expected result based on the span details and setting for the DBM propagation.

## Workflow

## Reviewer checklist

* [ ] CI is green


tests/integrations/test_dbm.py
```diff
@@ -24,17 +23,68 @@ def setup_trace_payload(self):
...
+            for _, trace in interfaces.library.get_traces(request=r):
+                db_span = None
+                for span in trace:
+                    if span.get("resource") == "SELECT version()" or span.get("resource") == "SELECT @@version":
+                        db_span = span
+                        break
+
+                assert (
+                    db_span is not None
+                ), "No DB span with expected resource 'SELECT version()' nor 'SELECT @@version' found."
```
andrewlock, line 54: nit: you could extract this to a helper function that gets the `db_span`, as it's duplicated between your three tests?

tests/integrations/test_dbm.py
```diff
@@ -24,17 +23,68 @@ def setup_trace_payload(self):
...
+                        db_span = span
+                        break
 
-                meta = span.get("meta", {})
-                assert "_dd.dbm_trace_injected" in meta
+                assert (
+                    db_span is not None
+                ), "No DB span with expected resource 'SELECT version()' nor 'SELECT @@version' found."
+                meta = db_span.get("meta", {})
+                assert "_dd.dbm_trace_injected" in meta, "_dd.dbm_trace_injected not found in span meta."
```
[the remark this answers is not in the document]
link04: Added all the changes on the https://github.com/DataDog/system-tests/pull/1485/commits/98e1c7c2d62c0e2844617e34c22e5193c24e360e commit , thank you!

tests/integrations/test_dbm.py
```diff
@@ -31,12 +31,56 @@ def setup_trace_payload(self):
...
-    def test_trace_payload(self):
+    def find_db_spans(self):
+        self.db_span = None
         for r in self.requests:
-            assert r.status_code == 200, f"{r.request.url} is not successful"
-            for _, _, span in interfaces.library.get_spans(request=r):
-                if span.get("span_type") != "sql":
-                    return
+            assert r.status_code == 200, f"Request: {r.request.url} wasn't successful."
+            for _, trace in interfaces.library.get_traces(request=r):
```
cbeauchesne, line 38: `get_spans` can replace the nested for

```suggestion
            for _, _, span in interfaces.library.get_spans(request=r):
```

tests/integrations/test_dbm.py
```diff
@@ -31,12 +31,56 @@ def setup_trace_payload(self):
...
+                for span in trace:
+                    if span.get("resource") == "SELECT version()" or span.get("resource") == "SELECT @@version":
+                        self.db_span = span
+                        break
```
cbeauchesne, line 42: If resource does not match, it'll be hard to debug : 


```suggestion
                        break
                    else:
                        logger.debug(f"Found a span, but ressource doesn't match: {span.get('resource')}"    
```

utils/_context/_scenarios.py
```diff
@@ -946,12 +946,16 @@ class scenarios:
 
     default = EndToEndScenario(
         "DEFAULT",
+        weblog_env={"DD_DBM_PROPAGATION_MODE": "disabled"},
         include_postgres_db=True,
-        doc="Default scenario, spwan tracer and agent, and run most of exisiting tests",
+        include_mysql_db=True,
```
cbeauchesne, line 951: Adding DB in a scenario slows it a lot, and we need to keep the default scenario as fast as possible. Let me one day to think about the best solution here.

cbeauchesne (changes requested):
Few requests changes, and also, CI is failing ;)

tests/integrations/test_dbm.py
```diff
@@ -22,7 +22,7 @@ def setup_trace_payload(self):
             self.requests = [
                 weblog.get("/dbm", params={"integration": "npgsql"}, timeout=20),
                 weblog.get("/dbm", params={"integration": "mysql"}),
-                weblog.get("/dbm", params={"integration": "sqlclient"}),
+                weblog.get("/dbm", params={"integration": "sqlclient"}, timeout=20),
```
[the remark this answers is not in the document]
link04: Was able to run it locally without it, so will remove it.

tests/integrations/test_dbm.py
```diff
@@ -31,12 +31,56 @@ def setup_trace_payload(self):
                 weblog.get("/dbm", params={"integration": "mysqli"}),
             ]
 
-    def test_trace_payload(self):
+    def find_db_spans(self):
```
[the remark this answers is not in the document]
link04: Hey @cbeauchesne, so in this case Im not sure how to apply said changes, I've tried for a couple of hours but the outcome is that instead of being able to find my spans for each request like in the screenshot below:
image

I end up breaking the test and having it only go trough one request, this is also part of why I stopped using `get_spans` because I needed to iterate trough each span to find the right one, before the test was flaky because it would match against any span and not check the tags so not sure how to accomplish that based on your latest recommendations:
image

The above is how the test logs end up looking like so if you have some time for us to sync on this tomorrow that would be great!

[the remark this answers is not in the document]
cbeauchesne: I've sent you an invite

cbeauchesne:
Hi @link04 ,

So about the scenario stuff. 

1. we want to have as many feature as possible in DEFAULT scenario, as it's the most tested
2. but we don't want to  add DB container inside it, because we want to keep it fast
3. and we don't want to create scenario if possible

So here is what I propose : 

1. we keep the scenario INTEGRATION with the full set of DB, with mode `full`
2. we use the mode `service` for the default scenario, and we test only postgres
3. we use the scenario `APPSEC_DISABLED` (yes, it should be called `DISABLED`) to test the mode `disabled`, and we add only one DB inside this scenario

 Pro : 

* no extra time on DEFAULT
* mode `service` battle tested
* no extra scenario

Con : 

* mode `service` and `disabled` are  tested  on 1 DB only
* mode `full` is not on the default scenario, (but having all DB on it is not a choice, so I guess we prefer to test it on a full set of DB)

If you want to test `disabled` and `service` on a bigger set of DB, we'll need another scenario. But I think we can start with that, and see later if it's needed or not. 

WDYT ?

utils/_context/_scenarios.py
link04: Thank you for the details below @cbeauchesne! Looking at how Postgres is the main DB that is being tested and service mode is the most used for .NET I have no objections: 
image

Will work on implementing these changes unless you have any other comments.

cbeauchesne: Ok for me !

tests/integrations/test_dbm.py
```diff
@@ -21,22 +28,87 @@ def setup_trace_payload(self):
...
+    setup_trace_payload_disabled = weblog_trace_payload
+
+    # Test Methods
+    @scenarios.appsec_disabled
+    def test_trace_payload_disabled(self):
+        self._assert_spans_are_untagged()
+
+    setup_trace_payload_service = weblog_trace_payload
+
+    @scenarios.default
```
cbeauchesne, line 100: For later PR, you can ommit `@scenarios.default`: If no scenario is specified, then it's the default one.

cbeauchesne (approved):

PR #1586, opened 2023-09-12
[Debugger] Add line probe snapshot tests 

shurivich:
## Description
Add line probe snapshot tests


tests/debugger/test_debugger.py
```diff
@@ -123,6 +123,30 @@ def test_line_probe_status(self):
         validate_data(expected_data, [])
 
 
+def wait_for_remote_config(self, data):
```
cbeauchesne, line 126: Could you use a base class for this, it will be a little bit cleaner : 

```python
class _Base_Debugger_Test:
    remote_config_is_sent = False
    probe_installed = False
   def wait_for_remote_config(self, data):
        ...

...

class Test_Debugger_Method_Probe_Snaphots(_Base_Debugger_Test):
    ...
```

And then, no need to use a lambda function

cbeauchesne (changes requested):

cbeauchesne (approved):

PR #1595, opened 2023-09-14
[DEBUGGER] Add span and span-decoration method probe snapshot tests

shurivich:
## Description
Add span and span-decoration method probe snapshot tests

### DIFF
--- a/tests/debugger/test_debugger.py
+++ b/tests/debugger/test_debugger.py
@@ -12,20 +12,24 @@ from utils import (
 )
 
 
-def validate_data(expected_probes, expected_snapshots):
+def validate_data(expected_probes, expected_snapshots, expected_traces):
     check_info_endpoint()
 
-    debugger_map = get_debugger_map()
+    log_map = get_debugger_log_map()
     for expected_probe in expected_probes:
-        check_probe_status(expected_probe, debugger_map["probes"])
+        check_probe_status(expected_probe, log_map["probes"])
 
     for expected_snapshot in expected_snapshots:
-        check_snapshot(expected_snapshot, debugger_map["snapshots"])
+        check_snapshot(expected_snapshot, log_map["snapshots"])
+
+    trace_map = get_debugger_tracer_map()
+    for expected_trace in expected_traces:
+        check_trace(expected_trace, trace_map)
 
 
-def get_debugger_map():
+def get_debugger_log_map():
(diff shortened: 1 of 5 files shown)


tests/debugger/test_debugger.py
```diff
@@ -155,19 +183,37 @@ def wait_for_probe(self, data):
...
+        self.span_decoration_probe_response = weblog.get("/debugger/span-decoration/asd/1")
 
     def test_method_probe_snaphots(self):
         assert self.remote_config_is_sent is True
         assert self.probe_installed is True
-        assert self.log_probe_response.status_code == 200
 
-        expected_data = ["logProbe-installed"]
-        validate_data(expected_data, expected_data)
+        assert self.log_probe_response.status_code is 200
```
cbeauchesne, line 202: Even if it works, `is` must be reserved for reference comparizon. See the output form the interpreter : 

image


```suggestion
        assert self.log_probe_response.status_code == 200
```

tests/debugger/test_debugger.py
```diff
@@ -35,13 +39,32 @@ def get_debugger_map():
...
+                    for span in chunk["spans"]:
+                        if span["name"] == "dd.dynamic.span":
+                            span_hash[span["meta"]["debugger.probeid"]] = span
+                        else:
+                            for key, value in span["meta"].items():
+                                if key.startswith("_dd.di"):
+                                    span_hash[value] = span["meta"][key.split(".")[2]]
 
-    return hash
+    print(span_hash)
```
cbeauchesne, line 66: Could you remove the print ?

cbeauchesne (changes requested):

tests/debugger/test_debugger.py
shurivich: Fixed here: de6a537028ac016bb10d972b0903b79c9c02bf26

tests/debugger/test_debugger.py
shurivich: Fixed here de6a537028ac016bb10d972b0903b79c9c02bf26

tests/debugger/test_debugger.py
shurivich: I changed back to `is True` here a04a833aa5bd8caaf3c415187c00b2f6648cdf1d

cbeauchesne (approved):

PR #1598, opened 2023-09-14
Fix a typo in agent.yml

songy23:


cbeauchesne (changes requested):
nope, all version must starts with a `v`.

songy23:
Got it, right now the comparison treats `7.48.0-rcx` as older than `v7.48.0`, sounds like it's the expected behavior

cbeauchesne:
Yes It is, we even have a test that ensure it : https://github.com/DataDog/system-tests/blob/main/tests/test_the_test/test_version.py#L43

PR #1602, opened 2023-09-15
Api security env variable

GustavoCaso:
## Description

Update the ENV variables for the API security base on the latest changes in the RFC.

Disable the Python tests. 

Also, update some of the tests to make the assertion more relax

## Workflow

## Reviewer checklist

* [ ] CI is green

### DIFF
--- a/tests/appsec/api_security/test_schemas.py
+++ b/tests/appsec/api_security/test_schemas.py
@@ -15,8 +15,8 @@ def get_schema(request, address):
     return
 
 
-def equal_without_meta(t1, t2):
-    """compare two schema types, ignoring any metadata"""
+def contains(t1, t2):
+    """validate that schema t1 contains all keys and values from t2 """
     if t1 is None or t2 is None:
         return False
     return equal_value(t1[0], t2[0])
@@ -25,9 +25,9 @@ def equal_without_meta(t1, t2):
 def equal_value(t1, t2):
     """compare two schema type values, ignoring any metadata"""
     if isinstance(t1, list) and isinstance(t2, list):
-        return all(equal_without_meta(a, b) for a, b in zip(t1, t2))
+        return all(contains(a, b) for a, b in zip(t1, t2))
     if isinstance(t1, dict) and isinstance(t2, dict):
-        return all(equal_without_meta(t1[k], t2.get(k)) for k in t2)
+        return all(contains(t1[k], t2.get(k)) for k in t2)
     if isinstance(t1, int) and isinstance(t2, int):
         return t1 == t2
     return False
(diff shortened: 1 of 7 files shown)


tests/appsec/api_security/test_schemas.py
```diff
@@ -25,9 +25,9 @@ def equal_without_meta(t1, t2):
 def equal_value(t1, t2):
     """compare two schema type values, ignoring any metadata"""
     if isinstance(t1, list) and isinstance(t2, list):
-        return len(t1) == len(t2) and all(equal_without_meta(a, b) for a, b in zip(t1, t2))
```
GustavoCaso, line 28: I removed the check for length because some frameworks might be adding specific parameters that could cause this check to fail.

tests/appsec/api_security/test_schemas.py
```diff
@@ -25,9 +25,9 @@ def equal_without_meta(t1, t2):
...
+        return all(equal_without_meta(a, b) for a, b in zip(t1, t2))
     if isinstance(t1, dict) and isinstance(t2, dict):
-        return len(t1) == len(t2) and all(equal_without_meta(t1[k], t2.get(k)) for k in t1)
```
GustavoCaso, line 30: Here I remove the check for length as above and also take the keys from `t2` rather than `t1` to make the assertion more relaxed. If all the keys are present in the one returned from weblog we can consider the test successful. In the case of ruby we were returning extra ones that the framework added

tests/appsec/api_security/test_schemas.py
```diff
@@ -147,8 +154,7 @@ def test_request_method(self):
         assert isinstance(schema, list)
         assert len(schema) == 1
         assert isinstance(schema[0], dict)
-        for key in ("content-length", "content-type", "x-option"):
-            assert key in schema[0]
```
GustavoCaso, line 151: The weblog endpoint only documentation specifies that the parameters by use as the response headers https://github.com/DataDog/system-tests/blob/8da99ebdd9f1bd8269c61752f583873d75b4bb7f/docs/weblog/README.md#L139-L140

In this case, checking `"content-length", "content-type"` is not following the specification and rather accounting for the framework to include those 

I think in this case is best to just check for `x-option`

cbeauchesne (commented):
I've updated the branch, and added the new label `run-appsec-scenarios`

tests/appsec/api_security/test_schemas.py
```diff
@@ -107,11 +107,18 @@ def test_request_method(self):
         assert self.request.status_code == 200
         assert schema
         assert isinstance(schema, list)
-        # There should have two parameters here, one for api_match_AS003, the other for 200
-        assert len(schema[0]) == 2
-        assert all(isinstance(v, list) for v in schema[0].values())
-        assert all(1 <= len(v) <= 2 for v in schema[0].values())
-        assert all(isinstance(v[0], int) for v in schema[0].values())
+        if context.library == "ruby" and context.weblog_variant == "sinatra14":
```
christophe-papazian, line 110: I would like the test to be the same for all libraries and framework, to underline the minimal common ground for all of us and help other libraries to support the feature. So I would like you to remove that condition if you agree?

cbeauchesne (approved):

christophe-papazian (changes requested):

tests/appsec/api_security/test_schemas.py
cbeauchesne: I 'm also not a fan of `ifs` in tests, it'd be better to have a more "open" tests that all tracers passes. But so far, not a blocker for me, up to you to decide.

christophe-papazian (approved):

PR #1632, opened 2023-09-28
[Debugger] Add mix log test scenario + refactor

shurivich:
## Description
Add the "mix log" test scenario.
Support RCM configuration generation with probe IDs in the guid format.
Split the test validations.


manifests/golang.yml
```diff
@@ -284,6 +284,7 @@ tests/:
       Test_Debugger_Line_Probe_Statuses: missing_feature
       Test_Debugger_Method_Probe_Snaphots: irrelevant
       Test_Debugger_Method_Probe_Statuses: missing_feature
+      Test_Debugger_Mix_Log_Probe: irrelevant
```
cbeauchesne, line 287: It means it won't never be implemented in golang ?

shurivich: It will be implemented eventually, but not in the plan for a really long time.

PR #1651, opened 2023-10-03
Remove missing_feature for PHP in 128 bit trace id tests

bwoebi:
I have no idea why these tests are disabled, but they work.


cbeauchesne (changes requested):
CI is failing on activated tests

cbeauchesne (changes requested):
Failing on `test_w3c_128_bit_propagation_tid_in_parent`

bwoebi:
I see, the test is broken, there's no meta on the child, because it's empty.

bwoebi:
@cbeauchesne Now it's actually green :-)

cbeauchesne (approved):

PR #1655, opened 2023-10-03
Add parametric tests for OpenTelemetry -> Datadog name conventions

bouwkast:
## Description

Maps the span kind and various tags on OpenTelemetry spans to a corresponding Datadog Span `OperationName`.
Additionally, test for the special tags to override various Datadog tags: `operation.name`, `resource.name`, `span.type`, `service.name` and `analytics.event`

## Motivation

Create a better mapping for the Datadog Operation Name which doesn't have a direct mapping in OpenTelemetry.

## Workflow

## Reviewer checklist

* [ ] CI is green


tests/parametric/test_otel_span_methods.py
```diff
@@ -286,3 +286,585 @@ def test_otel_get_span_context(self, test_agent, test_library):
...
+            with test_library.otel_start_span("otel_span_name", span_kind=SK_PRODUCER) as span:
+                span.end_span()
+        traces = test_agent.wait_for_num_traces(1)
+        trace = find_trace_by_root(traces, OtelSpan(name="otel_span_name"))
+        assert len(trace) == 1
+
+        root_span = get_span(test_agent)
+
+        assert root_span["name"] == "unknown.producer.request"
+        assert root_span["resource"] == "otel_span_name"
```
bouwkast, line 736: Wasn't sure on exactly what to put for the "generic producer" test case(s) and what the expected operation name would be.

tests/parametric/test_otel_span_methods.py
```diff
@@ -286,3 +286,585 @@ def test_otel_get_span_context(self, test_agent, test_library):
...
+            with test_library.otel_start_span("otel_span_name", span_kind=SK_CONSUMER) as span:
+                span.end_span()
+        traces = test_agent.wait_for_num_traces(1)
+        trace = find_trace_by_root(traces, OtelSpan(name="otel_span_name"))
+        assert len(trace) == 1
+
+        root_span = get_span(test_agent)
+
+        assert root_span["name"] == "unknown.consumer.request"
+        assert root_span["resource"] == "otel_span_name"
```
bouwkast, line 761: Wasn't sure on exactly what to put for the "generic consumer" test case(s) and what the expected operation name would be.

tests/parametric/test_otel_span_methods.py
```diff
@@ -286,3 +286,585 @@ def test_otel_get_span_context(self, test_agent, test_library):
...
+            with test_library.otel_start_span("otel_span_name", span_kind=SK_CLIENT) as span:
+                span.set_attributes({"rpc.system": "aws-api"})
+                span.end_span()
+        traces = test_agent.wait_for_num_traces(1)
+        trace = find_trace_by_root(traces, OtelSpan(name="otel_span_name"))
+        assert len(trace) == 1
+
+        root_span = get_span(test_agent)
+
+        assert root_span["name"] == "aws"  # TODO is there more stuff here that needs to be done?
```
bouwkast, line 520: It seemed like from the unofficial spec for the operation name that this may need more granularity here on what the operation name should be here, but I'm not sure what

marcotc: My suggestion is to append `otel_span.get_attribute("rpc.service").lower()` to `root_span["name"]` in the tracer: `root_span["name"] = "aws." + otel_span.get_attribute('rpc.service').lower()`:  https://opentelemetry.io/docs/specs/otel/trace/semantic_conventions/instrumentation/aws-sdk/#common-attributes

This would yield `aws.s3`, `aws.sns`, for example.

tests/parametric/test_otel_span_methods.py
```diff
@@ -286,3 +286,608 @@ def test_otel_get_span_context(self, test_agent, test_library):
...
+
+    @missing_feature(context.library == "go", reason="Not implemented")
+    @missing_feature(context.library == "java", reason="Not implemented")
+    @missing_feature(context.library == "nodejs", reason="Not implemented")
+    @missing_feature(context.library == "dotnet", reason="Not implemented")
+    @missing_feature(context.library == "python", reason="Not implemented")
+    @irrelevant(context.library == "cpp", reason="library does not implement OpenTelemetry")
+    @missing_feature(context.library == "ruby", reason="Not implemented")
+    @missing_feature(context.library == "php", reason="Not implemented")
+    def test_otel_span_operation_name_message_consumer_002(self, test_agent, test_library):
```
zacharycmontoya, line 448: @dougqh it looks like SpanKind.SERVER may also be used by messaging operations. What should the result be in this use case? I'm guessing the same `{messaging.system}.{messaging.operation}`?

tests/parametric/test_otel_span_methods.py
```diff
@@ -286,3 +286,608 @@ def test_otel_get_span_context(self, test_agent, test_library):
...
+
+    @missing_feature(context.library == "go", reason="Not implemented")
+    @missing_feature(context.library == "java", reason="Not implemented")
+    @missing_feature(context.library == "nodejs", reason="Not implemented")
+    @missing_feature(context.library == "dotnet", reason="Not implemented")
+    @missing_feature(context.library == "python", reason="Not implemented")
+    @irrelevant(context.library == "cpp", reason="library does not implement OpenTelemetry")
+    @missing_feature(context.library == "ruby", reason="Not implemented")
+    @missing_feature(context.library == "php", reason="Not implemented")
+    def test_otel_span_operation_name_message_consumer_001(self, test_agent, test_library):
```
zacharycmontoya, line 417: nit: I think it would help to rename the test method use `client` instead of `consumer` to help make it clear that the input span has kind `CLIENT`

```suggestion
    def test_otel_span_operation_name_message_client(self, test_agent, test_library):
```

tests/parametric/test_otel_span_methods.py
```diff
@@ -286,3 +286,608 @@ def test_otel_get_span_context(self, test_agent, test_library):
...
+        """
+            Tests that the operation name will be set to the `rpc.system + ".request"`
+            - Span kind is set to Client
+            - rpc.system is set to something
+
+            (https://opentelemetry.io/docs/specs/otel/trace/semantic_conventions/rpc/)
+        """
+        with test_library:
+            with test_library.otel_start_span("otel_span_name", span_kind=SK_CLIENT) as span:
+                span.set_attributes({"rpc.system": "grpc"})
```
zacharycmontoya, line 551: @dougqh @amarziali I'll restart this comment thread here, but should we special-case the RPC logic when `rpc.system == grpc`? Our v1 operation naming effort has standardized on operation names `grpc.client.request` and `grpc.server.request` so I wonder if we should do the same in our OTEL logic. WDYT?

zacharycmontoya: I'm on board with the general case BTW, so the existing logic in this method looks fine for other values of `rpc.system`

bouwkast: I've updated these to follow this (and per the comments on the doc): `rpc.system` + "." + `span.kind` + "." + `"request"`

tests/parametric/test_otel_span_methods.py
zacharycmontoya: I mean in the DD libraries. Here are the possible SpanKind's and the expected (as of this writing) operation name:

| SpanKind | Operation Name |
|----------|----------------|
| CLIENT   | {messaging.system}.{messaging.operation} |
| PRODUCER | {messaging.system}.{messaging.operation} |
| CONSUMER | {messaging.system}.{messaging.operation} |
| SERVER   | _undefined_ |

We're currently missing a mapping for the SERVER span kind (which is [possible](https://opentelemetry.io/docs/specs/otel/trace/semantic_conventions/messaging/#span-kind) for a messaging span) and I think it makes sense to apply the same convention to this case. Does this change make sense?

bouwkast: I've re-arranged the file, but I've kept this `SERVER` case in, but it was missing from our spec document.

[Updated link to OpenTelemetry docs outlining that it can be a server](https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#span-kind)

tests/parametric/test_otel_span_methods.py
bouwkast: The only "special" `rpc.system` is for `aws-api`, the others will use `rpc.system` + "." + `span.kind` + "." + "request"

tests/parametric/test_otel_span_methods.py
bouwkast: Now updated to follow the changes in our doc

`rpc.system`.`rpc.service`.request or `rpc.system`.request (all lowercase)

Ex -> `aws.s3.request` or `aws.request`

tests/parametric/test_otel_span_methods.py
```diff
@@ -286,3 +286,774 @@ def test_otel_get_span_context(self, test_agent, test_library):
...
+    @missing_feature(context.library == "nodejs", reason="Not implemented")
+    @missing_feature(context.library == "dotnet", reason="Not implemented")
+    @missing_feature(context.library == "python", reason="Not implemented")
+    @irrelevant(context.library == "cpp", reason="library does not implement OpenTelemetry")
+    @missing_feature(context.library == "ruby", reason="Not implemented")
+    @missing_feature(context.library == "php", reason="Not implemented")
+    def test_otel_span_analytics_sample_rate_override(self, test_agent, test_library):
+        """
+            Tests that the metric for the analytics sample rate (_dd1.sr.eausr) will be set when:
+            - analytics.event is present with "true" or "false"
```
marcotc, line 1044: TIL.
Should we test for the `false` case as well?

tests/parametric/test_otel_span_methods.py
```diff
@@ -286,3 +286,774 @@ def test_otel_get_span_context(self, test_agent, test_library):
...
                     assert context.get("trace_flags") == "01"
+
+    @missing_feature(context.library == "go", reason="Not implemented")
+    @missing_feature(context.library == "java", reason="Not implemented")
+    @missing_feature(context.library == "nodejs", reason="Not implemented")
+    @missing_feature(context.library == "dotnet", reason="Not implemented")
+    @missing_feature(context.library == "python", reason="Not implemented")
+    @irrelevant(context.library == "cpp", reason="library does not implement OpenTelemetry")
+    @missing_feature(context.library == "ruby", reason="Not implemented")
+    @missing_feature(context.library == "php", reason="Not implemented")
```
robertomonteromiguel, line 297: If you are going to mark as "missing_feature" all method of the class for one language, you should remove these class decorator and use manifest files instead. Check manifests folder and doc: https://github.com/DataDog/system-tests/blob/main/docs/RFCs/manifest.md

robertomonteromiguel (changes requested):
There are test failures in parametric

tests/parametric/test_otel_span_methods.py
```diff
@@ -286,3 +286,774 @@ def test_otel_get_span_context(self, test_agent, test_library):
...
+            - http.request.method is set to something (e.g., GET in this example)
+
+            (https://opentelemetry.io/docs/specs/otel/trace/semantic_conventions/http/)
+        """
+        with test_library:
+            with test_library.otel_start_span("otel_span_name", span_kind=SK_SERVER) as span:
+                span.set_attributes({"http.request.method": "GET"})
+                span.end_span()
+        traces = test_agent.wait_for_num_traces(1)
+        trace = find_trace_by_root(traces, OtelSpan(name="otel_span_name"))
```
PerfectSlayer, line 311: Have you tried the test on your tracer?
Here, you are looking for similar span by comparing OTel span names **which are resource names** to DD span names **which are operation names.**

How is this supposed to work? 🤔

PerfectSlayer (commented):
Thanks for the test suite but I fail to understand how they are supposed to work (wrt my inline comment).

According our Slack discussion I push:
* [2c6328c](https://github.com/DataDog/system-tests/pull/1655/commits/2c6328ca45cb9e867ade62961c806030e5effd22) as a fix for looking at OTel span from their names,
* [4a38ab5](https://github.com/DataDog/system-tests/pull/1655/commits/4a38ab5d60b4cabdb4bbb7b33a2178cf4792a59b) as a fix to resource name test.

PerfectSlayer (changes requested):
All existing test methods (like `test_otel_start_span`) also need to be updated.
You might need to duplicate them: 
* one version for before the naming fix,
* another for after the naming fix, only enabled using library version selector.

tests/parametric/test_otel_span_methods.py
bouwkast: Thanks for catching this and fixing it!

tests/parametric/test_otel_span_methods.py
bouwkast: Thanks @robertomonteromiguel I didn't know about the manifest files - I added to `cpp`, `ruby`, and `php`'s in https://github.com/DataDog/system-tests/pull/1655/commits/0a1f682d78999a11c3dce37dafe556b1fa1731ad

tests/parametric/test_otel_span_methods.py
```diff
@@ -348,3 +321,720 @@ def test_otel_get_span_context(self, test_agent, test_library):
...
+    @missing_feature(context.library == "nodejs", reason="Not implemented")
+    @missing_feature(context.library == "dotnet", reason="Not implemented")
+    @missing_feature(context.library == "python", reason="Not implemented")
+    @missing_feature(context.library == "python_http", reason="Not implemented")
+    def test_otel_span_operation_name_message_server(self, test_agent, test_library):
+        """
+            Tests that the operation name will be set to `messaging.system + "." + messaging.operation` when:
+            - Span kind is set to Server
+            - messaging.system is set to something
+            - messaging.operation is set to something
```
PerfectSlayer, line 504: I can't find this rule (with server kind) in the spec.

bouwkast: Yep, there was a comment from @zacharycmontoya  a while ago about whether or not we should add the case for `server` kind for messaging, but I think it has gotten lost with all of my commits/reorganizing in this file

https://github.com/DataDog/system-tests/pull/1655#discussion_r1376729962

I've asked about whether or not it should actually be added as it wasn't in our spec but does appear to be a valid span kind in OpenTelemetry for messaging https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#span-kind

bouwkast:
> All existing test methods (like `test_otel_start_span`) also need to be updated. You might need to duplicate them:
> 
> * one version for before the naming fix,
> * another for after the naming fix, only enabled using library version selector.

@PerfectSlayer I've duplicated the tests that I think need support for old naming vs new naming in https://github.com/DataDog/system-tests/pull/1655/commits/397faeb45a73a9f3567341b8bd1b39f290b263b5

In a few cases I've added a span kind to just better specify what the expected operation name should be.

I didn't add any of the version selectors for the various libraries as I wasn't too sure what to put there for now - I was thinking that maybe as the tracer libraries update to go from old -> new `Operation Names` that they'd be added then?

I don't have .NET's implementation for the new operation name mapping, so unsure if these new duplicates that are skipped for all libraries actually work at the moment.

tests/parametric/test_otel_span_methods.py
bouwkast: I've added a new test for the `false` case in https://github.com/DataDog/system-tests/pull/1655/commits/54e8c19b437c926ec78afe79d86d35546316cf21

I don't know if there is a way of parameterizing tests for these instead of me just copy/pasting the test, I looked around but couldn't find examples of that so unsure if it is a thing.

tests/parametric/test_otel_span_methods.py
```diff
@@ -45,10 +42,35 @@ def test_otel_start_span(self, test_agent, test_library):
...
-    def test_otel_set_service_name(self, test_agent, test_library):
+    @missing_feature(context.library == "golang", reason="New operation name mapping not yet implemented")
+    @missing_feature(context.library == "java", reason="New operation name mapping not yet implemented")
+    @missing_feature(context.library == "nodejs", reason="New operation name mapping not yet implemented")
+    @missing_feature(context.library == "dotnet", reason="New operation name mapping not yet implemented")
+    @missing_feature(context.library == "python", reason="New operation name mapping not yet implemented")
+    @missing_feature(context.library == "python_http", reason="New operation name mapping not yet implemented")
+    def test_otel_start_span(self, test_agent, test_library):
+        """
+            - Start/end a span with start and end options
```
katiehockman, line 53: So this is really checking that the `span_kind` is used as the operation name, by default, if nothing else is provided, right? Or is it testing something else?

tests/parametric/test_otel_span_methods.py
```diff
@@ -61,12 +83,29 @@ def test_otel_set_service_name(self, test_agent, test_library):
...
+        assert root_span["name"] == "internal"
+        assert root_span["resource"] == "parent_span"
+        assert root_span["service"] == "new_service"
+
     @irrelevant(context.library >= "java@1.22.0", reason="library implements the new array encoding")
     @missing_feature(context.library == "nodejs", reason="Empty string attribute value are not supported")
-    @missing_feature(context.library == "php", reason="Not implemented")
-    @missing_feature(context.library == "ruby", reason="Not implemented")
-    def test_otel_set_attributes_different_types(self, test_agent, test_library):
+    def test_otel_set_attributes_different_types_old_name(self, test_agent, test_library):
```
katiehockman, line 108: Is this going to be deleted then, since it now has "old" in the test name? It looks very similar to the following test.

tests/parametric/test_otel_span_methods.py
```diff
@@ -330,21 +588,797 @@ def test_otel_set_span_status_ok(self, test_agent, test_library):
...
+                span.set_attributes({"http.request.method": "GET"})
+                span.end_span()
+        traces = test_agent.wait_for_num_traces(1)
+        trace = find_trace_by_root(traces, otel_span(name="otel_span_name"))
+        assert len(trace) == 1
+
+        root_span = get_span(test_agent)
+
+        assert root_span["name"] == "http.server.request"
+        assert root_span["resource"] == "otel_span_name"
```
katiehockman, line 668:
```suggestion
        span = get_span(test_agent)
        assert span["name"] == "http.server.request"
        assert span["resource"] == "otel_span_name"
```
nit: just to keep things shorter and simpler, given the number of tests in this file. (Here and throughout)

tests/parametric/test_otel_span_methods.py
```diff
@@ -330,21 +588,797 @@ def test_otel_set_span_status_ok(self, test_agent, test_library):
...
+        """
+            Tests that the operation name will be set to `db.system + "." + "query" when:
+            - Span kind is set to Client
+            - db.system is set to some messaging system (e.g., redis in this example)
+
+            (https://opentelemetry.io/docs/specs/semconv/database/database-spans/)
+        """
+        with test_library:
+            with test_library.otel_start_span("otel_span_name", span_kind=SK_CLIENT) as span:
+                span.set_attributes({"db.system": "redis"})
```
katiehockman, line 713: For all of these, let's just put a capital letter somewhere in them, to make sure it gets put to lowercase. e.g. this can be `"Redis"` and it should yield `"redis.query"`

tests/parametric/test_otel_span_methods.py
```diff
@@ -330,21 +588,797 @@ def test_otel_set_span_status_ok(self, test_agent, test_library):
...
+                span.set_attributes({"messaging.operation": "receive"})
+                span.end_span()
+        traces = test_agent.wait_for_num_traces(1)
+        trace = find_trace_by_root(traces, otel_span(name="otel_span_name"))
+        assert len(trace) == 1
+
+        root_span = get_span(test_agent)
+
+        assert root_span["name"] == "kafka.receive"
+        assert root_span["resource"] == "otel_span_name"
```
katiehockman, line 838: You could consider creating a shared test the just accepts a span_kind that's different, to avoid code duplication. There's probably a few places in these tests where doing that would make it a bit easier to read.

I actually wonder if you could make a shared test utility for all/most of these tests. Basically the utility will create, end, and verify a span given the `span_kind` (optional), a map of attributes (optional), and the expected span name to verify. This might make a lot of these tests easy to understand at a glance.

tests/parametric/test_otel_span_methods.py
katiehockman: Sure sounds good, we can add `SERVER` to the list. I'll update the doc.

tests/parametric/test_otel_span_methods.py
```diff
@@ -330,21 +588,797 @@ def test_otel_set_span_status_ok(self, test_agent, test_library):
...
+
+    @missing_feature(context.library == "golang", reason="Not implemented")
+    @missing_feature(context.library == "java", reason="Not implemented")
+    @missing_feature(context.library == "nodejs", reason="Not implemented")
+    @missing_feature(context.library == "dotnet", reason="Not implemented")
+    @missing_feature(context.library == "python", reason="Not implemented")
+    @missing_feature(context.library == "python_http", reason="Not implemented")
+    def test_otel_span_operation_name_aws_client_002(self, test_agent, test_library):
+        """
+            Tests that the operation name will be set to `"aws.request" :
```
katiehockman, line 877: Looking at the sample code again, I think this should be `aws.client.request` to align with the others in that case. LMK if you disagree. I've gone ahead and updated the doc to reflect this.

tests/parametric/test_otel_span_methods.py
```diff
@@ -330,21 +588,797 @@ def test_otel_set_span_status_ok(self, test_agent, test_library):
...
+                span.end_span()
+        traces = test_agent.wait_for_num_traces(1)
+        trace = find_trace_by_root(traces, otel_span(name="otel_span_name"))
+        assert len(trace) == 1
+
+        root_span = get_span(test_agent)
+
+        assert root_span["name"] == "http.server.request"
+        assert root_span["resource"] == "otel_span_name"
+        assert root_span["metrics"].get("_dd1.sr.eausr") == "1.0"
```
katiehockman, line 1358: I personally think all of these could just be combined in one test that sets all of those attributes, and checks them all, to make it a little easier to maintain.

katiehockman (commented):
This is great, thanks so much, Steven! I verified that it matches the RFC.

tests/parametric/test_otel_span_methods.py
bouwkast: Deleted all the old ones in favor of the new ones in 4cb9bb49946547e19f35485a611bea6e56a23da2
I need to unskip them though still

tests/parametric/test_otel_span_methods.py
bouwkast: Thanks, done throughout in 74b668ed532bbe54657cc2701cecddf8491d71e9

tests/parametric/test_otel_span_methods.py
bouwkast: I added a capital to tags that map to the operation name in c029a1dd21878885ac5cf8d5faf3ea1e998765a3

tests/parametric/test_otel_span_methods.py
bouwkast: Thanks, updated in a109769a3d4df75228281f351d03035e2900fad9

tests/parametric/test_otel_span_methods.py
bouwkast: Combined them in a7487a14a48a932f5f5c5bb6fcc60c9cc6d2e5b0

tests/parametric/test_otel_span_methods.py
bouwkast: So this was a pre-existing one (I also previously added an `old_name` version of this but now removed in 4cb9bb49946547e19f35485a611bea6e56a23da2).

But yes, this is checking that the `span_kind` will be used for the operation name now, but also I think this one is specifically that the duration is set correctly

tests/parametric/test_otel_span_methods.py
bouwkast: That's a great suggestion, I'll see if I can make a utility for this

bouwkast: I've parametrized and combined the message ones with just a `@parametrize` for each of the span kinds in 6560381683418f94feec1ec36f7632a4450b4078

I can continue looking into reducing the rest of the duplication, but figured that seemed like an easy change to make

manifests/php.yml
```diff
@@ -70,6 +70,7 @@ tests/:
   parametric/:
     test_dynamic_configuration.py:
       TestDynamicConfigV1: missing_feature
+    test_otel_span_methods.py: missing_feature
```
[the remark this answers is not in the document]
pierotibou: It's not yet merged. The OTEL tests before this PR were passing locally, but we'll change that parameter when merging our support. cc @PROFeNoM

tests/parametric/test_otel_span_methods.py
bouwkast: I've made a helper method for the operation name tests in 6f1b74be2c55e871ee16c45a35a16906b00e23f3 and combined the other span kind ones with a `@parametrize`

tests/parametric/test_otel_span_methods.py
```diff
@@ -348,3 +377,437 @@ def test_otel_get_span_context(self, test_agent, test_library):
...
+
+        span = get_span(test_agent)
+        assert span["name"] == "overriden.name"
+        assert span["resource"] == "new.name"
+        assert span["service"] == "new.service.name"
+        assert span["type"] == "new.span.type"
+        assert span["metrics"].get("_dd1.sr.eausr") == "1.0"
+
+
+def run_operation_name_test(name: str, expected_name: str, span_kind: int, attributes: dict, test_library, test_agent):
```
katiehockman, line 803: Let's change `name` to `resource`. I realize that in the context of OTel, "name" is equivalent to "resource name", but I fear this will be confusing when `expected_name` is also there and means something different. So I'd suggest: `expected_operation_name`, `resource`, `span_kind`, `attributes`

tests/parametric/test_otel_span_methods.py
```diff
@@ -348,3 +377,437 @@ def test_otel_get_span_context(self, test_agent, test_library):
...
+    @missing_feature(context.library == "golang", reason="Not implemented")
+    @missing_feature(context.library == "java", reason="Not implemented")
+    @missing_feature(context.library == "nodejs", reason="Not implemented")
+    @missing_feature(context.library == "dotnet", reason="Not implemented")
+    @missing_feature(context.library == "python", reason="Not implemented")
+    @missing_feature(context.library == "python_http", reason="Not implemented")
+    def test_otel_span_special_tag_overrides(self, test_agent, test_library):
+        """
+            Tests that the special tags will override expected values
+        """
```
katiehockman, line 781: Let's call these "reserved attributes" instead, just to make this very clear what they are. (This came from a discussion with my team earlier today)
I'll update the docs and tickets accordingly

bouwkast: Done in 5d3887ccdb9ab5d9ace73786b4df961bbbf92679

katiehockman (approved):

tests/parametric/test_otel_span_methods.py
```diff
@@ -348,3 +377,442 @@ def test_otel_get_span_context(self, test_agent, test_library):
...
+    @missing_feature(context.library == "python_http", reason="Not implemented")
+    def test_otel_span_operation_name_http_server(self, test_agent, test_library):
+        """
+            Tests that the operation name will be set to "http.server.request" when:
+            - Span kind is set to Server
+            - http.request.method is set to something (e.g., GET in this example)
+
+            (https://opentelemetry.io/docs/specs/otel/trace/semantic_conventions/http/)
+        """
+        run_operation_name_test(
```
zacharycmontoya, line 395: Nice design! This makes it extremely easy to understand the inputs and outputs

zacharycmontoya (approved):

robertomonteromiguel (approved):

tests/parametric/test_otel_span_methods.py
PerfectSlayer: I still think most of the naming tests can be summarise in a single parametrised test.
That is what I did for Java tracer tests and it feels easier to ready: https://github.com/DataDog/dd-trace-java/pull/6104/files#diff-031f3e359dd7470cc178e857ab513ec88199f0365848a0478317cc2e2f252439R54-R90

bouwkast: @PerfectSlayer Yeah that seems to help a lot and will probably make it easier to add more cases if needed -> fba65e977761ab88aebe20214fbb6aa63d10060f

PerfectSlayer (approved):
Looks great! Thanks for taking care of this test suite 💪  and glad I could help 🤝

tests/parametric/test_otel_span_methods.py
```diff
@@ -348,3 +377,93 @@ def test_otel_get_span_context(self, test_agent, test_library):
...
+            ("consumer", SK_CONSUMER, None),
+            ("producer", SK_PRODUCER, None),
+            ("otel_unknown", None, None),
+        ],
+    )
+    def test_otel_span_operation_name(
+        self, expected_operation_name: str, span_kind: int, attributes: dict, test_agent, test_library
+    ):
+        run_operation_name_test(
+            resource="otel_span_name",
```
katiehockman, line 418: nit: if you are always using the same resource name, then just use it in `run_operation_name_test` and don't make it a parameter. i.e. only needs to be a parameter if it's ever different.

pierotibou (approved):
Good for PHP, FYI @PROFeNoM

katiehockman (approved):

PR #1662, opened 2023-10-04
Add nextjs application to test suite

uurien:
## Description

Add nextjs application to nodejs system tests.
## Workflow

## Reviewer checklist

* [ ] CI is green


manifests/nodejs.yml
```diff
@@ -91,16 +91,22 @@ tests/:
       test_addresses.py:
         Test_BodyJson: v2.2.0
         Test_BodyRaw: missing_feature
-        Test_BodyUrlEncoded: v2.2.0
-        Test_BodyXml: v2.2.0
+        Test_BodyUrlEncoded: 
+            '*': v2.2.0
+            nextjs: missing_feature  # request.text() method does not decode the content, so can't detect the threat
```
cbeauchesne, line 96: If it will never be supported by the framework, you can use `irrelevant`.

manifests/nodejs.yml
```diff
@@ -91,87 +91,189 @@ tests/:
...
+          nextjs: v4.17.0
         Test_FullGrpc: missing_feature
-        Test_GraphQL: v4.16.0
-        Test_Headers: v2.0.0
+        Test_GraphQL:
+          '*': v4.16.0
+          nextjs: irrelevant # nextjs is not related with graphql
+        Test_Headers:
+          '*': v2.0.0
+          nextjs: v4.17.0
```
simon-id, line 114: Aren't our headers protection framework independant ?

simon-id: same thing for status code, raw url, etc...

manifests/nodejs.yml
```diff
@@ -91,87 +91,189 @@ tests/:
...
+        Test_UrlQueryKey:
+          '*': v2.6.0
+          nextjs: v4.17.0
+        Test_UrlRaw:
+          '*': v2.0.0
+          nextjs: v4.17.0
         Test_gRPC: missing_feature
       test_blocking.py:
         Test_Blocking: v3.19.0
```
simon-id, line 134: Don't we need to add framework info for blocking ?

cbeauchesne (commented):
I converted the PR to draft. You can set it back to normal once you need a review from R&P team.

manifests/nodejs.yml
uurien: I had some doubts here. To add version or not... yes headers/url... features are independent, but "officially" nextjs is not supported until 4.17.0...
But you're right, is should work so I'll remove the nextjs specific version to these tests.

manifests/nodejs.yml
uurien: Yes, I think that I just forgot this test because they are just working in nexts...
But all test are related with headers, so we should support because are independant. As a iam removing `nextjs` reference from Headers and other independant test, I'll maintain this as it is now.

manifests/nodejs.yml
```diff
@@ -94,87 +94,154 @@ tests/:
...
+        Test_Cookies:
+          '*': v2.0.0
         Test_FullGrpc: missing_feature
-        Test_GraphQL: v4.16.0
-        Test_Headers: v2.0.0
+        Test_GraphQL:
+          '*': v4.16.0
+          nextjs: irrelevant # nextjs is not related with graphql
+        Test_Headers:
+          '*': v2.0.0
```
simon-id, line 115: Turning a simple `Test_Headers: v2.0.0` into this is ew imo, especially when doing it everywhere, it decreases readability

uurien: You're absolutely right. I modified it to add nextjs but when I remove it, I maintained it as before 🤦

tests/appsec/test_blocking_addresses.py
```diff
@@ -237,7 +240,7 @@ def test_blocking_before(self):
         """Test that blocked requests are blocked before being processed"""
         # first request should not block and must set the tag in span accordingly
         assert self.set_req1.status_code == 200
-        assert self.set_req1.text == "Value tagged"
+        assert self.set_req1.text == "Value tagged" or self.set_req1.text == '"Value tagged"'
```
uurien, line 243: In next we should respond with JSON, we can respond quoted string `"Value tagged"` but not just `Value tagged`

tests/appsec/waf/test_rules.py
```diff
@@ -309,7 +309,7 @@ class Test_DiscoveryScan:
     """AppSec WAF Tests on Discovery Scan rules"""
 
     def setup_security_scan(self):
-        self.r1 = weblog.get("/etc/")
+        self.r1 = weblog.get("/etc/something")
```
uurien, line 312: Next makes automatic redirect when the url is finished by `/`, and the rule says that path should contain `/etc/` and not `/etc`, so I modified the url to prevent finishing `/` and also contain `/etc/`.

utils/build/docker/nodejs/nextjs/.eslintrc.json
```diff
@@ -0,0 +1,39 @@
+{
```
uurien, line 1: Based in dd-trace-js linting file.

tests/appsec/test_reports.py
```diff
@@ -14,7 +14,7 @@ class Test_StatusCode:
     """Appsec reports good status code"""
 
     def setup_basic(self):
-        self.r = weblog.get("/path_that_doesn't_exists/", headers={"User-Agent": "Arachni/v1"})
+        self.r = weblog.get("/path_that_doesn't_exists", headers={"User-Agent": "Arachni/v1"})
```
uurien, line 17: in nextjs, when the url ends with `/` it makes automatic redirect, generating extra traces.

tests/appsec/waf/test_rules.py
simon-id: Maybe it's possible to toggle that feature off in next settings ? it's a bit weird to change a test just because one node framework is a bit stupid.

uurien:
Reopening the PR because because now the failures don't look related with the PR

cbeauchesne (approved):

tests/appsec/waf/test_rules.py
uurien: I can't find the config property for it...

uurien: Yes, It is configurable... but

> By default Next.js will redirect urls with trailing slashes to their counterpart without a trailing slash. For example /about/ will redirect to /about. You can configure this behavior to act the opposite way, where urls without trailing slashes are redirected to their counterparts with trailing slashes.

It can be configured to redirect from `path/` to `path` or from `path` to `path/`

tests/appsec/test_blocking_addresses.py
simon-id: Since this snippet is repeated 3 times, you could always do something like `assert self.set_req1.text.includes("Value tagged")` (not sure how you do that in python), so it matches a bit more broadly. But up to you

simon-id (approved):

tests/appsec/test_blocking_addresses.py
cbeauchesne:
```python
assert "Value tagged" in self.set_req1.text
```

simon-id (approved):

PR #1777, opened 2023-10-31
parametric: check credentials are removed from ssh urls.

lraucy:
## Description

The PR ensures that credentials in `ssh://` urls are removed when processing DD_GIT_REPOSITORY_URL for APM spans.

## Motivation

When using git over ssh, two syntax for specifying the address exist: scp-like syntax and `ssh://` syntax. scp-like syntax doesn't allow passing a password in the URL, but `ssh://` syntax does. We want to make sure all tracers strip this password, when this is happening

## Workflow

## Reviewer checklist

* [ ] CI is green


tests/parametric/test_tracer.py
```diff
@@ -101,10 +101,18 @@ def test_tracer_commit_sha_environment_variable(
                 "DD_GIT_REPOSITORY_URL": "https://gitlab-ci-token:AAA_bbb@gitlab.com/DataDog/systems-test.git",
                 "expected_repo_url": "https://gitlab.com/DataDog/systems-test.git",
             },
+            {
+                "DD_GIT_REPOSITORY_URL": "ssh://gitlab-ci-token:AAA_bbb@gitlab.com/DataDog/systems-test.git",
```
Letiste, line 105: What are the possible characters for the credentials? Should we also expect to receive `@`, `/`or other special characters?

robertomonteromiguel (approved):

tests/parametric/test_tracer.py
lraucy: those characters would make the URL invalid. The exact character set allowed is defined here https://datatracker.ietf.org/doc/html/rfc1738#section-2.2
and additionally https://datatracker.ietf.org/doc/html/rfc1738#section-3.1

```
The user name (and password), if present, are followed by a
   commercial at-sign "@". Within the user and password field, any ":",
   "@", or "/" must be encoded.
```

PR #1801, opened 2023-11-07
[node.js] express4 app linter [APPSEC-10464]

iunanua:
## Description

Adds eslint config file, dependencies and fixes linter errors

## Workflow

## Reviewer checklist

* [ ] CI is green


utils/build/docker/nodejs/express4/iast/index.js
```diff
@@ -31,107 +31,107 @@ function initMiddlewares (app) {
...
-    const span = tracer.scope().active();
-    span.setTag('appsec.event"', true);
- 
-    const key = crypto.randomBytes(32);
- 
-    const iv = crypto.randomBytes(16);
- 
-    const cipher = crypto.createCipheriv('aes-256-cbc', Buffer.from(key), iv);
+    const span = tracer.scope().active()
+    span.setTag('appsec.event"', true)
```
iunanua, line 84: Why is there a " in the tag value `'appsec.event"'`? :S

utils/build/docker/nodejs/express4/.eslintrc.json
simon-id, line 1: did you get this from dd-trace ?

utils/build/docker/nodejs/express4/app.js
```diff
@@ -1,130 +1,132 @@
-"use strict";
+/* eslint-disable no-console */
```
simon-id, line 1: disable it directly in the eslint config file i'd say

utils/build/docker/nodejs/express4/app.js
```diff
@@ -165,77 +167,80 @@ app.get("/dsm", (req, res) => {
...
+})
 
 app.get('/load_dependency', (req, res) => {
-  console.log('Load dependency endpoint');
-  const glob = require("glob")
-  res.send("Loaded a dependency")
-});
+  console.log('Load dependency endpoint')
+  // eslint-disable-next-line no-unused-vars
+  const glob = require('glob')
```
simon-id, line 190: would that work ?
```suggestion
  require('glob')
```

utils/build/docker/nodejs/express4/.eslintrc.json
iunanua: [from the system-tests nextjs app](https://github.com/DataDog/system-tests/blob/main/utils/build/docker/nodejs/nextjs/.eslintrc.json)

iunanua: but probably it comes from dd-trace

utils/build/docker/nodejs/express4/app.js
```diff
@@ -165,77 +167,80 @@ app.get("/dsm", (req, res) => {
...
 
   const response = spawnSync(req?.body?.command, args, options)
-  
+
   res.send(response)
 })
 
 app.get('/createextraservice', (req, res) => {
+  // eslint-disable-next-line dot-notation
   const serviceName = req.query['serviceName']
```
simon-id, line 244:
```suggestion
  const serviceName = req.query.serviceName
```

utils/build/docker/nodejs/express4/iast/index.js
simon-id: `appsec.event: true` is used to signal a WAF attack has been detected. Now why is it used here i have no idea haha

iunanua: it will have no effect anyway because the tests are setting `'appsec.event"'` and not `'appsec.event'`

utils/build/docker/nodejs/express4/integrations/db/postgres.js
```diff
@@ -1,87 +1,85 @@
...
+      client.query(query).then((queryResult) => {
+        console.log('Postgres query result:' + queryResult)
+        resolve('Postgres query done!')
+      }).catch((err) => {
+        console.log(err)
+        resolve('Error on postgres query')
+      })
+    }).catch((err) => {
+      console.log(err)
+      reject(err)
```
simon-id, line 20: Heads up: that's not a linting change

utils/build/docker/nodejs/express4/integrations/db/postgres.js
```diff
@@ -1,87 +1,85 @@
...
-    return await launchQuery(query);
+async function initData () {
+  console.log('loading sql data')
+  const query = readFileSync(join(__dirname, 'resources', 'postgres.sql')).toString()
+  return await launchQuery(query)
 }
 
-async function select() {
-    const sql = "SELECT * FROM demo where id=1 or id IN (3, 4)"
-    const client = new Client()
```
simon-id, line 33: potential side effect here

utils/build/docker/nodejs/express4/integrations/db/postgres.js
iunanua: it's the `prefer-promise-reject-errors` rule.
Do you mean it could change the app behavior?

utils/build/docker/nodejs/express4/integrations/db/postgres.js
iunanua: I don't think so. 
The rest of the methods do not create an instance and tests are passing without it.

utils/build/docker/nodejs/express4/iast/index.js
simon-id: Oh shit I didn't even see the double quote here, that's probably not right. Which also probably means that this setTag is useless

simon-id: @uurien the git blame points at you

utils/build/docker/nodejs/express4/integrations/db/postgres.js
simon-id: true, i'll let the APM team decide

utils/build/docker/nodejs/express4/iast/index.js
uurien: wow, it seems that it is wrong, and we should remove these lines. Probably copied from java test (I checked that they are adding `appsec.event=true`, but it doesn't make any sense.
And as mentioned before... it has a typo with the last `"` ...

simon-id (approved):
LGTM but we should still wait for APM approval too

utils/build/docker/nodejs/express4/integrations/db/postgres.js
simon-id: Yes, idk if it's important or not, but just a heads up for APM to check if they actually want this change or not

uurien:
You have added the file, that is great. But I think that we should add a task in the image building or CI process to check that the new code is not breaking the linter.

robertomonteromiguel:
> You have added the file, that is great. But I think that we should add a task in the image building or CI process to check that the new code is not breaking the linter.

+1

robertomonteromiguel (changes requested):
As Ugaitz says: You have added the file, that is great. But I think that we should add a task in the image building or CI process to check that the new code is not breaking the linter.

utils/build/docker/nodejs/express4/iast/index.js
iunanua: removed

iunanua:
done, but I have added two new steps in the `lint_code` action:
- one to set up node in the runner. Is it too much invasive? If there are problems with that I could skip the installation of node and just check if npm exists before running npm...
- and another one to run npm lint

robertomonteromiguel (approved):

simon-id:
Not sure it's an amazing idea to put the lint job in this file. I'll let charles decide where's the best place

iunanua:
@cbeauchesne wdyt?

cbeauchesne:
> Not sure it's an amazing idea to put the lint job in this file. I'll let charles decide where's the best place

@simon-id what is the alternative ?

.github/actions/lint_code/action.yaml
```diff
@@ -23,3 +23,14 @@ runs:
...
       run: ./utils/scripts/shellcheck.sh
+    - name: Install node
+      uses: actions/setup-node@v3
+      with:
+        node-version: 20
+    - name: 'Run nodejs lint'
+      shell: bash
+      working-directory: ./utils/build/docker/nodejs/express4
+      run: |
+        npm install
```
cbeauchesne, line 34: What about installing only lint deps to  speed up the job ?

simon-id:
@cbeauchesne no idea, just looks weird to add something specific to node in here

cbeauchesne:
No, it's ok for me, I prefer to have in in the lint job : 

* we do not spawn more than 130 worker for nothing
* user are warned sooned

.github/actions/lint_code/action.yaml
iunanua: done!

simon-id: I see you added `-D`. Did you test it to make sure it was only installing the dev deps ? I just tried it and I don't think it's doing that ?

iunanua: There seem to have been different ways: `--dev`, `--only=dev`, `--production=false` and `-D` and I've read comments stating that `-D` was working with the latest npm versions

But I haven't checked it :S

iunanua: There may not be a way to install only the dev dependencies...

uurien (approved):

.github/actions/lint_code/action.yaml
simon-id: Back to doing `npm install eslint eslint-standard....` then i guess

simon-id: what about using the global arg ? `-g` that shouldn't use any package/lock file no ?

uurien: I don't know if it could work with the other dependencies, `eslint` yes, but `eslint-plugin-mocha`,... I think that they should be in `node_modules`.
I mentioned to Igor to try with something like:
```shell
cp package.json package-back.json && node -pe "const pack = require('./package.json');delete pack.dependencies;require('fs').writeFileSync('./package.json',JSON.stringify(pack))" && npm install --no-package-lock axios && mv package-back.json package.json
```

It should install only `devDependencies`, because we are removing `dependencies` from `package.json`.
Just to see how much time can we earn, if the time difference is insignificant... we can just maintain the normal `npm install`

iunanua: Around the same time:
- npm install -> 41s
- your hack -> 43s

cbeauchesne (approved):

PR #1814, opened 2023-11-08
[parametric] Add system-test for client libraries to generate 128 bit trace-id's by default

zacharycmontoya:
## Description

Adds test to assert that tracing client libraries generate 128-bit trace-id's by default. The test behavior is identical to the enabled case.

## Motivation

This test asserts that all libraries implement the new behavior.

## Workflow

## Reviewer checklist

* [ ] CI is green

### DIFF
--- a/tests/parametric/test_128_bit_traceids.py
+++ b/tests/parametric/test_128_bit_traceids.py
@@ -63,7 +63,7 @@ class Test_128_Bit_Traceids:
         assert trace_id == 1234567890123456789
         assert int(headers["x-datadog-trace-id"], 10) == trace_id
         assert dd_p_tid is None
-        assert "_dd.p.tid=" not in headers["x-datadog-tags"]
+        assert "x-datadog-tags" not in headers or "_dd.p.tid=" not in headers["x-datadog-tags"]
 
     @missing_feature(context.library == "nodejs", reason="not implemented")
     @missing_feature(context.library == "python_http", reason="not implemented")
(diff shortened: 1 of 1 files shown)


tests/parametric/test_128_bit_traceids.py
```diff
@@ -197,6 +197,32 @@ def test_datadog_128_bit_generation_enabled(self, test_agent, test_library):
...
+        datadog headers, and populated in trace data.
+        """
+        with test_library:
+            headers = make_single_request_and_get_inject_headers(test_library, [])
+        span = get_span(test_agent)
+        trace_id = span.get("trace_id")
+        dd_p_tid = span["meta"].get("_dd.p.tid")
+
+        assert trace_id < POWER_2_64
+        assert int(headers["x-datadog-trace-id"], 10) == trace_id
```
lucaspimentel, line 223: Since we're validating the lower 64 bits, should we also compare `tid` from `x-datadog-tags` to `dd_p_tid`? I looks like `validate_dd_p_tid()` doesn't compare it to the expected value.

zacharycmontoya: Good call, added in 8e6b1892e859da0f5984456927da3c962ce60ad5

lucaspimentel (approved):

tests/parametric/test_128_bit_traceids.py
```diff
@@ -158,6 +162,7 @@ def test_datadog_128_bit_propagation_and_generation(self, test_agent, test_libra
         assert trace_id == 1234567890123456789
         assert int(headers["x-datadog-trace-id"], 10) == trace_id
         assert dd_p_tid is None
+        assert "_dd.p.tid=" not in headers["x-datadog-tags"]
```
dinooliva: This (and similar additions) isn't wrong but it seems redundant with the previous test.

dinooliva (approved):
The added asserts (both positive and negative) redundant - i.e. they wouldn't be in meta if they weren't in the header (or missing if they were) - so they're not strictly necessary.

That said, they're not wrong so I'll leave it up to your discretion.

robertomonteromiguel (approved):

PR #1829, opened 2023-11-13
When no span.kind, name should default to `internal`

bouwkast:
## Description

We should always be able to rely on `span.kind` for the operation name as it will always have a value - by default the `span.kind` is `SpanKind.Internal`.
This removes the possibility of having `otel_unknown` as the operation name as we can always fallback on `span.kind`.

## Motivation

We will always have a value for `span.kind` to fallback to for the operation name. 

## Workflow

## Reviewer checklist

* [ ] CI is green

### DIFF
--- a/tests/parametric/test_otel_span_methods.py
+++ b/tests/parametric/test_otel_span_methods.py
@@ -434,7 +434,7 @@ class Test_Otel_Span_Methods:
             ("internal", SK_INTERNAL, None),
             ("consumer", SK_CONSUMER, None),
             ("producer", SK_PRODUCER, None),
-            ("otel_unknown", None, None),
+            ("internal", None, None),
         ],
     )
     def test_otel_span_operation_name(


katiehockman:
> If there is no value for span.kind then internal will be used as a fallback, replacing otel_unknown with internal for the name.

I suggest rewording this a bit. It's not that we should fallback to internal as the name, it's that we should always be able to rely on `span.kind` as a fallback, and the default `span.kind` value should be `SpanKind.Internal`

bouwkast:
> > If there is no value for span.kind then internal will be used as a fallback, replacing otel_unknown with internal for the name.
> 
> I suggest rewording this a bit. It's not that we should fallback to internal as the name, it's that we should always be able to rely on `span.kind` as a fallback, and the default `span.kind` value should be `SpanKind.Internal`

Thanks, updated the description

robertomonteromiguel (approved):

PerfectSlayer (approved):
Thanks for the quick update! 👍

PR #1871, opened 2023-11-27
Explicity library flushing

simon-id:
## Description

Introduce an opt-in endpoint `GET /flush` that will be called before running test assertions.

## Motivation

5 seconds is too long.

## Workflow

## Reviewer checklist

* [ ] CI is green


utils/_context/_scenarios.py
```diff
@@ -519,7 +521,19 @@ def post_setup(self):
 
         elif self.use_proxy:
             self._wait_interface(interfaces.library, self.library_interface_timeout)
-            self.weblog_container.stop()
+
+            if self.library in ("nodejs",):
```
simon-id, line 525: @cbeauchesne why not do if else here with _wait_interface ?

simon-id: answer: because some tests might want to wait before flushing (time based tests)

utils/build/docker/nodejs/express4/app.js
```diff
@@ -250,7 +251,37 @@ iast.initRoutes(app, tracer)
 require('./auth')(app, passport, tracer)
 require('./graphql')(app)
 
+// try to flush as much stuff as possible from the library
+app.get('/flush', (req, res) => {
+  // doesn't have a callback :(
+  // tracer._tracer?._dataStreamsProcessor?.writer?.flush?.()
```
simon-id, line 257: TODO: for some reason running this function makes the schema non regression tests fail ???

utils/build/docker/nodejs/nextjs/src/app/flush/route.js
```diff
@@ -0,0 +1,39 @@
+import { promisify } from 'util'
+import { NextResponse } from 'next/server'
+
+export const dynamic = 'force-dynamic'
+
+export async function GET (request) {
+  await flush()
```
simon-id, line 7: the nextjs variant is failing because the request to this endpoint is failing. Idk why, it looks like it's throwing an empty error ?

Julio-Guerra:
Libraries don't expose public SDK functions to flush things out. Gracefully stopping the tracer when weblog quits is the most standard approach to me. WDYT of this limitation where libs like dd-trace-go won't be able to implement this "live flushing"?

simon-id:
@Julio-Guerra the nodejs library do not expose a public api to flush stuff either, i had to dig into internals in a very dirty way.
As you said flushing when the process stops is a good idea and it actually is already the case in the Node library! but it's safer to force flushing asynchronously to be sure it's completed, as the ST will wait for the response to that request before running the assertions.
If your language can flush everything reliably at shutdown then go for that sure, but in our case the shutdown flushing is a best effort only.

utils/build/docker/nodejs/nextjs/src/app/flush/route.js
simon-id: profiling doesn't work with nexjts apparently

docs/internals/flushing.md
```diff
@@ -0,0 +1,16 @@
...
+
+## Implicit flushing
+
+Before running the assertions, the system-tests will stop the weblog application. If your library supports reliably flushing all data at shutdown time, you may use that mecanism to reduce the default delay to 0, and instead rely on that shutdown event. Please make sure your implementation is not flaky, before making that modification.
+
+## Explicit flushing
+
+It is possible to use explicit flushing instead of the default delay.
+To do so, implement an endpoint in your weblog `GET /flush`, that when called, will force the flushing of all library data.
+You then need to add your library to the list of supported explicity flushing in: TODO LINK, and finally you can set your library delay to 0.
```
simon-id, line 16: need to update that link once the code is merged in master

cbeauchesne (approved):

utils/_context/_scenarios.py
```diff
@@ -391,13 +391,15 @@ def configure(self, config):
             container.interface = getattr(interfaces, container.name)
             container.interface.configure(self.replay)
 
-        if self.library_interface_timeout is None:
+        # to override scenarii timeouts like profiling
+        if self.weblog_container.library.library == "nodejs":
+            self.library_interface_timeout = 0
+            self.agent_interface_timeout = 0
```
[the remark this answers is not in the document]
simon-id: Oh nevermind, this is from an old diff, it's not included in the PR anymore :+1:

CarlesDD (approved):

PR #1905, opened 2023-12-11
Implement span links parametric tests using OpenTelemetry API and JSON tag encoding

PerfectSlayer:
## Description

This PR brings parametric tests for Span Links using OpenTelemetry API and span tag encoding.

The test logic is based on @bwoebi original tests for Span Links using Dadadog tracer API: #1811 

## Motivation

The two things (OTel & JSON tag encoding) were missing from the original test implementation.

## Workflow

## Reviewer checklist

* [ ] CI is green


tests/parametric/test_otel_span_methods.py
```diff
@@ -396,6 +398,212 @@ def test_otel_set_attributes_separately(self, test_agent, test_library):
...
+        # If 0, it is not sampled so it won't be send to the agent.
+        # assert (link.get("flags") or 0) == 0
+
+    @missing_feature(context.library < "java@1.26.0", reason="Implemented in 1.26.0")
+    @missing_feature(context.library == "nodejs", reason="Not implemented")
+    @missing_feature(context.library == "dotnet", reason="Not implemented")
+    @missing_feature(context.library == "python", reason="Not implemented")
+    @missing_feature(context.library == "python_http", reason="Not implemented")
+    def test_otel_span_started_with_link_from_datadog_headers(self, test_agent, test_library):
+        """Properly inject datadog distributed tracing information into span links.
```
mabdinur, line 452: Should we always generate span links from distributed trace headers? 

In the python tracer we only generate a span link if conflicting spanids/traceids are found in http headers. In all other cases `x-datadog-parent-id` is only used to set parent-child relationships.

tests/parametric/test_otel_span_methods.py
```diff
@@ -535,3 +743,25 @@ def run_otel_span_reserved_attributes_overrides_analytics_event(
     span = get_span(test_agent)
     assert span["metrics"].get("_dd1.sr.eausr") == expected_metric_value
     assert "analytics.event" not in span["meta"]
+
+
+def retrieve_span_links(span):
```
mabdinur, line 748: It would be nice if we could be explicit about the expected format. This could help us catch regressions:

```suggestion
def retrieve_span_links(span, api="v0.5"):
```

tests/parametric/test_otel_span_methods.py
mabdinur: Ahhh we'll need to test this behavior (from the RFC): 

```
When continuing a distributed trace, the context having extracted it must add a span link referencing the active span in the extraction context, if present, to any span which the distributed tracing context is applied to. (For example, when there is an existing local trace and then an incoming request comes in.)
Informational: This ensures that it will be possible to find the true source context for a span.
Example
Given:
The current active trace context is T1/S1 (TraceID: 1, SpanID: 1),
An incoming request arrives with the trace context T2/S2,
The tracer extracts the incoming trace context and starts a new span (T2/S3) with the extracted incoming context.
```

tests/parametric/test_otel_span_methods.py
```diff
@@ -396,6 +398,212 @@ def test_otel_set_attributes_separately(self, test_agent, test_library):
...
+        This tests the functionality of "create a direct link between two spans
+        given two valid span (or SpanContext) objects" as specified in the RFC.
+        """
+        with test_library:
+            with test_library.otel_start_span("root") as parent:
+                parent.end_span()
+                with test_library.otel_start_span(
+                    "child",
+                    parent_id=parent.span_id,
+                    links=[Link(parent_id=parent.span_id, attributes={"foo": "bar", "array": ["a", "b", "c"]})],
```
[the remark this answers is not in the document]
PerfectSlayer: Agreed, but this logic comes from Bob test suite 😞 
https://github.com/DataDog/system-tests/pull/1811/files#diff-e63af22f5cd81bf600c30dfb393aef024e358dd336ede3786d147f929cab2b54R28

We would need to refactor both to keep the tests consistent

tests/parametric/test_otel_span_methods.py
```diff
@@ -396,6 +398,212 @@ def test_otel_set_attributes_separately(self, test_agent, test_library):
...
+        assert span_links is not None
+        assert len(span_links) == 1
+
+        link = span_links[0]
+        assert link.get("span_id") == root.get("span_id")
+        assert link.get("trace_id") == root.get("trace_id")
+        root_tid = root["meta"].get("_dd.p.tid") or "0" if "meta" in root else "0"
+        assert (link.get("trace_id_high") or 0) == int(root_tid, 16)
+        assert link["attributes"].get("foo") == "bar"
+        assert link["attributes"].get("array.0") == "a"
```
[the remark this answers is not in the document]
PerfectSlayer: I think it is part of the RFC: § Attribute values 😉

tests/parametric/test_otel_span_methods.py
```diff
@@ -396,6 +398,212 @@ def test_otel_set_attributes_separately(self, test_agent, test_library):
...
+    def test_otel_span_started_with_link_from_datadog_headers(self, test_agent, test_library):
+        """Properly inject datadog distributed tracing information into span links.
+        """
+        with test_library:
+            with test_library.otel_start_span(
+                "root",
+                links=[
+                    Link(
+                        http_headers=[
+                            ["x-datadog-trace-id", "1234567890"],
```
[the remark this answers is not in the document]
PerfectSlayer: It is the purpose of the test: to create span links from `distributed tracing information`, so headers here.
The parametric client app is supposed to treat differently programmatic link creation (the thing you suggest) from link creation of extracted tracing information.
This test triggers the second case.

tests/parametric/test_otel_span_methods.py
PerfectSlayer: I am not very familiar to this. What API refers to?
And what would be the purpose?

mabdinur: API refers to the the value set via the `DD_TRACE_API_VERSION` environment variable. It is the the trace encoding format. In python customers can use either `v0.4` or `v0.5`. I think most traces just implement `v0.4` and hide this implementation detail. 

When a span containing links is encoded:
 - With `v0.4` span["span_links"] will be set.
 - With `v0.5` span["meta"]["_dd.span_links"] will be set.
 
This helper already searches for both cases. I think should make this mapping more explicit. This suggestion is honestly just a nit.

tests/parametric/test_otel_span_methods.py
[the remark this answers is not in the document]
mabdinur: According to the spec if there is an active span in a process (Span A) and then ddtrace distributed headers are received (contain data about Span B). Then Span A should be the **parent** of nested spans (Span C) and Span B should be a **link**. To test this logic we do not need to send a link object to the parametric app. We just need to:
- Create a span (Span A)
- Send distributed tracing headers (containing info about Span B)
- Create a second Span (Span C)
- End Span A and Span C
- Then ensure: Span A <- parent-child -> Span C, Span B <- linked -> Span C

I am not sure if I am missing something here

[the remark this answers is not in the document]
PerfectSlayer: The test is here not related to the case you describe.
It is use to test the span link creation from distributed tracing informations:

```
There must be functionality to:
- create a span link from distributed tracing information (e.g. datadog, b3, w3c headers)
```

tests/parametric/test_otel_span_methods.py
```diff
@@ -396,6 +398,212 @@ def test_otel_set_attributes_separately(self, test_agent, test_library):
             assert span["name"] == "kafka.receive"
             assert span["resource"] == "operation"
 
+    @missing_feature(context.library < "java@1.26.0", reason="Implemented in 1.26.0")
+    @missing_feature(context.library == "nodejs", reason="Not implemented")
+    @missing_feature(context.library == "dotnet", reason="Not implemented")
+    @missing_feature(context.library == "python", reason="Not implemented")
+    @missing_feature(context.library == "python_http", reason="Not implemented")
```
mabdinur, line 405: `python_http` was renamed to `python`. The original `python` grpc client was removed (we are planning to convert all grpc parametric endpoints to http).

```suggestion
    @missing_feature(context.library == "python", reason="Not implemented")
```

tests/parametric/test_otel_span_methods.py
```diff
@@ -396,6 +398,212 @@ def test_otel_set_attributes_separately(self, test_agent, test_library):
...
+        """
+        with test_library:
+            with test_library.otel_start_span("root") as parent:
+                parent.end_span()
+                with test_library.otel_start_span(
+                    "child",
+                    parent_id=parent.span_id,
+                    links=[Link(parent_id=parent.span_id, attributes={"foo": "bar", "array": ["a", "b", "c"]})],
+                ) as child:
+                    child.end_span()
```
mabdinur, line 419: nit: The span should ended after the `with` block terminates
```suggestion
                    pass
```

tests/parametric/test_otel_span_methods.py
```diff
@@ -396,6 +398,212 @@ def test_otel_set_attributes_separately(self, test_agent, test_library):
...
+
+        traces = test_agent.wait_for_num_traces(1)
+        trace = find_trace_by_root(traces, otel_span(name="root"))
+        assert len(trace) == 2
+
+        root = find_span(trace, otel_span(name="root"))
+        child = find_span(trace, otel_span(name="child"))
```
mabdinur, line 426: nit: Spans are sorted by their start_time. We can just access the spans by index. The find_span method sometimes results in false matches. Honestly I think we can should remove it but that's beyond the scope of this PR.

```suggestion
        root = trace[0][0]
        child = trace[0][1]
```

tests/parametric/test_otel_span_methods.py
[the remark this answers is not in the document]
mabdinur: The specification has `what need to be tested` but it appears to be missing `how` distributed tracing information should be used to generate span links. The RFC makes some recommendations but it's missing definitive statements. The example I gave above is the most concrete [example](https://docs.google.com/document/d/1RcTjO8ER5FNweJ0rqVcQEbf6tb1PrlPsQ6coZZww8Xg/edit#heading=h.f7rbvhg00e8f) I could find but you're right it may not be relevant to this existing test.

I have these two high level questions:
1. Should there be a special value in the tracestate/x-datadog-tags to distinguish between when span links or parent-child relationships could be created? [decision made by the client/producer]
2. Should the downstream application (ex: consumer/server) make the decision to generate a link depending on the context [multiple parents/sources, special integration is used, traced servers, etc]? 
   - I think this is what we are trying to do but it would be nice if we could create an exhaustive list and have a test case for each scenario.

 
@bwoebi What are your thoughts here? Should we spell out some specific test cases in the RFC?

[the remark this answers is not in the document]
mabdinur: The python tracer can generate span links from distributed tracing using the following code:

```python
 # http_headers={"x-datadog-trace-id": "1234567890", "x-datadog-parent-id": "9876543210", "x-datadog-sampling-priority": "2", "x-datadog-tags": "_dd.p.dm=-4,_dd.p.tid=0000000000000010"}

from ddtrace import tracer
from ddtrace.propagation.http import HTTPPropagator

context = HTTPPropagator.extract(headers=http_headers)
span = tracer.trace("span")
span.link_span(context)
```

3. Is this sufficient to say the python tracer supports generating span links from distributed tracing headers or should there be a helper/api that encapsulate this logic?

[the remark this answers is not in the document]
bwoebi: 1. I don't really understand which distinction you mean? You mean providing information in producing contexts as part of the distributed headers, so that consumers may decide to create new linked traces or do a distributed trace? This isn't really part of the specification, because I haven't seen the need for it. There's no fundamental reason not to have it though. What we have been doing in PHP instead is having the caller create a new linked trace - and propagate this spans information to downstream.
2. By default a downstream application should make the decision to do a distributed trace, for now at least. If you have multiple parents/sources, span links are the only realistic option. The RFC talks about this, but ultimately, these may need to be judgment calls what's easiest to use. Eventually we'll move to span links first, when we get better UI and UX for span links.
3. I think this is fully sufficient.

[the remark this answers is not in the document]
mabdinur: Awesome @bwoebi. This really clears things up for me. Thanks for the detailed response. I don't think we need to make any changes here 😄

tests/parametric/test_otel_span_methods.py
PerfectSlayer: I find the `find_span()` method more explicit. Can we rather fix it? 
It is way more explicit for test source reader to understand to expected behavior rather than relying on an implicit order.

tests/parametric/test_otel_span_methods.py
PerfectSlayer: I just tried and I only have a trace with 1 span instead of 2

cbeauchesne (changes requested):
Failure in CI on the added methods

cbeauchesne:
I'm converting to draft (it help us to triage our reviews). You can convert it back when the PR is ready.

tests/parametric/test_otel_span_methods.py
mabdinur: Yup we should fix it. We can leave this test as is.

tests/parametric/test_otel_span_methods.py
```diff
@@ -398,6 +400,204 @@ def test_otel_set_attributes_separately(self, test_agent, test_library):
...
+        span_links = retrieve_span_links(span)
+        assert span_links is not None
+        assert len(span_links) == 1
+
+        link = span_links[0]
+        assert link.get("span_id") == 9876543210
+        assert link.get("trace_id") == 1234567890
+        assert link.get("trace_id_high") == 16
+
+        # TODO Tracestate is not part of the extracted headers.
```
mabdinur, line 479: tracestate is not required but if it is valid it should be present

tests/parametric/test_otel_span_methods.py
```diff
@@ -398,6 +400,204 @@ def test_otel_set_attributes_separately(self, test_agent, test_library):
...
+        # tracestateDD = tracestateArr[0][3:].split(";")
+        # assert len(tracestateDD) == 3
+        # assert "o:synthetics" in tracestateDD
+        # assert "s:2" in tracestateDD
+        # assert "t.dm:-4" in tracestateDD
+
+        # TODO Why does flags is expected to be 0, meaning "not set"?
+        # If 0, it is not sampled so it won't be send to the agent.
+        # Moreover, sampling priority is set to 2 (KEEP) and decision marker to -4 (MANUAL).
+        # assert (link.get("flags") or 0) == 0
```
mabdinur, line 494:
```suggestion
        assert link.get("flags", 0) == 1 | -2147483648
```

tests/parametric/test_otel_span_methods.py
```diff
@@ -398,6 +400,204 @@ def test_otel_set_attributes_separately(self, test_agent, test_library):
...
+        assert span_links is not None
+        assert len(span_links) == 1
+
+        link = span_links[0]
+        assert link.get("span_id") == 1311768467284833366
+        assert link.get("trace_id") == 8687463697196027922
+        assert link.get("trace_id_high") == 1311768467284833366
+
+        assert link.get("tracestate") is not None
+        print("tracestate is '%s'" % link["tracestate"])
```
mabdinur, line 531: nit: we can remove this print statement

tests/parametric/test_otel_span_methods.py
```diff
@@ -398,6 +400,204 @@ def test_otel_set_attributes_separately(self, test_agent, test_library):
...
+        span_links = retrieve_span_links(second)
+        assert span_links is not None
+        assert len(span_links) == 2
+
+        link = span_links[0]
+        assert link.get("span_id") == root.get("span_id")
+        assert link.get("trace_id") == root.get("trace_id")
+        assert link.get("trace_id_high") == int(root_tid, 16)
+        assert len(link.get("attributes")) == 0
+        assert link.get("tracestate") == ""
```
mabdinur, line 588: In python a span will always have at least two values in the tracestate (sampling decision and sampling priority). This tracestate value comes from the parent span's context ([python implementation](https://github.com/DataDog/dd-trace-py/blob/28b75850a72f670e5edc2bfa6c3935a7e2d30759/ddtrace/span.py#L546)). This is similar to the opentelemetry implementation. 

Should we adjust the the tracestate assertion to check the value if it is set? Should we make this behavior consistent across languages? 

```suggestion
        if link.get("tracestate"):
            assert link.get("tracestate") == "dd=s:1;t.dm:-0"
```

mabdinur (commented):
Left three suggestions. Aside from that LGTM

tests/parametric/test_otel_span_methods.py
PerfectSlayer: > Should we adjust the the tracestate assertion to check the value if it is set? 

For now, I keep it consistent with the original test: https://github.com/DataDog/system-tests/pull/1811/files#diff-e63af22f5cd81bf600c30dfb393aef024e358dd336ede3786d147f929cab2b54R181
If we do any change, I think we should do it in a dedicated PR for both OTel and DD API.

> Should we make this behavior consistent across languages?

I don't think we need. It is costly to generate a `tracestate` and might not meet the performance requirements for the Java tracer.
We do need to state the benefit of such data before requiring to support it in the RFC.

PerfectSlayer:
@cbeauchesne Ready to be merged 🟢

tests/parametric/test_otel_span_methods.py
```diff
@@ -582,3 +790,25 @@ def run_otel_span_reserved_attributes_overrides_analytics_event(
...
+        links = []
+        for json_link in json_links:
+            link = {}
+            link["trace_id"] = int(json_link["trace_id"][-16:], base=16)
+            link["trace_id_high"] = int(json_link["trace_id"][:16], base=16) if len(json_link["trace_id"]) > 16 else 0
+            link["span_id"] = int(json_link["span_id"], base=16)
+            link["attributes"] = json_link.get("attributes") or {}
+            link["tracestate"] = json_link.get("tracestate") or ""
+            # If set, the high bit (bit 31) should be set according the RFC
+            link["flags"] = 0 if json_link.get("flags") is None else (json_link.get("flags") | -2147483648)
```
mabdinur, line 810: Naive question: Traceflags are encoded as unsigned 32bit integers. Why are using `-1 << 31` to set the high bit. Shouldn't we use `1 << 31`?

mabdinur (approved):

robertomonteromiguel (approved):

tests/parametric/test_otel_span_methods.py
PerfectSlayer: Not so naive question... That is something I borrowed from @bwoebi because I don't use it when encoded through tags. 
You can find here: https://github.com/DataDog/system-tests/pull/1811/files#diff-e63af22f5cd81bf600c30dfb393aef024e358dd336ede3786d147f929cab2b54R145

@bwoebi Can you provide guidance here?

mabdinur: `1 | ( -1 <<31)` has the binary representation of `-0b1111111111111111111111111111111` (instead of 
`0b10000000000000000000000000000001`)

I think using `-1 << 31` to set the highest order bit is incorrect. We should use `1 << 31`. 

@bwoebi Is the `flags` field set in the php parametric app? We could be consistently hitting the `0 if json_link.get("flags")`  case. This would return the expected value in these tests:   https://github.com/DataDog/system-tests/pull/1811/files#diff-e63af22f5cd81bf600c30dfb393aef024e358dd336ede3786d147f929cab2b54R145

PerfectSlayer: I created a dedicated constant so we can fix it later for any test once you find the right value to use:
```python
from utils.parametric.spec.tracecontext import TRACECONTEXT_FLAGS_SET
```

PR #1911, opened 2023-12-11
improve compatibility with `colima` as a Docker driver on macOS

RomainMuller:
## Motivation

When using `colima`[^1] instead of Docker Desktop on macOS, there may not be a `/var/run/docker.sock` file as `colima` uses Docker Contexts.

The correct endpoint to use can be found by using `docker context` subcommands, such as:
```console
$ docker context show # Shows the name of the currently active docker context
colima
$ docker context inspect colima -f '{{ .Endpoints.docker.Host }}' # Extract its endpoint
unix:///Users/romain.marcadier/.colima/default/docker.sock
```

[^1]: https://github.com/abiosoft/colima

## Changes

This PR changes two specific behaviors:
1. The `_get_client()` function in `containers.py` will try to resolve an endpoint from `docker context` if the default `docker.DockerClient.from_env()` raises an exception (re-raising the original error if anything goes wrong while doing so);
2. The `_Weblog` class constructor sets `self.domain` to `localhost` if the value in `$DOCKER_HOST` isn't an `ssh://`, `tcp://`, or `fd://` URI, since the `docker context` commands tend to use `unix://` URIs.

## Workflow

## Reviewer checklist


run.sh
```diff
@@ -333,6 +333,22 @@ function main() {
         run_mode='docker'
     else
         run_mode='direct'
+
+        # if $DOCKER_HOST is not set, check if a non-default docker context is
+        # in use, and set DOCKER_HOST to its endpoint, so that the runner can
+        # appropriately use it without having to foray into the `docker context`
+        # stuff.
+        if [[ "${DOCKER_HOST:-}" == "" ]]; then
```
cbeauchesne, line 341: This change troubles me : It adds two command on the happy path of system tests, making the flow on `run.sh` different from direct call to `pytest`. And it also adds two more reason for a failure for "standard install" (ok, actually, one thanks to the defaulting).

If your need is only specifying `DOCKER_HOST`, what about simply defining it in `.env` in the root of your system test folder ?

RomainMuller: The thing that bothers me with that particular approach is that it implies every single user (of `colima` & otherwise of Docker Contexts) needs to know they have to do this, and it feels awkward (to me) to document user actions that can be easily automatically done.

>making the flow on `run.sh` different from direct call to `pytest`.

It didn't quite strike me as obvious that running `run.sh` is _the same flow_ as calling `pytest`. It seems like there is quite a bit happening in this shell script already... I can however understand the reluctance to add a new dependency in the hot path (`docker`).

---

Maybe we can split the pear in two... The constructor fix is necessary anyway... I could also see if it'd be possible to catch the exception happening when the "default" `DOCKER_HOST` fails, and wrap it into another exception that hints that perhaps a `.env` file needs creating w/ the appropriate `DOCKER_HOST` value?

cbeauchesne: > It didn't quite strike me as obvious that running run.sh is the same flow as calling pytest

Yes, there are lot of code, unfortunately.

Basically, `run.sh` do **four** official features : 

1. if the first argument matches `[AZ_]+`, then consider it's a scenario name, and add the `-S` option
2. if the scenario name is a scenario set name, iterate overs the scenario in this scenario set
3. handling the number process that run on parametric scenario. But it should be removed as soon as we find the time to fix the different CI that use it.
4. And last "hack": removing warnings, because we needed to use a third party deps that pollutes stdout.

Appart form that, any call to `run.sh` with `-` options must be equivalent of `pytest` call

Everything else is included as a best effort, but with no guarantee. Basically, we try to make the framework stable for the most common use case. But we don't have the capacity of handling every particularity. It does not mean we cannot accept changes that help particular uses cases, but we need to keep the happy path totally clean.

> Maybe we can split the pear in two... The constructor fix is necessary anyway... I could also see if it'd be possible to catch the exception happening when the "default" DOCKER_HOST fails, and wrap it into another exception that hints that perhaps a .env file needs creating w/ the appropriate DOCKER_HOST value?

The change you made in the constuctor is not in the happy path, feel free to change anything here  :)  The `.env` file is already loaded in `os.environ`. In short, you can do what you need outside from the happy path, (just no guarantee on our side).

cbeauchesne (approved):

PR #1922, opened 2023-12-13
[OTel] Test `analytics.event` override with respect to OTLP Code

PROFeNoM:
## Motivation

The [RFC](https://docs.google.com/document/d/1KB6XQxZDUgSSQHp1TEk_0X6zgOzmB9-r8NKE9TVaRJw/edit#bookmark=id.44nna34oqd3j) states that the reserved span attributes must match the behavior of the [OTLP code](https://github.com/DataDog/datadog-agent/blob/ea1911d5618df80c1dd25b362369a2e0491910be/pkg/trace/api/otlp.go#L466-L483) which uses Go's [strconv.ParseBool](https://cs.opensource.google/go/go/+/refs/tags/go1.21.5:src/strconv/atob.go;l=10).
```
// ParseBool returns the boolean value represented by the string.
// It accepts 1, t, T, TRUE, true, True, 0, f, F, FALSE, false, False.
```

More specifically, this means that 1, t, T, 0, f, and F strings are also valid entries, while anything other than those listed above shouldn't set the `_dd1.sr.eausr` in the metrics. This PR tests this behavior.

## Workflow

## Reviewer checklist

### DIFF
--- a/tests/parametric/test_otel_span_methods.py
+++ b/tests/parametric/test_otel_span_methods.py
@@ -483,7 +483,7 @@ class Test_Otel_Span_Methods:
         assert "span.type" not in span["meta"]
         assert "analytics.event" not in span["meta"]
 
-    @missing_feature(context.library == "java", reason="OTLP behavior not respected")
+    @missing_feature(context.library == "java", reason="Choose to not implement Go parsing logic")
     @missing_feature(context.library == "golang", reason="OTLP behavior not respected")
     @missing_feature(context.library == "ruby", reason="OTLP behavior not respected")
     @missing_feature(context.library == "nodejs", reason="Not implemented")
--- a/tests/parametric/test_otel_span_methods.py
+++ b/tests/parametric/test_otel_span_methods.py
@@ -483,9 +483,10 @@ class Test_Otel_Span_Methods:
         assert "span.type" not in span["meta"]
         assert "analytics.event" not in span["meta"]
 
-    @missing_feature(context.library < "java@1.25.0", reason="Implemented in 1.25.0")
+    @missing_feature(context.library == "java", reason="OTLP behavior not respected")
+    @missing_feature(context.library == "golang", reason="OTLP behavior not respected")
+    @missing_feature(context.library == "ruby", reason="OTLP behavior not respected")
     @missing_feature(context.library == "nodejs", reason="Not implemented")
-    @missing_feature(context.library == "dotnet", reason=".NET dosn't treat 'something-else' as a valid input here.")
     @missing_feature(context.library == "python", reason="Not implemented")
     @missing_feature(context.library == "python_http", reason="Not implemented")
     @pytest.mark.parametrize(
--- a/tests/parametric/test_otel_span_methods.py
+++ b/tests/parametric/test_otel_span_methods.py
@@ -522,16 +522,7 @@ class Test_Otel_Span_Methods:
     @missing_feature(context.library == "python_http", reason="Not implemented")
     @pytest.mark.parametrize(
         "analytics_event_value,expected_metric_value",
-        [
-            ("t", 1),
-            ("T", 1),
-            ("f", 0),
-            ("F", 0),
-            ("1", 1),
-            ("0", 0),
-            ("fAlse", None),
-            ("trUe", None),
-        ],
+        [("t", 1), ("T", 1), ("f", 0), ("F", 0), ("1", 1), ("0", 0), ("fAlse", None), ("trUe", None),],
     )
     def test_otel_span_extended_reserved_attributes_overrides_analytics_event(
         self, analytics_event_value: Union[bool, str], expected_metric_value: Union[int, None], test_agent, test_library
@@ -571,7 +562,7 @@ def run_otel_span_reserved_attributes_overrides_analytics_event(
     assert len(trace) == 1
 
     span = get_span(test_agent)
-    if (expected_metric_value is not None) or (context.library not in ['dotnet']):
+    if (expected_metric_value is not None) or (context.library not in ["dotnet"]):
         assert span["metrics"].get("_dd1.sr.eausr") == (0 if expected_metric_value is None else expected_metric_value)
     else:
         assert "_dd1.sr.eausr" not in span["metrics"]
--- a/tests/parametric/test_otel_span_methods.py
+++ b/tests/parametric/test_otel_span_methods.py
@@ -537,6 +537,7 @@ class Test_Otel_Span_Methods:
             test_agent=test_agent,
         )
 
+
 def run_operation_name_test(expected_operation_name: str, span_kind: int, attributes: dict, test_library, test_agent):
     with test_library:
         with test_library.otel_start_span("otel_span_name", span_kind=span_kind, attributes=attributes) as span:
--- a/tests/parametric/test_otel_span_methods.py
+++ b/tests/parametric/test_otel_span_methods.py
@@ -507,15 +507,15 @@ class Test_Otel_Span_Methods:
 
     @irrelevant(
         context.library == "java",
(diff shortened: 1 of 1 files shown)


PerfectSlayer (commented):
I left a comment on Slack about it :)

cbeauchesne (changes requested):
CI is failing

cbeauchesne (approved):
So far, ok for me. But I suggest getting a review from another eyes familiar with the RFC to validate the test logic.

PROFeNoM:
Added apm Go/Java/Ruby/Dotnet teams as reviewers since they will be directly impacted by this change. :+1:

tests/parametric/test_otel_span_methods.py
```diff
@@ -483,9 +483,10 @@ def test_otel_span_reserved_attributes_overrides(self, test_agent, test_library)
         assert "span.type" not in span["meta"]
         assert "analytics.event" not in span["meta"]
 
-    @missing_feature(context.library < "java@1.25.0", reason="Implemented in 1.25.0")
+    @missing_feature(context.library == "java", reason="OTLP behavior not respected")
```
PerfectSlayer, line 486: This is not related to OTLP. We choose to not follow the golang boolean string parsing convention.

An honest reason would be "choose to not implement go parsing logic".

PROFeNoM: Ok, will change this :+1:

PerfectSlayer: What about making a second test case with the extended parameters? So we can keep testing the expected behavior on the usual boolean (including using Java) and have a test covering the extended values.
As the tests is parametrize, it should be easy.

PROFeNoM: Makes sense. I'd obviously prefer that we all have the same behavior, but yeah.

PerfectSlayer: In the end, it will still result of having a partial support feature dashboard for Java so it won't be unnoticed. 
Instead, you will just have a more granular evaluation.

bouwkast (commented):
I think that it follows the RFC and OTLP.go better, but I think that it may make sense to ask the overall team whether it makes sense to change the RFC to reduce the overall scope to require the simpler boolean logic that was defined previously and then _optionally_ support the more extensive `strconv.ParseBool` if it makes sense for your language. (I'm assuming that there may be performance concerns with this way?)

The main issue with .NET previously was whether `something-else` should _add_ and _set_ the `_dd1.sr.eausr` tag. From a behavior perspective the lack of the tag or it being set to `0` will be equivalent - so I'm thinking this is more of a semantic question.

Maybe further splitting the tests to define three tests in total:

- One that covers the simpler logic of `"false"`, `"False"`, '"true"` (basically previous test without `"something-else"`
- One that covers _how_ random string data is handled (e.g., `"something-else"` either setting the tag to `0`  or not setting the tag at all
- One that covers the more extended parsing logic as it is currently

Maybe this is taking it too far for this?

tests/parametric/test_otel_span_methods.py
```diff
@@ -541,5 +563,8 @@ def run_otel_span_reserved_attributes_overrides_analytics_event(
     assert len(trace) == 1
 
     span = get_span(test_agent)
-    assert span["metrics"].get("_dd1.sr.eausr") == expected_metric_value
+    if (expected_metric_value is not None) or (context.library not in ["dotnet"]):
```
ajgajg1134, line 566: Hmm why are we checking for library here and not using the `missing_feature` decorator?

tests/parametric/test_otel_span_methods.py
marcotc: >We choose to not follow the golang boolean string parsing convention.

Same for Ruby, and I see that you have already updated the skip reason description @PROFeNoM so thank you!

I also don't foresee us making this test pass for the provided values in Ruby, given that:
* This specific boolean parsing is not consistent with other boolean environment variables at Datadog
* The full list of values that `strconv#ParseBool` considers true and false look pretty foreign to the Ruby environment.

tests/parametric/test_otel_span_methods.py
```diff
@@ -484,29 +484,63 @@ def test_otel_span_reserved_attributes_overrides(self, test_agent, test_library)
...
+            for inputs that aren't accepted by strconv.ParseBool
+        """
+        run_otel_span_reserved_attributes_overrides_analytics_event(
+            analytics_event_value=analytics_event_value,
+            expected_metric_value=expected_metric_value,
+            test_library=test_library,
+            test_agent=test_agent,
+        )
+
+    @missing_feature(context.library == "java", reason="Choose to not implement Go parsing logic")
```
cbeauchesne, line 531: If the reason is really an educated choice, then you should decorate the test with `@irrelevant`, so it won't show up as not. implemented in the feature parity dashbaord.

PROFeNoM: Thanks, I didn't know! Changed it 😃

tests/parametric/test_otel_span_methods.py
```diff
@@ -484,29 +484,63 @@ def test_otel_span_reserved_attributes_overrides(self, test_agent, test_library)
...
+            Tests the analytics.event reserved attribute override with basic inputs
+        """
+        run_otel_span_reserved_attributes_overrides_analytics_event(
+            analytics_event_value=analytics_event_value,
+            expected_metric_value=expected_metric_value,
+            test_library=test_library,
+            test_agent=test_agent,
+        )
+
+    @irrelevant(context.library == "java", reason="The expected metric value is zero")
```
cbeauchesne, line 508: For this one, the reason does not explain very well why it's irrelevant : 

* does it mean the test is not good because *The expected metric value is zero* -> so we need to update the test
* does it mean that those tracer decided to set the metric to zero -> the reason must make claer it's an educated choice
* the expected metric is really zero, but those tracer are failing ? -> it must be decorated with `@bug`
* other interpretation I may not see ?

cbeauchesne (changes requested):

tests/parametric/test_otel_span_methods.py
PROFeNoM: With respect to the RFC, it would be the second one. Here is what comes to my mind:
- `Choose to set the metric value to zero for parseBool-rejected values` 
- `Chose zero metric for parseBool-rejected inputs`
- `Set metric to zero when rejected by parseBool`

cbeauchesne: WDYT  about a message making  crystal clear the fact it's a design choice ? Something like  `Java tracer decided to ...`

PROFeNoM: Alright, I changed the message to `<language> tracer decided to always set _dd1.sr.eausr: 1 for truthy analytics.event inputs, else 0", which imo, clearly says what's going on, although it could be argued that "truthy" is vague (and it is).

cbeauchesne (approved):
AGTM

tests/parametric/test_otel_span_methods.py
```diff
@@ -484,29 +484,72 @@ def test_otel_span_reserved_attributes_overrides(self, test_agent, test_library)
         assert "analytics.event" not in span["meta"]
 
     @missing_feature(context.library < "java@1.25.0", reason="Implemented in 1.25.0")
-    @missing_feature(context.library == "nodejs", reason="Not implemented")
     @missing_feature(context.library == "dotnet", reason=".NET dosn't treat 'something-else' as a valid input here.")
```
bouwkast, line 487: This can be un-skipped for .NET as `something-else` is no longer in this test and we pass all of them.

bouwkast:
```suggestion
```

PR #1949, opened 2023-12-20
Add NodeJS to the Crossed Tracer integration tests

tabgok:
## Motivation

This request expands the crossed-tracer integration tests to include NodeJS

## Changes

This change adds the 'produce' and 'consume' endpoints to the express4 integration in support of the crossed tracer integration tests.  These endpoints are used to produce/consume from Kafka queues.

## Reviewer checklist

* [x] A docker base image is modified?
    * [x] the relevant `build-XXX-image` label is present
* [x] A scenario is added (or removed)?
    * [x] Get a review from [R&P team](https://dd.enterprise.slack.com/archives/C025TJ4RZ8X)
    * [x] Once merged, add (or remove) it in system-test-dasboard nightly

### DIFF
--- a/tests/integrations/test_crossed_integrations.py
+++ b/tests/integrations/test_crossed_integrations.py
@@ -24,14 +24,13 @@ class _PythonBuddy(_Weblog):
 @coverage.basic
 @features.kafkaspan_creationcontext_propagation_with_dd_trace_py
 class Test_PythonKafka:
-    """ Test kafka compatibility with datadog python tracer """
+    """Test kafka compatibility with datadog python tracer"""
 
     WEBLOG_TO_BUDDY_TOPIC = "Test_PythonKafka_weblog_to_buddy"
     BUDDY_TO_WEBLOG_TOPIC = "Test_PythonKafka_buddy_to_weblog"
 
     @classmethod
     def get_span(cls, interface, span_kind, topic):
-
         logger.debug(f"Trying to find traces with span kind: {span_kind} and topic: {topic} in {interface}")
 
         for data, trace in interface.get_traces():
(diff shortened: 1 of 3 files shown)


robertomonteromiguel:
Added the label to run the CROSSED_TRACING_LIBRARIES

utils/build/docker/nodejs/express4/app.js
```diff
@@ -182,6 +182,82 @@ app.get('/dsm', (req, res) => {
...
+  const producer = kafka.producer()
+  const doKafkaOperations = async () => {
+    await admin.connect()
+    await producer.connect()
+    await admin.createTopics({
+      waitForLeaders: true,
+      topics: [
+        { topic }
+      ]
+    })
```
vandonr, line 205: do we actually need to do that ? I think just sending a message to a non-existent topic creates it. Not sure if it's a feature of kafka, or of the frameworks.
(Also, won't it fail if that endpoint is called twice ?)

tests/integrations/test_crossed_integrations.py
```diff
@@ -87,6 +86,7 @@ def test_produce(self):
 
     @missing_feature(library="python")
     @missing_feature(library="java")
+    @missing_feature(library="nodejs")
```
robertomonteromiguel, line 89: It would be good to add the reason.
Even the version (if possible)

robertomonteromiguel (approved):

utils/build/docker/nodejs/express4/app.js
tabgok: The topic _is_ created twice, with the second attempt generating a WARNING ("topic already exists").  However, without waiting for leaders the test was flaky because the client attempted to publish to the topic before the Leader was elected, generating an ERROR ("topic does not have a leader").  

This line causes gets us the WARNING to replace the ERROR, which means no more flaky test

tests/integrations/test_crossed_integrations.py
vandonr: the reason is that python does not propagate context in a normal way, so no one else can hook up with it :/

tabgok: Added the "reason" field

vandonr (approved):
lessgo

cbeauchesne:
@tabgok good old repo here, you can merge the old way :)

PR #2016, opened 2024-01-12
parametric: Updates the interface for creating span links in parametric apps

mabdinur:
## Motivation

- Updates the interface for creating a span with span links 
- Updates the interface for adding links to spans [after span creation]
- Enables 4/6 span link tests for the python tracer
- Adds tests to validate `v0.4` and `v0.5` span links encoding 
- Decouple SpanLink object from distributing tracing headers.

Note: This PR does not introduce span link support for parametric apps that use GRPC. GRPC apps are broken and will need to be removed/fixed.

Note: This PR does not address issues with generating span links from distributed traces. This will be addressed in a future PR.

## Workflow

## Reviewer checklist


tests/parametric/test_span_links.py
```diff
@@ -15,56 +15,68 @@
...
-
-        traces = test_agent.wait_for_num_traces(1)
-        trace = find_trace_by_root(traces, Span(name="root"))
-        assert len(trace) == 2
+            with test_library.start_span("first") as first:
+                pass
 
-        root = root_span(trace)
-        span = find_span(trace, Span(name="child"))
+            first_span_link = Link(
```
PerfectSlayer, line 46: This change break the test.
This test was supposed to create a span links between to span instances.
Now, you only test the span link addition. The span link is created within the test, not the client app.

tests/parametric/test_span_links.py
```diff
@@ -73,20 +85,27 @@ def test_span_link_from_distributed_datadog_headers(self, test_agent, test_libra
...
+                    ],
+                ):
+                    pass
+
+                with test_library.start_span("second") as second:
+                    pass
+
+        traces = test_agent.wait_for_num_traces(1)
+
+        span = traces[0][2]
```
PerfectSlayer, line 108: I don' understand this test. The original test ask to create a span links from HTTP headers (using the original way, it has changed to use the headers field of the link object).

Here, you create a first span.
Then a second span from http headers,
Finally, a last span (called second). This one is supposed to have a span link? And used the second created span headers? Why is that?

robertomonteromiguel (changes requested):
are the python tests failures related with these changes?

tests/parametric/test_span_links.py
```diff
@@ -123,15 +142,19 @@ def test_span_link_from_distributed_w3c_headers(self, test_agent, test_library):
...
             ):
                 pass
 
-        span = get_span(test_agent)
+            with test_library.start_span("span"):
+                pass
+
+        traces = test_agent.wait_for_num_traces(2)
+        root = traces[0][0]
+        span = traces[1][0]
```
PerfectSlayer, line 157: Same question here: Why the second span is supposed to have link?

tests/parametric/test_span_links.py
```diff
@@ -161,33 +184,36 @@ def test_span_link_from_distributed_w3c_headers(self, test_agent, test_library):
...
+                second.add_link(
+                    trace_id=first.trace_id,
+                    span_id=first.span_id,
+                    attributes={"bools": [True, False], "nested": [1, 2]},
+                )
 
-        root = root_span(trace)
-        root_tid = root["meta"].get("_dd.p.tid") or "0" if "meta" in root else "0"
+        traces = test_agent.wait_for_num_traces(2)
+        print(f"ggg {traces}")
```
PerfectSlayer, line 206: `print` command to clean up

tests/parametric/test_span_links.py
```diff
@@ -203,8 +229,8 @@ def test_span_with_attached_links(self, test_agent, test_library):
         assert link.get("trace_id") == first.get("trace_id")
         assert (link.get("trace_id_high") or 0) == int(root_tid, 16)
         assert len(link.get("attributes")) == 4
-        assert link["attributes"].get("bools.0") == "true"
-        assert link["attributes"].get("bools.1") == "false"
+        assert link["attributes"].get("bools.0") == "True"
+        assert link["attributes"].get("bools.1") == "False"
```
PerfectSlayer, line 233: I don't think we should do that. We already had this discussion with Bob I think

PerfectSlayer (changes requested):

PerfectSlayer:
> Note: This PR does not introduce span link support for parametric apps that use GRPC. GRPC apps are broken and will need to be removed/fixed.

I only review the test cases here. Waiting for the PR to support gRPC to review the model / logic change.

tests/parametric/test_span_links.py
```diff
@@ -161,33 +184,36 @@ def test_span_link_from_distributed_w3c_headers(self, test_agent, test_library):
...
-        traces = test_agent.wait_for_num_traces(1)
-        trace = find_trace_by_root(traces, Span(name="root"))
-        assert len(trace) == 3
+            with test_library.start_span("second") as second:
+                second.add_link(trace_id=parent.trace_id, span_id=parent.span_id)
+                second.add_link(
+                    trace_id=first.trace_id,
+                    span_id=first.span_id,
+                    attributes={"bools": [True, False], "nested": [1, 2]},
+                )
```
bwoebi, line 203: add_link intentionally has no trace_id. Users of the API are already expected by other functions to have a mapping <span-id, span> in their test suite. The test is meant to create a link from a given span to another. This may be achieved by a dedicated API, by creating a context from the span, by whatever the preferred method in your language is, but the test itself doesn't directly require a capability "create span link with this span id and trace id", but "create span link to this span".

tests/parametric/test_span_links.py
mabdinur: Actually this test still creates a link between two spans in the client app. The main change here is how the link is sent to the client app. Instead of just sending a parent_id and attributes to the client app this test sends all the fields that can be used to create a span link. This makes the test easier to read/maintain.

This test aims to verify that span links are serialized in the expected format. I think that's the key element we need to test.

tests/parametric/test_span_links.py
mabdinur: The naming is confusing. I should have named the spans: "parent", "linked_span", "child". The goal here is to create an integration test for how span links would be set in a normal application. This is the flow I am trying to recreate: 
 - Span 1 is open (root/parent) -> distributed tracing headers containing span 2 are received (first/linked_span) -> Span 3 is created (second/child). Span 3 is a child of Span 1. -> Span 1 and Span 3 are ended

In this flow Span 3 is a child of Span 1 and is LINKED to Span 2. With this flow span links are never set by the test case. It is set in the application. I can refactor this test to make these relationships more clear. I am noticing a bug that is making this test hard to follow. Does this flow make sense? Is this what we should be doing?

tests/parametric/test_span_links.py
mabdinur: I will update this test to make it more clear

tests/parametric/test_span_links.py
mabdinur: Yes this is ideal but we could run into this scenario: 
Span A -> is created, finished, and flushed
Span B -> is created and attempts to link to Span A
  - Span B can not reference span A in a link because it it no longer stored in the client app.

The only way to guarantee this does not happen is to ensure Span A and Span B are both active. In some client apps this means that Span A will always be the parent of Span B. 

In my refactor I was trying to create a scenarios where Span A and Span B can be started/finished independently and the information about the trace_id is not lost. 

Is this a valid concern? If this is not a valid concern, then we can just close this PR

mabdinur:
> > Note: This PR does not introduce span link support for parametric apps that use GRPC. GRPC apps are broken and will need to be removed/fixed.
> 
> I only review the test cases here. Waiting for the PR to support gRPC to review the model / logic change.

I will hold off on adding GRPC support until we can align on whether this change is even necessary. It is likely that we can close this PR and move forward with the current implementation.

tests/parametric/test_span_links.py
mabdinur: Ahhh interesting. So we are no longer flattening attributes in span links. We need to update the RFC: https://docs.google.com/document/d/1RcTjO8ER5FNweJ0rqVcQEbf6tb1PrlPsQ6coZZww8Xg/edit?disco=AAAA_TaDBkU. 

@bwoebi  The functionality was implemented in the python tracer last week. Should we revert this or should we support both approaches?

mabdinur:
> are the python tests failures related with these changes?

Yes. The python tests are expected to pass once 2.5.0 is released (probably later this week)

tests/parametric/test_span_links.py
bwoebi: No it isn't, it's just relevant for the purposes of testing - in a real application, the user would actually hold a reference to a span or spancontext keeping the span(context) in memory even after it's flushed.

tests/parametric/test_span_links.py
bwoebi: I.e. we are taking a piece from the OTel textbook here, which is mandating a json-compatible representation, and in json it's "true" and not "True".

mabdinur: I opened this fix for the python tracer: https://github.com/DataDog/dd-trace-py/pull/8123. Once this fix is merged the 4/6 system tests should pass

mabdinur:
Closing this refactor PR. We can revisit this at a later date. This does not block any OKR work. 

Thanks for your reviews @bwoebi and @PerfectSlayer 🙏

PR #2028, opened 2024-01-17
Refine shell tracing tests

smola:
## Changes

* Make initial asserts more explicit.
* Assert `resource`.

## Workflow

## Reviewer checklist

### DIFF
--- a/tests/appsec/test_shell_execution.py
+++ b/tests/appsec/test_shell_execution.py
@@ -2,7 +2,7 @@
 # This product includes software developed at Datadog (https://www.datadoghq.com/).
 # Copyright 2021 Datadog, Inc.
 
-from utils import context, coverage, interfaces, weblog, features, irrelevant, rfc
+from utils import bug, context, coverage, interfaces, weblog, features, irrelevant, rfc
 
 
 @rfc("https://docs.google.com/document/d/1YYxOB1nM032H-lgXrVml9mukMhF4eHVIzyK9H_PvrSY/edit#heading=h.o5gstqo08gu5")
--- a/tests/appsec/test_shell_execution.py
+++ b/tests/appsec/test_shell_execution.py
@@ -50,10 +50,10 @@ class Test_ShellExecution:
         )
 
     @irrelevant(library="java", reason="No method for shell execution in Java")
+    @bug(library="nodejs", reason="resource name handling is inconsistent with the RFC")
     def test_track_shell_exec(self):
         span = self.fetch_command_execution_span(self.r_shell_exec)
-        # Some implementations will compute the shell command as "sh -c echo foo".
-        assert span["resource"] in ("echo", "sh")
+        assert span["resource"] == "sh"
         assert span["meta"]["cmd.shell"] == "echo foo"
         assert span["meta"]["cmd.exit_code"] == "0"
 
--- a/tests/appsec/test_shell_execution.py
+++ b/tests/appsec/test_shell_execution.py
@@ -2,11 +2,10 @@
 # This product includes software developed at Datadog (https://www.datadoghq.com/).
 # Copyright 2021 Datadog, Inc.
 
-import json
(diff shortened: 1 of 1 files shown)


robertomonteromiguel (approved):

tests/appsec/test_shell_execution.py
```diff
@@ -14,22 +12,29 @@ class Test_ShellExecution:
...
+        spans = [s for s in spans if s["name"] == "command_execution"]
+        assert spans, "No command_execution span found"
+        assert len(spans) == 1, "More than one command_execution span found"
+
+        command_exec_span = spans[0]
+        assert command_exec_span["name"] == "command_execution"
+        assert command_exec_span["type"] == "system"
+        assert command_exec_span["meta"]["component"] == "subprocess"
+
         assert command_exec_span is not None, f"command execution span hasn't be found for {r.url}"
```
cataphract, line 30: stuff on line 30-32 became duplicative

tests/appsec/test_shell_execution.py
```diff
@@ -38,29 +43,27 @@ def setup_track_cmd_exec(self):
...
-            data=json.dumps({"command": "echo", "options": {"shell": True}, "args": "foo"}),
+            "/shell_execution", json={"command": "echo", "options": {"shell": True}, "args": "foo"},
         )
 
     @irrelevant(library="java", reason="No method for shell execution in Java")
     def test_track_shell_exec(self):
         shell_exec_span = self.fetch_command_execution_span(self.r_shell_exec)
-
+        # Some implementations will compute the shell command as "sh -c echo foo".
+        assert shell_exec_span["resource"] in ("echo", "sh")
```
cataphract, line 59: From my reading of the spec, the only possibilities here:

* resource=sh, cmd.shell=echo foo
* resource=sh, cmd.exec=["sh", "-c", "echo foo"]
* resource=echo, cmd.exec=["echo", "foo"], if the echo binary is executed directly

Since this test is about shell execution, we can exclude the third. Arguably, the second is not appropriate for this test either, but in any case what doesn't make sense is to have resource=echo, cmd.shell=echo foo. The spec says:

> If a shell command is being executed (i.e. cmd.shell is populated), the resource is the name of the shell used (default to sh).

tests/appsec/test_shell_execution.py
```diff
@@ -69,16 +72,15 @@ def setup_truncated(self):
     )
     def test_truncated(self):
         shell_exec_span = self.fetch_command_execution_span(self.r_truncation)
+        assert shell_exec_span["resource"] == "echo"
```
cataphract, line 75: already a problem in the code before, but the local var (shell_exec_span) should be renamed

tests/appsec/test_shell_execution.py
```diff
@@ -87,6 +89,6 @@ def setup_obfuscation(self):
     )
     def test_obfuscation(self):
         shell_exec_span = self.fetch_command_execution_span(self.r_obfuscation)
-
+        assert shell_exec_span["resource"] == "echo"
```
cataphract, line 92: idem

tests/appsec/test_shell_execution.py
smola: That's correct. cc @hoolioh Node's implementation is not consistent with this. Is it ok if I mark this test as `@bug` for it?

tests/appsec/test_shell_execution.py
smola: Changing it to just `span`, I don't think we need that much disambiguation in such a short scope.

tests/appsec/test_shell_execution.py
hoolioh: I'm looking into it. Anyway Shell Execution feature is not merged yet in the nodejs tracer so I think you don't need to annotate the test.

hoolioh (approved):

PR #2100, opened 2024-02-02
Script to check test status changes across tracer versions

smola:
## To Do

* [ ] Add CLI arg to set the weblog variant.
* [ ] Avoid hardcoding the list of tracer versions.
* [ ] Extend support beyond Java.
* [ ] Add bisect-like mode.

### DIFF
--- a/version-check
+++ b/version-check
@@ -1,0 +1,88 @@
+#!/usr/bin/env bash
+set -eu
+
+library="java"
+weblog_variant="ratpack"
+
+readonly OUTPUT_DIR=version-check-output
+mkdir -p "$OUTPUT_DIR"
+
+declare -A SEEN_XPASS
+declare -A SEEN_FAIL
+
+versions=(
+    0.110.0
+    0.111.0
+    0.112.0
+    0.113.0
+    0.114.0
+    0.115.0
+    1.0.0
+    1.1.0
+    1.2.0
+    1.3.0
+    1.4.0
+    1.5.0
+    1.6.0
+    1.7.0
+    1.8.0
+    1.9.0
+    1.10.0
+    1.11.0
+    1.12.0
+    1.13.0
+    1.14.0
+    1.15.0
+    1.16.0
+    1.17.0
+    1.18.0
+    1.19.0
+    1.20.0
+    1.21.0
+    1.22.0
+    1.23.0
+    1.24.0
+    1.25.1
+    1.26.0
+    1.27.0
+    1.28.0
+)
+
+for v in "${versions[@]}"; do
+    output_file="$OUTPUT_DIR/$library-$v-$weblog_variant.txt"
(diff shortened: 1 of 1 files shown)


version-check
```diff
@@ -0,0 +1,88 @@
...
+)
+
+for v in "${versions[@]}"; do
+    output_file="$OUTPUT_DIR/$library-$v-$weblog_variant.txt"
+    if [[ -f "$output_file" ]]; then
+        echo "Skipping version $v (already tested)"
+    else
+        echo "Testing version $v"
+        jar_name="dd-java-agent-$v.jar"
+        if [ ! -f "$jar_name" ]; then
```
cbeauchesne, line 58: Could you include this logic inside `utils/scripts/load-binary.sh`. I never took the time to implement this, but the plan was : 

```bash 
utils/scripts/load-binary.sh java # load last artifacts from main branch of dd-trace-java
utils/scripts/load-binary.sh java dev  # same as above, dev is the default value for the second argument
utils/scripts/load-binary.sh java 1.26.0 # TODO : load 1.26.0 of java tracer
```

With that, we'll have a unique script that handle this loading.

version-check
```diff
@@ -0,0 +1,88 @@
+#!/usr/bin/env bash
```
cbeauchesne, line 1: Could you put this script in `utils/scripts`, and prepare a CLI that will supports all tracers (even if only java is supported).

As now, I see : 

```bash
./utils/script/version-check.sh <tracer_name> <SCENARIO_NAME> [-w WEBLOG_VARIANT]
```

probably a `-n` option to give how versions it should check ?

version-check
```diff
@@ -0,0 +1,88 @@
...
+weblog_variant="ratpack"
+
+readonly OUTPUT_DIR=version-check-output
+mkdir -p "$OUTPUT_DIR"
+
+declare -A SEEN_XPASS
+declare -A SEEN_FAIL
+
+versions=(
+    0.110.0
```
cbeauchesne, line 14: is there a way to get this lit automatically ?

simon-id: I'm guessing yes, but it would have to be a different method for each library, so it's a bit annoying to implement in a single script like that. What if we accepted the list from stdin instead ?

cbeauchesne: > I'm guessing yes, but it would have to be a different method for each library, so it's a bit annoying to implement in a single script like that

it wouldn't be the first time 😄 

> What if we accepted the list from stdin instead ?

That's another possibility (actually compatible with the first solution).

But in a sense, I feel that the hardcoded solution is not the good one : maintenance, and possibly missing version. Actually, I prefer : 

1. @simon-id's  idea (list provided in stdin)
2. nice to have : a script that load that list (giving the liberty to each tracer team to implement it if they feel it's needed).

PR #2108, opened 2024-02-02
Maximo/dotnet parametric service to http

link04:
## Motivation

There is an effort to change the services used form GRPC to HTTP.

## Changes

Changing the `parametric/conftest` file to point to the newly added directory that contains the HTTP service that can be used instead.

## Workflow

## Reviewer checklist


utils/build/docker/dotnet_http/parametric/ApmTestApi.sln.DotSettings.user
zacharycmontoya, line 1: We probably want to remove this file

utils/build/docker/dotnet_http/parametric/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+WORKDIR /app
+
+# ensure that the Datadog.Trace.dlls are installed from /binaries
+COPY install_ddtrace.sh query-versions.fsx binaries* /binaries/
+RUN dos2unix /binaries/install_ddtrace.sh
+RUN /binaries/install_ddtrace.sh
+
+# restore nuget packages
+COPY ["ApmTestApi.csproj", "nuget.config", "*.nupkg", "./"]
+RUN dotnet restore "./././ApmTestApi.csproj"
```
zacharycmontoya, line 16:
```suggestion
RUN dotnet restore "./ApmTestApi.csproj"
```

utils/build/docker/dotnet_http/parametric/ApmTestApi.csproj.user
zacharycmontoya, line 1: I think we can get rid of this file

utils/build/docker/dotnet_http/parametric/ApmTestApi.csproj
```diff
@@ -0,0 +1,18 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+    <Nullable>enable</Nullable>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <InvariantGlobalization>true</InvariantGlobalization>
+    <UserSecretsId>f56b479c-4bfa-4286-bb36-96a6273695e3</UserSecretsId>
+    <Configuration Condition="'$(Configuration)' == ''">Development</Configuration>
```
zacharycmontoya, line 9: I think we should remove this line. The Configuration property tends to be `Debug` or `Release` so we don't need to mess with it, unless you had a specific reason in mind

tests/parametric/conftest.py
```diff
@@ -191,36 +191,39 @@ def golang_library_factory():
...
-COPY utils/build/docker/dotnet/install_ddtrace.sh utils/build/docker/dotnet/query-versions.fsx binaries* /binaries/
+COPY install_ddtrace.sh query-versions.fsx binaries* /binaries/
 RUN dos2unix /binaries/install_ddtrace.sh
 RUN /binaries/install_ddtrace.sh
 
 # restore nuget packages
-COPY ["{dotnet_reldir}/ApmTestClient.csproj", "{dotnet_reldir}/nuget.config", "{dotnet_reldir}/*.nupkg", "./"]
-RUN dotnet restore "./ApmTestClient.csproj"
+COPY ["ApmTestApi.csproj", "nuget.config", "*.nupkg", "./"]
+RUN dotnet restore "./././ApmTestApi.csproj"
```
zacharycmontoya, line 217: Another small nit, even though it shouldn't change the behavior:

```suggestion
RUN dotnet restore "./ApmTestApi.csproj"
```

utils/build/docker/dotnet_http/parametric/ddtracer_version.Dockerfile
zacharycmontoya, line 1: Can we delete this dockerfile? It looks like it may have been used for debugging purposes

utils/build/docker/dotnet_http/parametric/Program.cs
```diff
@@ -0,0 +1,27 @@
+using System.Diagnostics;
+using ApmTestApi.Endpoints;
+
+// Force the initialization of the tracer
+_ = Datadog.Trace.Tracer.Instance;
+
+var builder = WebApplication.CreateBuilder(args);
+
+// Add services to the container.
+builder.Services.AddEndpointsApiExplorer();
```
zacharycmontoya, line 10: It doesn't look like the Api Explorer isn't used in the app, can we remove it? I looked at this blog post conveniently written by Andrew and it doesn't seem like you're injecting the information anywhere: https://andrewlock.net/introduction-to-the-apiexplorer-in-asp-net-core/

bouwkast (commented):
Overall it looks good to me

I would think that with this we could remove `dotnet/parametric` but it doesn't seem to work when I do that locally 🤔 

e.g., it gets an invalid value for the library `dotnet@99999.99999.99999` used
and doesn't run anything `gw0 [0]`

Maybe it has something to do with the `dotnet_http` folder name?

utils/build/docker/dotnet_http/parametric/Endpoints/ApmTestApi.cs
```diff
@@ -0,0 +1,325 @@
...
+    
+    public static async Task StopTracer()
+    {
+        await Tracer.Instance.ForceFlushAsync();
+    }
+
+    private static async Task<string> StartSpan(HttpRequest request)
+    {
+        var headerRequestBody = await new StreamReader(request.Body).ReadToEndAsync();
+        var parsedDictionary = JsonConvert.DeserializeObject<Dictionary<string, Object>>(headerRequestBody);
```
zacharycmontoya, line 88: C# style suggestion: Since the DeserializeObject returns a value that may be null, you could try the following so you always have a non-null value assigned to `parsedDictionary`, to avoid having to use the `!` operator later. This will require you to add a using statement to bring in the `System.Collections.Immutable` namespace. Also I recommend changing `Object` to `object` so you're using the built-in alias type

```suggestion
        var parsedDictionary = JsonConvert.DeserializeObject<Dictionary<string, object>>(headerRequestBody) ?? ImmutableDictionary<string, object>.Empty;
```

utils/build/docker/dotnet_http/parametric/Endpoints/ApmTestApi.cs
```diff
@@ -0,0 +1,325 @@
...
+        parsedDictionary.TryGetValue("value", out var value);
+
+        var span = Spans[Convert.ToUInt64(id)];
+        span.SetTag(key, value);
+    }
+
+    private static async Task SpanSetMetric(HttpRequest request)
+    {
+        var headerBodyDictionary = await new StreamReader(request.Body).ReadToEndAsync();
+        var parsedDictionary = JsonConvert.DeserializeObject<Dictionary<string, object>>(headerBodyDictionary)!;
```
zacharycmontoya, line 164: Since you do this in every method, it might be easier to break this out to a new method and call that. Whether you refactor into a new method or not, you should make sure to dispose the StreamReader once you are done with it. For example, you could do something like:

```csharp
using var streamReader = new StreamReader(request.Body);
var body = await streamReader.ReadToEndAsync();
```

utils/build/docker/dotnet_http/parametric/Endpoints/ApmTestApi.cs
```diff
@@ -0,0 +1,325 @@
...
+        parsedDictionary.TryGetValue("key", out var key);
+        parsedDictionary.TryGetValue("value", out var value);
+
+        var span = Spans[Convert.ToUInt64(id)];
+        SetMetric.Invoke(span, new object[] { key!, Convert.ToDouble(value) });
+    }
+
+    private static async Task SpanSetError(HttpRequest request)
+    {
+        var span = Spans[Convert.ToUInt64(await FindBodyKeyValueAsync(request, "span_id"))];
```
zacharycmontoya, line 175: In this method I would recommend following the similar pattern as other endpoints, where you extract the dictionary first, and then later in the method you just look to see if the key exists in the extracted dictionary. This change would mean you only read the request body once, which is safer and more efficient 😄

utils/build/docker/dotnet_http/parametric/Endpoints/ApmTestApiOtel.cs
```diff
@@ -0,0 +1,333 @@
...
+    {
+        _logger.LogInformation("OtelFlushTraceStats: {Request}", request);
+
+        await FlushTraceStats();
+
+        _logger.LogInformation("OtelFlushTraceStatsReturn");
+    }
+    
+    // Helper methods:
+    private static async Task<Dictionary<string, object>> DeserializeRequestObjectAsync(Stream requestBody)
```
zacharycmontoya, line 288: Ah I didn't see this earlier, this helper method does exactly what I was suggesting in the `ApmTestApi.cs` file 😅 You should definitely share this between the two endpoints files

utils/build/docker/dotnet_http/parametric/Endpoints/ApmTestApiOtel.cs
```diff
@@ -0,0 +1,333 @@
...
+                    kind = ActivityKind.Consumer;
+                    break;
+                default:
+                    kind = ActivityKind.Internal; // this is the default in Activity
+                    break;
+            }
+        }
+        
+        var activity = ApmTestApiActivitySource.StartActivity(
+            (string)requestBodyObject["name"],
```
zacharycmontoya, line 114: Super small nit: Although we always expect the incoming value to be a string, I suggest casting the object using the `as` operator. This will do the same thing as the hard `(string)` cast except if the object is not of type `string` then the `as` operator will return `null` whereas the cast would throw an `InvalidCastException`.

```suggestion
            requestBodyObject["name"] as string,
```

zacharycmontoya (approved):
LGTM and the tests are passing! 🎉  I left a number of comments about refactoring and C# best practices that I hope are helpful, so it would be great to see those incorporated before merging but they're not blocking. Thank you for working on this!

tests/parametric/conftest.py
```diff
@@ -191,36 +191,39 @@ def golang_library_factory():
...
-FROM mcr.microsoft.com/dotnet/sdk:7.0
-RUN apt-get update && apt-get install dos2unix
-WORKDIR /app
+FROM mcr.microsoft.com/dotnet/sdk:8.0
 
-# Opt-out of .NET SDK CLI telemetry (prevent unexpected http client spans)
-ENV DOTNET_CLI_TELEMETRY_OPTOUT=1
+RUN apt-get update && apt-get install -y dos2unix --no-install-recommends
+RUN apt-get update && apt-get install -y curl
+USER root
```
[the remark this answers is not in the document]
zacharycmontoya: +1 we should see if it works now after removing `USER root`

utils/build/docker/dotnet_http/parametric/.dockerignore
```diff
@@ -0,0 +1,30 @@
+**/.classpath
```
lucaspimentel, line 1: Where did this come from? Why not copy the previous one from https://github.com/DataDog/system-tests/blob/main/utils/build/docker/dotnet/parametric/.dockerignore ?

utils/build/docker/dotnet_http/parametric/ApmTestApi.csproj
```diff
@@ -0,0 +1,18 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
+    <PackageReference Include="Datadog.Trace" Version="*" Condition="'$(Configuration)' != 'Release'" />
```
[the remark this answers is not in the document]
lucaspimentel: I like the idea here, but I don't think we should tie this to the `Configuration`. What if I want to do a `Release` build locally? (Isn't that the default nowadays anyway?)

Instead, can we use a new msbuild property with a sensible default that we can override? Something like

```xml
<PackageReference Include="Datadog.Trace" Version="*" Condition="'$(USE_DATADOG_TRACE_NUGET)' == 'true'" />
```

utils/build/docker/dotnet_http/parametric/ApmTestApi.csproj.user
lucaspimentel: We should have a `.gitignore` file with `*.user`, like this one:
https://github.com/DataDog/system-tests/blob/main/utils/build/docker/dotnet/parametric/.gitignore

utils/build/docker/dotnet_http/parametric/ApmTestApi.csproj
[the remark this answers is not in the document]
bouwkast: To clarify, this is useful when making changes to the project and getting Intellisense to work and building locally
After you do that though, you have to delete the the `bin`/`obj` directories (even though they are excluded in the `.dockerignore` not really sure why)

We could do something like that but the reason to include the `PackageReference` is just to get Intellisense support, so it's not really to "use" it as an actual NuGet

I'd be fine making it a bit more explicit just so it doesn't accidentally get included though with something like that though

utils/build/docker/dotnet_http/parametric/Endpoints/ApmTestApi.cs
```diff
@@ -0,0 +1,325 @@
...
+using Newtonsoft.Json;
+
+namespace ApmTestApi.Endpoints;
+
+public abstract class ApmTestApi
+{
+    public static void MapApmTraceEndpoints(WebApplication app, ILogger<ApmTestApi> logger)
+    {
+        _logger = logger;
+        _ = Tracer.Instance;
```
bouwkast, line 14: I was wondering why this was here and there is a comment on the other project that I think should be copied over

```
            // TODO: Remove when the Tracer sets the correct results in the SpanContextPropagator.Instance getter
            // This should avoid a bug in the SpanContextPropagator.Instance getter where it is populated WITHOUT consulting the TracerSettings.
            // By instantiating the Tracer first, that faulty getter code path will not be invoked
```

utils/build/docker/dotnet_http/parametric/ddtracer_version.Dockerfile
bouwkast: Just copy/pasting the info from the slack

This one isn't being used as there is a check here: https://github.com/DataDog/system-tests/blob/eb1d2bb572227dc89569cd6de1f3013e462fcca1/utils/_context/_scenarios.py#L1014

That will look for the one in `dotnet/parametric` still

bouwkast (approved):
Nice! Looks good to me, just a few small things

utils/build/docker/dotnet_http/parametric/.dockerignore
link04: That one was auto generated, will change it now!

utils/build/docker/dotnet_http/parametric/ApmTestApi.csproj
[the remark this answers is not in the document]
link04: So I didn't get issues with bin/obj with this project as I did with the old one, will need to play around with it to see which option matches better, lets see how it runs after rolling back the new folder change.

utils/build/docker/dotnet/parametric/ddtracer_version.Dockerfile
```diff
@@ -1,11 +1,10 @@
 
-FROM mcr.microsoft.com/dotnet/sdk:6.0
+FROM mcr.microsoft.com/dotnet/sdk:8.0
 
 RUN apt-get update && apt-get install dos2unix
-
 WORKDIR /app
 
-COPY utils/build/docker/dotnet/install_ddtrace.sh utils/build/docker/dotnet/query-versions.fsx binaries* /binaries/
+COPY install_ddtrace.sh query-versions.fsx binaries* /binaries/
```
zacharycmontoya, line 7: The build context used for this Docker build is set to the base folder, so you'll need to include the relative directories like this:

```suggestion
COPY utils/build/docker/dotnet/install_ddtrace.sh utils/build/docker/dotnet/query-versions.fsx binaries* /binaries/
```

utils/build/docker/dotnet/parametric/install_ddtrace.sh
zacharycmontoya, line 1: The mode on this file has been changed from 100755 to 100644 which means that it no longer has execute permissions, so this will cause a Dockerfile build failure when it tries to run this script. To fix this, you'll need to run the following command:

```
git update-index --chmod=+x utils/build/docker/dotnet/parametric/install_ddtrace.sh
```

[Source](https://stackoverflow.com/questions/40978921/how-to-add-chmod-permissions-to-file-in-git)

utils/build/docker/dotnet/parametric/.gitignore
lucaspimentel, line 1: Don't delete this file if you're keeping this project here. Either delete the entire folder or don't delete anything. (We need it to keep git from adding files added by the IDE and when building.)

utils/build/docker/dotnet_http/parametric/ApmTestApi.csproj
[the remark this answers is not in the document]
lucaspimentel: > To clarify, this is useful when...

No, I get it. I think it's a good idea! I also add the nuget package locally when working on parametric tests (well, since it was removed 😉 ), I just never push that change.

My issue is with trying this to `Condition="'$(Configuration)' != 'Release'"`. I should be able to choose to build either debug or release locally for different reasons without breaking this nuget include.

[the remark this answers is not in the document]
bouwkast: Ahh gotcha
Yeah that makes sense, so yeah it'd probably be best to wrap this with `USE_DATADOG_TRACE_NUGET` instead of the build configuration

utils/build/docker/dotnet_http/parametric/Dockerfile
lucaspimentel: Is this file used at all? Can we just delete it?

utils/build/docker/dotnet/parametric/ddtracer_version.Dockerfile
```diff
@@ -1,12 +1,11 @@
 
-FROM mcr.microsoft.com/dotnet/sdk:6.0
+FROM mcr.microsoft.com/dotnet/sdk:8.0
 
 RUN apt-get update && apt-get install dos2unix
-
 WORKDIR /app
 
-COPY utils/build/docker/dotnet/install_ddtrace.sh utils/build/docker/dotnet/query-versions.fsx binaries* /binaries/
+COPY utils/build/docker/dotnet/parametric/install_ddtrace.sh utils/build/docker/dotnet/parametric/query-versions.fsx binaries* /binaries/
```
cbeauchesne, line 7: Is there a way to use the same script? If possible, I'd prefer maintain only one script

utils/build/docker/dotnet/parametric/query-versions.fsx
```diff
@@ -0,0 +1,75 @@
+#r "nuget: Newtonsoft.Json, 13.0.1"
```
cbeauchesne, line 1: This script is exactly the same as `utils/build/docker/dotnet/query-versions.fsx`, why duplicate it ?

cbeauchesne:
Appart from the duplicated scripts, all good to me

utils/build/docker/dotnet/parametric/query-versions.fsx
link04: Makes sense, just removed it from the extra folder and same for the installer one.

zacharycmontoya:
Note: The current dotnet parametric dev run still utilizes the faulty `dd-trace-dotnet:latest_snapshot` tag so I think we should ignore that failure. This should be fixed by the in-flight dd-trace-dotnet PR https://github.com/DataDog/dd-trace-dotnet/pull/5182

zacharycmontoya (approved):
Approving again after the latest code changes 👍🏼

cbeauchesne (approved):

PR #2230, opened 2024-03-14
[Debugger] Add PII Redaction

shurivich:
## Motivation
Implement PII Redaction tests

## Changes
Add PII Redaction tests
Add PII Redaction feature


manifests/nodejs.yml
```diff
@@ -313,6 +313,8 @@ tests/:
       Test_Debugger_Method_Probe_Snaphots: irrelevant
       Test_Debugger_Mix_Log_Probe: irrelevant
       Test_Debugger_Probe_Statuses: irrelevant
+    test_debugger_pii.py:
+      Test_Debugger_PII_Redaction: irrelevant
```
cbeauchesne, line 317: I'm a bit surprised to see `irrelevant` for almost all languages. Should it be `missing_feature` ?

cbeauchesne (approved):

manifests/nodejs.yml
simon-id: Well that should be marked as missing_feature then ? Irrelevant, is when the test is never going to be useful, missing_feature it's when it will be useful someday

shurivich: Opened PR for that here
https://github.com/DataDog/system-tests/pull/2257

PR #2348, opened 2024-04-16
less strict test for Test_Headers

christophe-papazian:
## Motivation

`test_specific_key2` is a very strict test and it should pass for all reasonnable implementation of blocking feature

## Changes

Many frameworks replace underscore with dash to normalize header names. Allow that behaviour in that test too.
Known impacted frameworks:

- Flask (Python)
- Django (Python)
- Rack (Ruby)
- ...

## Workflow

## Reviewer checklist

### DIFF
--- a/tests/appsec/waf/test_addresses.py
+++ b/tests/appsec/waf/test_addresses.py
@@ -112,7 +112,6 @@ class Test_Headers:
                 self.r_sk_4, pattern="routing.yml", address="server.request.headers.no_cookies", key_path=["x-filename"]
             )
 
-
     def setup_specific_key3(self):
         self.r_sk_5 = weblog.get("/waf/", headers={"referer": "<script >"})
         self.r_sk_6 = weblog.get("/waf/", headers={"RefErEr": "<script >"})
--- a/tests/appsec/waf/test_addresses.py
+++ b/tests/appsec/waf/test_addresses.py
@@ -97,16 +97,21 @@ class Test_Headers:
     def setup_specific_key2(self):
         self.r_sk_4 = weblog.get("/waf/", headers={"X_Filename": "routing.yml"})
 
-    @missing_feature(library="python")
     @irrelevant(library="ruby", reason="Rack transforms underscores into dashes")
     @irrelevant(library="php", reason="PHP normalizes into dashes; additionally, matching on keys is not supported")
     @missing_feature(weblog_variant="spring-boot-3-native", reason="GraalVM. Tracing support only")
     def test_specific_key2(self):
         """attacks on specific header X_Filename, and report it"""
+        try:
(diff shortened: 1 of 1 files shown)


tests/appsec/waf/test_addresses.py
```diff
@@ -97,16 +97,20 @@ def test_specific_key(self):
...
-
-        interfaces.library.assert_waf_attack(
-            self.r_sk_4, pattern="routing.yml", address="server.request.headers.no_cookies", key_path=["x_filename"]
-        )
+        try:
+            interfaces.library.assert_waf_attack(
+                self.r_sk_4, pattern="routing.yml", address="server.request.headers.no_cookies", key_path=["x_filename"]
+            )
+        except ValueError:
+            # also accept report on x-filename
```
smola, line 110: Name Python and WSGi here? (if that's the case) That will make it easier to later reason about where the exception comes from.

smola (approved):

tests/appsec/waf/test_addresses.py
christophe-papazian: Ruby seems to be impacted too, and php as well. I'm not sure this is really language specific, any new web framework could decide to do it as well.

tests/appsec/waf/test_addresses.py
```diff
@@ -97,16 +97,20 @@ def test_specific_key(self):
...
         """attacks on specific header X_Filename, and report it"""
-
-        interfaces.library.assert_waf_attack(
-            self.r_sk_4, pattern="routing.yml", address="server.request.headers.no_cookies", key_path=["x_filename"]
-        )
+        try:
+            interfaces.library.assert_waf_attack(
+                self.r_sk_4, pattern="routing.yml", address="server.request.headers.no_cookies", key_path=["x_filename"]
+            )
+        except ValueError:
```
anna-git, line 109: can it be an assert_waf_attack(x_filename) or assert_waf_attack(x-filename) 
instead of a try clause?

anna-git: (maybe a comment saying some frameworks normalize _ to - before it can get to the waf)

christophe-papazian: no because if it fails, it doesn't return False but raise an Exception. So the only way is to catch it.

cbeauchesne: Well, the try/except is indeed not the best. But adding the capacity of supporting several key_path (or a regex) to `assert_waf_attack` will over-complexify it. As it's the only use case, let's go with it. i'll reconsider and do a cleaner way if there are more use case in the futur like this

cbeauchesne (approved):

uurien (approved):

PR #2407, opened 2024-04-29
fix: w3c trace header test

dmehala:
## Workflow

## Reviewer checklist

### DIFF
--- a/tests/parametric/test_headers_tracecontext.py
+++ b/tests/parametric/test_headers_tracecontext.py
@@ -479,8 +479,8 @@ class Test_Headers_Tracecontext:
         expects the tracestate to be discarded
         """
         with test_library:
-            _, tracestate1 = make_single_request_and_get_tracecontext(test_library, [["tracestate", "foo=1"],])
-            _, tracestate2 = make_single_request_and_get_tracecontext(test_library, [["tracestate", "foo=1,bar=2"],])
+            _, tracestate1 = make_single_request_and_get_tracecontext(test_library, [["tracestate", "foo=1"],],)
+            _, tracestate2 = make_single_request_and_get_tracecontext(test_library, [["tracestate", "foo=1,bar=2"],],)
 
         # Updated the test to check that the number of tracestate list-members is the same,
         # since Datadog will add an entry.
@@ -719,7 +719,6 @@ class Test_Headers_Tracecontext:
     @missing_feature(context.library < "php@0.99.0", reason="Not implemented")
     @missing_feature(context.library < "nodejs@5.6.0", reason="Not implemented")
     @missing_feature(context.library == "java", reason="Not implemented")
-    @missing_feature(context.library == "cpp", reason="Not implemented")
     @missing_feature(context.library < "ruby@2.0.0", reason="Not implemented")
     @missing_feature(context.library == "golang", reason="Not implemented")
     def test_tracestate_w3c_p_extract(self, test_agent, test_library):
@@ -774,7 +773,7 @@ class Test_Headers_Tracecontext:
         assert case1["meta"]["_dd.parent_id"] == "0123456789abcdef"
 
         assert case2["name"] == "p_invalid"
-        assert case2["meta"]["_dd.parent_id"] == "XX!X"
+        assert "_dd.parent_id" not in case2["meta"]
 
         assert case3["name"] == "datadog_headers_used_in_propagation"
         assert case3["trace_id"] == 5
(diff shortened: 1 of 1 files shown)


tests/parametric/test_headers_tracecontext.py
```diff
@@ -774,22 +773,21 @@ def test_tracestate_w3c_p_extract(self, test_agent, test_library):
         assert case1["meta"]["_dd.parent_id"] == "0123456789abcdef"
 
         assert case2["name"] == "p_invalid"
-        assert case2["meta"]["_dd.parent_id"] == "XX!X"
+        assert "_dd.parent_id" not in case2["meta"]
```
mabdinur, line 776:
```suggestion
        assert case2["meta"]["_dd.parent_id"] == "0000000000000000"
```

Good catch here. Invalid parent_ids should not be propagated. `_dd.parent_id` should be set to `0000000000000000`.

tests/parametric/test_headers_tracecontext.py
```diff
@@ -774,22 +773,21 @@ def test_tracestate_w3c_p_extract(self, test_agent, test_library):
...
 
         assert case3["name"] == "datadog_headers_used_in_propagation"
         assert case3["trace_id"] == 5
         assert case3["parent_id"] == 11
         assert "_dd.parent_id" not in case3["meta"]
 
         assert case4["name"] == "p_not_propagated"
-        assert case4["meta"]["_dd.parent_id"] == "0000000000000000"
+        assert "_dd.parent_id" not in case4["meta"]
```
mabdinur, line 784: According to the extraction spec `_dd.parent_id` tag should always be set. If p is not propagated or is invalid.
```suggestion
        assert case4["meta"]["_dd.parent_id"] == "0000000000000000"
```

mabdinur (commented):
I will update the ruby, python, and node tracer to better align with the spec. Can we move case2 and case4 to their own system tests and then skip it for all failing languages. This way we can maintain test coverage for case1 and case3 which should be passing for all languages. 

Thanks for catching this

tests/parametric/test_headers_tracecontext.py
dmehala: I might had overlooked the RFC. If `p` is invalid, why `_dd.parent_id` should be set to "0000000000000000"?

bouwkast: Validation of the propagated `p` sub-key on extraction wasn't considered to be important - when it gets to the backend it'll need to be validated there regardless of whether we validated it or not as we don't have exclusive control over the contents of `tracestate`.

But fine with me either way, if validation whenever it gets extracted makes more sense then 👍

dmehala: @bouwkast not sure to understand. In scenario 2, the RFC spec that `_dd.parent_id` should be `0000000000000000`. Do you mean the backend will still have to validate `p`, thus `_dd.parent-id=0000000000000000` or `_dd.parent_id` is equivalent? If yes, then I think not sending `_dd.parent_id` is better resource wise.

dmehala: Thank you @mabdinur.

I think the confusion comes from:

> The RFC also states that when an **_invalid_** `p` value is extracted `_dd.parent_id` should be set to `0000000000000000`.

To know if `p` is invalid, we must run some validation process on `p`. 

My understanding of that _validation_ was:
> - The p sub-key MUST be injected into the dd list-member of the tracestate.
> - The value of p MUST be a 16-character lowercase hexadecimal encoded string.
> - The value of p MUST be set to the ID of the currently active span.
> - The value of p MUST be a non-zero value (0000000000000000 is invalid).
> - A : (colon) character MUST be placed between the key and the value. For example, p:0123456789abcdef

While the rest of the team understanding was (I guess):
> - The initial value MUST be a hexadecimal string 16 characters in length.

In the end, that's a misinterpretation on me. The initial intent is to centralize the validation, I think it's a good thing idea but at the expense of sending unnecessary data in some cases. 

The goal is not to add more work for everybody, I hereby declare this PR as unnecessary. Thank you for your patience ;)

mabdinur: Oh I am sorry for not clarifying this earlier. I will close this PR 😄

mabdinur:
This change is no longer required. 

Context: https://github.com/DataDog/system-tests/pull/2407#discussion_r1591915860

tests/parametric/test_headers_tracecontext.py
dmehala: Actually, this change is still required as `p:XX!X` is less than 16, thus not valid. right? 👀 haha I'm sorry.

PR #2409, opened 2024-04-29
Testing remote sampling rules with tags.

yuanyuanzhao3:
## Workflow

## Reviewer checklist

### DIFF
--- a/tests/parametric/test_dynamic_configuration.py
+++ b/tests/parametric/test_dynamic_configuration.py
@@ -560,8 +560,7 @@ class TestDynamicConfigV2:
 class TestDynamicConfigSamplingRules:
     @parametrize("library_env", [{**DEFAULT_ENVVARS}])
     def test_capability_tracing_sample_rules(self, library_env, test_agent, test_library):
-        """Ensure the RC request contains the trace sampling rules capability.
-        """
+        """Ensure the RC request contains the trace sampling rules capability."""
         test_agent.wait_for_rc_capabilities([Capabilities.APM_TRACING_SAMPLE_RULES])
 
     @parametrize(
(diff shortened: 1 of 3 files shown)


tests/parametric/test_dynamic_configuration.py
```diff
@@ -121,10 +120,10 @@ def is_sampled(trace: List[Dict]):
     return trace[0]["metrics"].get("_sampling_priority_v1", 0) > 0
 
 
-def get_sampled_trace(test_library, test_agent, service, name):
-    trace = send_and_wait_trace(test_library, test_agent, service=service, name=name)
+def get_sampled_trace(test_library, test_agent, service, name, tags=None):
+    trace = send_and_wait_trace(test_library, test_agent, service=service, name=name, tags=tags)
```
ZStriker19, line 124: I think we need to add tags to the json being sent for the HTTP library implementations: https://github.com/DataDog/system-tests/blob/main/utils/parametric/_library_client.py#L178 Here we can see tags don't get put in the outgoing JSON. It does like it is added for GRPC already though, so don't need to do anything there.

tests/parametric/test_dynamic_configuration.py
```diff
@@ -56,7 +56,6 @@ def _default_config(service: str, env: str) -> Dict[str, Any]:
             "tracing_debug": None,
             "tracing_service_mapping": None,
             "tracing_sampling_rules": None,
-            "span_sampling_rules": None,
```
ZStriker19, line 59: We just removed this because it was never supposed to be there?

tests/parametric/test_dynamic_configuration.py
ZStriker19: Other work that might need to be done is per language on the server side. Looking at server.py (the python app actually generating the spans) it doesn't look like we're pulling out the tags and adding them to the span right now: https://github.com/DataDog/system-tests/blob/main/utils/build/docker/python/parametric/apm_test_client/server.py#L68-L98 

Of course you can just leave this to each language when they want to make sure they pass the tests, but just something to note for the future work that will be needed for each person trying to get this to pass for their language.

ZStriker19 (approved):

robertomonteromiguel (approved):

tests/parametric/test_dynamic_configuration.py
yuanyuanzhao3: yeah. we do not have span rules support.

tests/parametric/test_dynamic_configuration.py
yuanyuanzhao3: Good catch.

I was wondering how the test passed until I found the GRPC client which has the `span_tags` support.

Golang trace server/lib has the tags support. I also checked cpp tracer. It needs to add that as well, not just your Python tracer.

I'll leave the other language support in the test library to themselves. They should be catchable when they added tags feature and trying to pass this test.

PR #2455, opened 2024-05-15
Nginx with AppSec support

cataphract:
## Motivation

Nginx now supports AppSec. Enable those tesets

## Workflow

## Reviewer checklist


.github/workflows/run-end-to-end.yml
```diff
@@ -230,7 +235,9 @@ jobs:
         DD_API_KEY: ${{ secrets.DD_API_KEY }}
     - name: Run APPSEC_MISSING_RULES scenario
       if: always() && steps.build.outcome == 'success' && contains(inputs.scenarios, '"APPSEC_MISSING_RULES"')
-      run: ./run.sh APPSEC_MISSING_RULES
+      run: |
+        # nginx does not start with missing rules file, thereby stalling the job
```
robertomonteromiguel, line 239: could we add this condition to "if:"?
`      if: always() && steps.build.outcome == 'success' && contains(inputs.scenarios, '"APPSEC_MISSING_RULES"') && matrix.weblog != 'nginx'
`

.github/workflows/run-end-to-end.yml
```diff
@@ -240,7 +247,9 @@ jobs:
         DD_API_KEY: ${{ secrets.DD_API_KEY }}
     - name: Run APPSEC_CORRUPTED_RULES scenario
       if: always() && steps.build.outcome == 'success' && contains(inputs.scenarios, '"APPSEC_CORRUPTED_RULES"')
-      run: ./run.sh APPSEC_CORRUPTED_RULES
+      run: |
+        # nginx does not start with corrupted rules file, thereby stalling the job
+        if [[ $WEBLOG_VARIANT != nginx ]]; then ./run.sh APPSEC_CORRUPTED_RULES; fi
```
robertomonteromiguel, line 252: same....

tests/appsec/test_traces.py
```diff
@@ -329,7 +330,7 @@ class Test_DistributedTraceInfo:
     """Distributed traces info (Services, URL, trace id)"""
 
     def test_main(self):
-        assert False, "Test not implemented"
+        pytest.skip("Test not implemented")
```
robertomonteromiguel, line 333: same

robertomonteromiguel (changes requested):

utils/_context/containers.py
```diff
@@ -195,14 +199,23 @@ def execute_command(self, test, retries=10, interval=1_000_000_000, start_period
...
             time.sleep(interval)
 
+        logs = "<cannot retrieve logs>"
+        try:
+            logs = self._container.logs(stdout=True, stderr=True)
+            logs = logs.decode("utf-8")
+        except:
+            pass
+
+        print(logs)
```
robertomonteromiguel, line 217: why this? anyway we should use the logger

robertomonteromiguel (changes requested):

.github/workflows/run-end-to-end.yml
```diff
@@ -145,6 +145,11 @@ jobs:
     - name: Load PHP library binary (skipping release tests, use snapshot instead)
       if: ${{ matrix.version == 'prod' && inputs.library == 'php'}}
       run: ./utils/scripts/load-binary.sh ${{ inputs.library }}
+    - name: Load dev nginx module
+      if: ${{ matrix.version == 'dev' && inputs.library == 'cpp'}}
+      run: ./utils/scripts/load-binary.sh nginx
+      env:
+        CIRCLECI_TOKEN: ${{ secrets.CIRCLECI_TOKEN }}
```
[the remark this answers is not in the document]
cataphract: The previous step does:

```
./utils/scripts/load-binary.sh ${{ inputs.library }}
```

`inputs.library` will be cpp. However, I need to call it with `nginx`.

.github/workflows/run-end-to-end.yml
cataphract: I can try

tests/appsec/test_blocking_addresses.py
```diff
@@ -638,16 +654,16 @@ class Test_Suspicious_Request_Blocking:
 
     def test_blocking(self):
         """Test if requests that should be blocked are blocked"""
-        assert False, "TODO"
+        pytest.skip("TODO: test not implemented")
```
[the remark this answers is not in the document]
cataphract: The problem with `assert False` is that it requires that all the languages explicitly skip the test, which seems a bit silly to me, especially because what typically happens next when the test is implemented is that a bunch of libraries will have XPASS test results that take months (I can think of cases where it's more like years) to be tackled. But let me know if I should revert it.

utils/_context/containers.py
cataphract: When the healthcheck fails, this helps determine the cause of the failure.

tests/appsec/test_blocking_addresses.py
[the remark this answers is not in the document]
cbeauchesne: > The problem with assert False is that it requires that all the languages explicitly skip the test

It is exactly what we want :) without a proper test, we want to show in the output that the feature is not tested, and thus, we need to explicitly make it fail, and skip it. 

The good decorator for that is `missing_feature`. We could have done a decorator `missing_test`, but we decided a while to not use it, following this  key idea : `test is a part of the feature`.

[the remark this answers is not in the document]
cataphract: @cbeauchesne saying the test if part of the feature only (arguably) makes sense conceptually. In practice, the feature may be there, exposed to clients and reporting (correctly or incorrectly) data to the backend. The only obvious interpretation of a failure+skip if a missing or incorrect implementation. And by not explicitly skipping the test in every body you would immediately know once the test is implemented where the feature works fine (test passing, rather than XPASS, would indicates an _unexpected_ passing), and where it doesn't (test failing, and then an XFAIL would convey the proper meaning).

That said, this only affects me insofar as I need to explicitly marking all these unimplemented tests as failing. These annotations convey absolutely no useful information.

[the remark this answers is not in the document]
cbeauchesne: > In practice, the feature may be there, exposed to clients and reporting (correctly or incorrectly) data to the backend.

From testing POV, it does not make a difference. For instance,  "incorrectly" can include "not reported", which is exactly the same behavior as "not implemented". 

> especially because what typically happens next when the test is implemented is that a bunch of libraries will have XPASS test results that take months (I can think of cases where it's more like years) to be tackled

"years" is probably a little bit extreme, especially since we have the easy-wins reports. But anyway, when the test is  implemented, removing the decorators/update the manifest will tells you what are the passing libraries.

Actually, you're right, the current workflow requires : 

* to declare the test as failing using decorator/manifest
* to remove the decorator/manifest when you implement the test 

Which are two extra steps that can be seen as painfull 😄 

But the issue with your solution is that the report won't be correctly interpreted by the feature parity dashbaord (decorator formats the reason with identifiers that must be respected).

So if you really don't want to declare them with the correct way, you can simply removes the test (or totally comment it) : it will achieve what you want (no extra step), and won't mess up the feature parity dashbaord.

.github/workflows/ci.yml
```diff
@@ -89,6 +89,11 @@ jobs:
         uses: actions/checkout@v4
       - name: Get library artifact
         run: ./utils/scripts/load-binary.sh ${{ matrix.library }}
+      - name: Get nginx module
```
cbeauchesne, line 92: I feel like `./utils/scripts/load-binary.sh cpp` should load both nginx and cpp, WDYT ?

tests/appsec/test_traces.py
cbeauchesne: Can you either remove, or use official decorators ? 

Same for later `pytest.skip`

utils/_context/containers.py
```diff
@@ -143,6 +143,10 @@ def start(self) -> Container:
 
         logger.info(f"Start container {self.container_name}")
 
+        # the whole thing is reimplemented in python...
+        if self.healthcheck is not None:
```
cbeauchesne, line 147: I don't get this change ? 🤔

utils/_context/containers.py
```diff
@@ -205,14 +209,23 @@ def execute_command(self, test, retries=10, interval=1_000_000_000, start_period
                     return result
 
             except APIError as e:
-                logger.exception(f"Try #{i} failed")
-                pytest.exit(f"Command {cmd} failed for {self._container.name}: {e.explanation}", 1)
+                logger.exception(f"Try #{i} failed for {self._container.name}: {e.explanation}")
+                break
```
cbeauchesne, line 213: I think the process does not exit correctly without a `pytest.exit`

utils/_context/containers.py
```diff
@@ -205,14 +209,23 @@ def execute_command(self, test, retries=10, interval=1_000_000_000, start_period
...
 
             except Exception as e:
                 logger.debug(f"Try #{i}: {e}")
 
             time.sleep(interval)
 
+        logs = "<cannot retrieve logs>"
+        try:
```
cbeauchesne, line 221: Those logs are available in `logs/docker/<name>/stdout`

cbeauchesne (changes requested):

cbeauchesne:
Hi @cataphract ,

Do you need any help to fix issues/comments on this PR ?

manifests/cpp.yml
```diff
@@ -8,7 +8,37 @@ tests/:
     api_security/:
       test_api_security_rc.py: irrelevant (ASM is not implemented in C++)
       test_apisec_sampling.py: irrelevant (ASM is not implemented in C++)
-      test_schemas.py: irrelevant (ASM is not implemented in C++)
+      test_schemas.py:
+        Test_Scanners:
+          "*": irrelevant (ASM is not implemented in C++)
```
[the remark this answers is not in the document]
cataphract: ASM is implemented on the nginx module, not on dd-trace-cpp. If we're collapsing the distinction because the only supported weblog is nginx, then we should stop treating nginx separately in all the places in this file, no?

utils/_context/containers.py
cataphract: there's a healthcheck implemented in python if the healthcheck argument is passed. In this case, there's no point also running the healthcheck defined in the docker file

.github/actions/pull_images/action.yml
```diff
@@ -59,4 +59,4 @@ runs:
 
     - name: Pull
       shell: bash
-      run: docker compose pull
+      run: cat compose.yaml && docker compose pull
```
cbeauchesne, line 62: leftover ?

.github/workflows/ci.yml
cbeauchesne: Same question :)

manifests/cpp.yml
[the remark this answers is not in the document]
cbeauchesne: > then we should stop treating nginx separately in all the places in this file, no?

as long as it's the only supported weblog, yes, it'll be easier for you to maintain

tests/appsec/test_blocking_addresses.py
```diff
@@ -54,6 +57,7 @@ def test_blocking(self):
     def setup_blocking_before(self):
         self.block_req2 = weblog.get("/tag_value/tainted_value_6512/200", headers={"X-Forwarded-For": "1.1.1.1"})
 
+    @irrelevant(context.weblog_variant == "nginx", reason="Tag adding happens before WAF run")
```
cbeauchesne, line 60: out of curiosity, it means that it'll never be supported ?

tests/appsec/test_blocking_addresses.py
```diff
@@ -3,6 +3,9 @@
 # Copyright 2021 Datadog, Inc.
 
 import json
+
+import pytest
```
cbeauchesne, line 7: leftover ?

utils/_context/containers.py
cbeauchesne: interesting, do you have a dockerfile with such an healthcheck ?

utils/_context/containers.py
```diff
@@ -994,9 +998,10 @@ def __init__(self, host_log_folder) -> None:
 class MySqlContainer(SqlDbTestedContainer):
     def __init__(self, host_log_folder) -> None:
         super().__init__(
-            image_name="mysql/mysql-server:latest",
+            image_name="mysql/mysql-server:8.0.32",
             name="mysqldb",
-            command="--default-authentication-plugin=mysql_native_password",
+            command="--lc-messages-dir=/usr/share/mysql-8.0/english "
```
cbeauchesne, line 1003: out of curiosity, why this change ?

utils/build/docker/php/common/install_ddtrace.sh
```diff
@@ -44,6 +44,7 @@ fi
 #Extract version info
 php -d error_reporting='' -d extension=ddtrace.so -d extension=ddappsec.so -r 'echo phpversion("ddtrace");' > \
   /binaries/SYSTEM_TESTS_LIBRARY_VERSION
+#echo 1.2.0+f70ce41e4dfa0342fe5fac418fa1d8bffb02a52d > /binaries/SYSTEM_TESTS_LIBRARY_VERSION
```
cbeauchesne, line 47: leftover ?

cbeauchesne (changes requested):

.github/actions/pull_images/action.yml
cataphract: I left it on purpose because it shows what's actually going on, but i can remove it

.github/workflows/ci.yml
cataphract: depends if you want to collapse cpp and nginx or if you antecipate testing dd-trace-cpp directly

utils/_context/containers.py
cataphract: some of the images used in system-tests had them; it was the point of the change IIRC. But anyway, if you just examples you can [search GitHub](https://github.com/search?q=HEALTHCHECK+path%3ADockerfile&type=code&query=HEALTHCHECK+path%3ADockerfile)

manifests/cpp.yml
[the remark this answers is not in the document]
cataphract: Upon reflection, I think the simplification might not be a good idea, as there are other dd-trace-cpp based projects, like for Apache httpd.

 In light of that, the annotation may be correct insofar as dd-trace-cpp itself is not meant to implement ASM. It's only the nginx module that in addition to including dd-trace-cpp also has an ASM implementation. But you tell me

tests/appsec/test_blocking_addresses.py
cataphract: I'm not sure about never, but I can't see other use cases to set tags only after and conditionally on the blocking of response, except doing that just for having these tests passing. If there is a compelling, non purely test related use case, we can look at the viability.

tests/appsec/test_blocking_addresses.py
cataphract: so it seems

utils/_context/containers.py
cataphract: I got an error at some point about missing messages, probably related to a different locale in my machine? I don't remember exactly.

manifests/cpp.yml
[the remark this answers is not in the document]
cbeauchesne: The thing to get is that `cpp` in the context of system-tests does not mean `dd-trace-cpp`. It means  "the C++ language support by Datadog". 

Once this point clear in all minds, I'm fine with whatever you feel confortable with.

.github/workflows/ci.yml
cbeauchesne: See my comment above, `cpp` means "C++ language support", it does not mean dd-trace-cpp.

So basically, `./utils/scripts/load-binary.sh cpp` must load everything needed to test "last commit on main" from every repo involved in test the C++ language support by Datadog.

.github/actions/pull_images/action.yml
cbeauchesne: nope, that's fine.

cbeauchesne (dismissed):
We are close from the goal, there is only the change on `load-binary`, and AGTM

manifests/cpp.yml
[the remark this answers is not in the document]
cataphract: That's a bit of a distinction without a difference because support for C++ projects is provided through `dd-trace-cpp`.

On the other hand, neither nginx nor Apache httpd are implemented in C++; it just happens that due to C++ compatibility with C, we can use dd-trace-cpp (if we had dd-trace-zig, we could probably have used that instead). And because we're using dd-trace-cpp, the baseline compatibility matrix is the same as for dd-trace-cpp, so it wouldn't make much sense to separate them completely.

requirements.txt
```diff
@@ -45,3 +45,5 @@ types-retry==0.9.9.20250322
 watchdog==3.0.0
 yamllint==1.35.1
 yarl==1.9.4
+
+pydevd-pycharm~=221.6008.13
```
cbeauchesne, line 49: are you sure it won't mess up existing local installation ?

utils/build/docker/cpp_nginx/nginx/app.sh
```diff
@@ -3,7 +3,7 @@
 backend &
 
 if [[ "${DDPROF_ENABLE:-,,}" == "yes" ]]; then
-  ddprof -l notice nginx -g 'daemon off;'
+  ddprof -l notice nginx-debug -g 'daemon off;'
```
cbeauchesne, line 6: it would be better to have the default to be using the prod binaries. Though, there is an option that activate debug logs (`--force-dd-trace-debug`), this option sets the env var `DD_TRACE_DEBUG=true`. What about using `nginx-debug` if `DD_TRACE_DEBUG=true`, and use `nginx` otherwise ? 

Note that if you add `SYSTEM_TESTS_FORCE_DD_TRACE_DEBUG=true` to your `.env`, you'll always have this activated locally (more info here : https://github.com/DataDog/system-tests/blob/main/docs/execute/dd-trace-debug.md)

utils/interfaces/_logs.py
```diff
@@ -14,6 +14,10 @@
 from utils.interfaces._core import InterfaceValidator
 from utils._context.component_version import ComponentVersion
 
+import pydevd_pycharm
+
+pydevd_pycharm.settrace("localhost", port=5555, stdoutToServer=True, stderrToServer=True, suspend=False)
```
cbeauchesne, line 19: I'm a little bit afraid to have this everywhere (local laptops on mac/windows/linux, CI ...)

.github/workflows/compute-impacted-libraries.yml
```diff
@@ -91,7 +91,7 @@ jobs:
               "version": "dev",
             }
             for library in sorted(result)
-            if "otel" not in library and library not in ("cpp_nginx", )
+            if "otel" not in library and library not in ("cpp_httpd", "cpp_nginx")
```
cbeauchesne, line 94: it means that your removed `dev` CI  for `cpp_httpd`, is it what you want? If yes, what is needed to put it back ?

utils/_context/containers.py
```diff
@@ -217,6 +217,8 @@ def start(self, network: Network) -> Container:
             user=self.user,
             cap_add=self.cap_add,
             security_opt=self.security_opt,
+            # the whole thing is reimplemented in python...
```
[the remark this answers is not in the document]
cataphract: The tests already have their own healtcheck implemented in python with `wait_for_health`. Using the healthcheck of the container will just make it take longer for docker to declare the container available.

cataphract:
@cbeauchesne @robertomonteromiguel can you check again?

cbeauchesne:
Updating the branch to trigger the CI

cbeauchesne (changes requested):
Just a small RC, then AGTM

utils/_context/containers.py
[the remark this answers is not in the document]
cataphract: The line is:

```
            healthcheck={"test": ["NONE"]} if self.healthcheck is not None else None,
```

so the container healthcheck is only removed if a python check is provided.

cbeauchesne (approved):
Looks good to me. Could you also get a review from @dmehala  ?

run.sh
```diff
@@ -7,6 +7,7 @@
 set -e
 set -u
 set -o pipefail
+set -x
```
dmehala, line 10: Is this intended to keep?

utils/_context/containers.py
```diff
@@ -1239,9 +1239,10 @@ def __init__(self, host_log_folder: str) -> None:
 class MySqlContainer(SqlDbTestedContainer):
     def __init__(self, host_log_folder: str) -> None:
         super().__init__(
-            image_name="mysql/mysql-server:latest",
+            image_name="mysql/mysql-server:8.0.32",
```
dmehala, line 1242: expected to pin or leftover?

dmehala (approved):
LGTM.

cbeauchesne (approved):

PR #2501, opened 2024-05-28
parametric(cpp): enable prod and more parametric tests

dmehala:
## Motivation
dd-trace-cpp released a new version two days ago. This PR enable prod and enable more parametric tests.

## Changes
- CI
- Parametric tests for cpp

### DIFF
--- a/tests/parametric/test_dynamic_configuration.py
+++ b/tests/parametric/test_dynamic_configuration.py
@@ -685,7 +685,7 @@ class TestDynamicConfigSamplingRules:
             }
         ],
     )
-    @bug(context.library == "cpp")
+    @bug(context.library == "cpp", reason = "JSON tag format in RC differs from the JSON tag format used in DD_TRACE_SAMPLING_RULES")
     def test_trace_sampling_rules_with_tags(self, test_agent, test_library):
         """RC sampling rules with tags should match/skip spans with/without corresponding tag values.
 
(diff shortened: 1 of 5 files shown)


cbeauchesne:
FYI #2506 is trying to fix C++ failures

tests/parametric/test_dynamic_configuration.py
```diff
@@ -684,7 +685,8 @@ def test_trace_sampling_rules_override_rate(self, library_env, test_agent, test_
             }
         ],
     )
-    def test_trace_sampling_rules_with_tags(self, library_env, test_agent, test_library):
+    @bug(context.library == "cpp")
```
robertomonteromiguel, line 688: add the cause

robertomonteromiguel (changes requested):
There are test failures on CPP variant, but not in parametric. I guess it isn't related with this PR, right?

dmehala:
> There are test failures on CPP variant, but not in parametric. I guess it isn't related with this PR, right?

No. If you are referencing #2506 , then it's appsec related.

tests/parametric/test_dynamic_configuration.py
dmehala: > ~~add the cause~~
Could you add the cause? FTFY. Yes, I can.

dmehala, line 688:
```suggestion
    @bug(context.library == "cpp", reason = "JSON tag format in RC differs from the JSON tag format used in DD_TRACE_SAMPLING_RULES")
```

robertomonteromiguel (approved):

PR #2508, opened 2024-05-30
Include prerelases in SemVer version comparator

simon-id:
## Motivation

We use prerelease versions in our master branch like `6.0.0-pre`. Those prerelease versions cannot be compared to the semver version definitions in the manifest, and thus put all tests as xpass instead of pass.

## Changes

Trim that "-pre" suffit, it's not important really.

## Workflow

## Reviewer checklist


robertomonteromiguel (changes requested):
if you remove the version suffix, could the dev and prod have the same version?

simon-id:
@robertomonteromiguel on master we always have a major above what we release, so for example latest is 5.x.x but on master we have 6.0.0-pre

uurien (approved):

cbeauchesne:
> Those prerelease versions cannot be compared to the semver version definitions in the manifest

If I'm not wrong, in semver, `6.0.0-pre` will be after any `5.x.x` version, which is what we want, no ? Or do I miss something ?

simon-id:
@cbeauchesne
> 6.0.0-pre will be after any 5.x.x version

From what I can tell the spec is not super clear on it, but npm and the python library used in the system-tests do not agree with you, this is what i'm trying to solve. Basically without this fix, all the test from dev are xpass because 6.0.0-pre is considered not satisfying the >=5 filter

See:
image
image

robertomonteromiguel (approved):

simon-id:
I'm trying to fix it for node for now, we'll see if it impacts other libs when they start using it

simon-id:
I think the reason why they differ is because the `NpmSpec` is mimic'ing the [npm semver module](https://www.npmjs.com/package/semver). Here is a quote from its docs:
image
So a prerelease version will only match if compared to another prerelase version.
@cbeauchesne The problem with the `SimpleSpec`, is that it's actually too simple. It doesn't look like it supports OR clauses like the npmspec do, and that's like the whole reason why we're using it in the first place.

cbeauchesne:
indeed. 

let's go with a custom one !

cbeauchesne (approved):

simon-id:
LGTM thank you so much, way better solution, and i confirmed it works great locally :+1:

PR #2509, opened 2024-05-30
add system tests for otel env var mapping

khanayan123:
## Motivation

adds system tests for OTEL env var to DD env var mapping 

## Changes

adds a new test telemetry test to test telemetry metrics, a new otel env var test file under parametric/ to test correct mapping, & new http/grpc routes for sending config data from the test app

### DIFF
--- a/tests/parametric/test_otel_env_var.py
+++ b/tests/parametric/test_otel_env_var.py
@@ -1,0 +1,264 @@
+
+import pytest
+from utils import missing_feature, context, scenarios
+
+otel_env_var_to_internal_name_mapper = {
+    'nodejs': {
+        'OTEL_SERVICE_NAME': 'service',
+        'OTEL_LOG_LEVEL': 'logLevel',
+        'OTEL_TRACES_SAMPLER': 'sampleRate',
+        #'OTEL_TRACES_EXPORTER' not exposed by the tracer
+        'OTEL_METRICS_EXPORTER': 'runtimeMetrics',
+        'OTEL_RESOURCE_ATTRIBUTES': 'tags',
+        # both tracePropagationStyle.inject & tracePropagationStyle.extract set the value of OTEL_PROPAGATORS, so test either
+        'OTEL_PROPAGATORS': 'tracePropagationStyle.inject',
+        'OTEL_VERSION': 'version',
+        'OTEL_ENV': 'env'
+    }
+}
+
+def get_nested_dict_value(data, dot_key):
+    keys = dot_key.split('.')
+    for key in keys:
+        if isinstance(data, dict):
+            data = data.get(key)
+        else:
+            return None
+    return data
+
+def get_config_value(config, language, env_var):
+    return get_nested_dict_value(config, otel_env_var_to_internal_name_mapper[language][env_var])
+
+@scenarios.parametric
+class Test_Otel_Env_Var:
+    @missing_feature(context.library == "dotnet", reason="Not implemented")
+    @missing_feature(context.library == "java", reason="Not implemented")
+    @missing_feature(context.library == "golang", reason="Not implemented")
+    @missing_feature(context.library < "nodejs@5.11.0", reason="Implemented in v5.11.0, v4.35.0, &v3.56.0")
+    @missing_feature(context.library == "ruby", reason="Not implemented")
+    @missing_feature(context.library == "php", reason="Not implemented")
+    @pytest.mark.parametrize(
+        "library_env",
+        [
+            {
+                "DD_SERVICE": 'service',
+                "OTEL_SERVICE_NAME":  'otel_service',
+                "DD_TRACE_LOG_LEVEL": 'error',
+                "OTEL_LOG_LEVEL": 'debug',
+                "DD_TRACE_SAMPLE_RATE": '0.5',
+                "OTEL_TRACES_SAMPLER": 'traceidratio',
+                "OTEL_TRACES_SAMPLER_ARG": '0.1',
+                "DD_TRACE_ENABLED": 'true',
+                "OTEL_TRACES_EXPORTER": 'none',
+                "DD_RUNTIME_METRICS_ENABLED": 'true',
+                "OTEL_METRICS_EXPORTER": 'none',
+                "DD_TAGS": 'foo:bar,baz:qux',
+                "OTEL_RESOURCE_ATTRIBUTES": 'foo=bar,baz=qux',
+                "DD_TRACE_PROPAGATION_STYLE_INJECT": 'b3,tracecontext',
+                "DD_TRACE_PROPAGATION_STYLE_EXTRACT": 'b3,tracecontext',
+                "OTEL_PROPAGATORS": 'datadog,tracecontext'
+            }
+        ],
+    )
+    def test_dd_env_var_take_precedence(self, test_agent, test_library):
+        resp = test_library.get_tracer_config()
+        config = resp['config']
+        lang = resp['language']
+
+        assert get_config_value(config, lang, 'OTEL_SERVICE_NAME') == 'service'
+        assert get_config_value(config, lang, 'OTEL_LOG_LEVEL') == 'error'
+        assert get_config_value(config, lang, 'OTEL_TRACES_SAMPLER') == 0.5
+        assert get_config_value(config, lang, 'OTEL_METRICS_EXPORTER') == True
+        attr = get_config_value(config, lang, 'OTEL_RESOURCE_ATTRIBUTES')
+        assert attr.get("foo") == "bar"
+        assert attr.get("baz") == "qux"
+        assert get_config_value(config, lang, 'OTEL_PROPAGATORS') == ['b3', 'tracecontext']
+
+    @missing_feature(context.library == "dotnet", reason="Not implemented")
+    @missing_feature(context.library == "java", reason="Not implemented")
+    @missing_feature(context.library == "golang", reason="Not implemented")
+    @missing_feature(context.library < "nodejs@5.11.0", reason="Implemented in v5.11.0, v4.35.0, &v3.56.0")
+    @missing_feature(context.library == "ruby", reason="Not implemented")
+    @missing_feature(context.library == "php", reason="Not implemented")
+    @pytest.mark.parametrize(
+        "library_env",
+        [
+            {
+              'OTEL_SERVICE_NAME': 'otel_service',
+              'OTEL_LOG_LEVEL': 'warn',
+              'OTEL_TRACES_SAMPLER': 'traceidratio',
+              'OTEL_TRACES_SAMPLER_ARG': '0.1',
+              # when set this disables the tracer for Node and thus it cant return a config object
+              # 'OTEL_TRACES_EXPORTER': 'none',
+              'OTEL_METRICS_EXPORTER': 'none',
+              'OTEL_RESOURCE_ATTRIBUTES': 'foo=bar1,baz=qux1',
+              'OTEL_PROPAGATORS': 'b3,datadog'
+            }
+        ],
+    )
+    def test_otel_env_vars_set(self, test_agent, test_library):
+        resp = test_library.get_tracer_config()
+        config = resp['config']
+        lang = resp['language']
+        
+        assert get_config_value(config, lang, 'OTEL_SERVICE_NAME') == 'otel_service'
+        assert get_config_value(config, lang, 'OTEL_LOG_LEVEL') == 'warn'
+        assert get_config_value(config, lang, 'OTEL_TRACES_SAMPLER') == 0.1
+        assert get_config_value(config, lang, 'OTEL_METRICS_EXPORTER') == False
+        attr = get_config_value(config, lang, 'OTEL_RESOURCE_ATTRIBUTES')
+        assert attr.get("foo") == "bar1"
+        assert attr.get("baz") == "qux1"
+        assert get_config_value(config, lang, 'OTEL_PROPAGATORS') == ['b3', 'datadog']
+
+    @missing_feature(context.library == "dotnet", reason="Not implemented")
+    @missing_feature(context.library == "java", reason="Not implemented")
+    @missing_feature(context.library == "golang", reason="Not implemented")
+    @missing_feature(context.library < "nodejs@5.11.0", reason="Implemented in v5.11.0, v4.35.0, &v3.56.0")
+    @missing_feature(context.library == "ruby", reason="Not implemented")
+    @missing_feature(context.library == "php", reason="Not implemented")
+    @pytest.mark.parametrize(
+        "library_env",
+        [
+            {
+              'OTEL_RESOURCE_ATTRIBUTES': 
+                'deployment.environment=test1,service.name=test2,service.version=5,foo=bar1,baz=qux1'
+            }
+        ],
+    )
+    def test_otel_attribute_mapping(self, test_agent, test_library):
+        resp = test_library.get_tracer_config()
+        config = resp['config']
+        lang = resp['language']
+
+        assert get_config_value(config, lang, 'OTEL_SERVICE_NAME') == 'test2'
+        assert get_config_value(config, lang, 'OTEL_ENV') == 'test1'
+        assert get_config_value(config, lang, 'OTEL_VERSION') == '5'
+        attr = get_config_value(config, lang, 'OTEL_RESOURCE_ATTRIBUTES')
+        assert attr.get("foo") == "bar1"
+        assert attr.get("baz") == "qux1"
+
+    @missing_feature(context.library == "dotnet", reason="Not implemented")
+    @missing_feature(context.library == "java", reason="Not implemented")
+    @missing_feature(context.library == "golang", reason="Not implemented")
+    @missing_feature(context.library < "nodejs@5.11.0", reason="Implemented in v5.11.0, v4.35.0, &v3.56.0")
+    @missing_feature(context.library == "ruby", reason="Not implemented")
+    @missing_feature(context.library == "php", reason="Not implemented")
+    @pytest.mark.parametrize(
+        "library_env",
+        [
+            {
+              'OTEL_TRACES_SAMPLER': 'always_on',
+            }
+        ],
+    )
+    def test_otel_traces_always_on(self, test_agent, test_library):
+        resp = test_library.get_tracer_config()
+        config = resp['config']
+        lang = resp['language']
+        assert get_config_value(config, lang, 'OTEL_TRACES_SAMPLER') == 1.0
+
+    @missing_feature(context.library == "dotnet", reason="Not implemented")
+    @missing_feature(context.library == "java", reason="Not implemented")
+    @missing_feature(context.library == "golang", reason="Not implemented")
+    @missing_feature(context.library < "nodejs@5.11.0", reason="Implemented in v5.11.0, v4.35.0, &v3.56.0")
+    @missing_feature(context.library == "ruby", reason="Not implemented")
+    @missing_feature(context.library == "php", reason="Not implemented")
+    @pytest.mark.parametrize(
+        "library_env",
+        [
+            {
+              'OTEL_TRACES_SAMPLER': 'always_off',
+            }
+        ],
+    )
+    def test_otel_traces_always_off(self, test_agent, test_library):
+        resp = test_library.get_tracer_config()
+        config = resp['config']
+        lang = resp['language']
+        assert get_config_value(config, lang, 'OTEL_TRACES_SAMPLER') == 0.0
+
+    @missing_feature(context.library == "dotnet", reason="Not implemented")
+    @missing_feature(context.library == "java", reason="Not implemented")
+    @missing_feature(context.library == "golang", reason="Not implemented")
+    @missing_feature(context.library < "nodejs@5.11.0", reason="Implemented in v5.11.0, v4.35.0, &v3.56.0")
+    @missing_feature(context.library == "ruby", reason="Not implemented")
+    @missing_feature(context.library == "php", reason="Not implemented")
+    @pytest.mark.parametrize(
+        "library_env",
+        [
+            {
+              'OTEL_TRACES_SAMPLER': 'traceidratio',
+              'OTEL_TRACES_SAMPLER_ARG': '0.1'
+            }
+        ],
+    )
+    def test_otel_traces_traceidratio(self, test_agent, test_library):
+        resp = test_library.get_tracer_config()
+        config = resp['config']
+        lang = resp['language']
+
+        assert get_config_value(config, lang, 'OTEL_TRACES_SAMPLER') == 0.1
+
+    @missing_feature(context.library == "dotnet", reason="Not implemented")
+    @missing_feature(context.library == "java", reason="Not implemented")
+    @missing_feature(context.library == "golang", reason="Not implemented")
+    @missing_feature(context.library < "nodejs@5.11.0", reason="Implemented in v5.11.0, v4.35.0, &v3.56.0")
+    @missing_feature(context.library == "ruby", reason="Not implemented")
+    @missing_feature(context.library == "php", reason="Not implemented")
+    @pytest.mark.parametrize(
+      "library_env",
+      [
+          {
+            'OTEL_TRACES_SAMPLER': 'parentbased_always_on',
+          }
+      ],
+    )
+    def test_otel_traces_parentbased_on(self, test_agent, test_library):
+      resp = test_library.get_tracer_config()
+      config = resp['config']
+      lang = resp['language']
+
+      assert get_config_value(config, lang, 'OTEL_TRACES_SAMPLER') == 1.0
+
+    @missing_feature(context.library == "dotnet", reason="Not implemented")
+    @missing_feature(context.library == "java", reason="Not implemented")
+    @missing_feature(context.library == "golang", reason="Not implemented")
+    @missing_feature(context.library < "nodejs@5.11.0", reason="Implemented in v5.11.0, v4.35.0, &v3.56.0")
+    @missing_feature(context.library == "ruby", reason="Not implemented")
+    @missing_feature(context.library == "php", reason="Not implemented")
+    @pytest.mark.parametrize(
+      "library_env",
+      [
+          {
+            'OTEL_TRACES_SAMPLER': 'parentbased_always_off',
+          }
+      ],
+    )
+    def test_otel_traces_parentbased_off(self, test_agent, test_library):
+      resp = test_library.get_tracer_config()
+      config = resp['config']
+      lang = resp['language']
+
+      assert get_config_value(config, lang, 'OTEL_TRACES_SAMPLER') == 0.0
+
+    @missing_feature(context.library == "dotnet", reason="Not implemented")
+    @missing_feature(context.library == "java", reason="Not implemented")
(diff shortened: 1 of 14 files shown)


utils/build/docker/nodejs/parametric/server.js
```diff
@@ -285,6 +285,11 @@ app.post('/trace/otel/set_attributes', (req, res) => {
   res.json({});
 });
 
+app.get('/trace/config', (req, res) => {
+  const dummyTracer = require('dd-trace').init()
+  res.json( { config: dummyTracer._tracer._config, language: 'nodejs' });
```
mabdinur, line 290: Instead of returning the whole config object can we return just a set of key's and values. This will allow us to standardize the return value across languages. 

Ex:
```suggestion
  config = dummyTracer._tracer._config
  res.json( { "tracer_enabled": config._tracer_enabled,  "propagation_styles": config._propagation_styles, ...... });
```

We can add to this list as needed.

tests/parametric/test_otel_env_var.py
```diff
@@ -0,0 +1,233 @@
...
+        [
+            {
+                "DD_SERVICE": "service",
+                "OTEL_SERVICE_NAME": "otel_service",
+                "DD_TRACE_LOG_LEVEL": "error",
+                "OTEL_LOG_LEVEL": "debug",
+                "DD_TRACE_SAMPLE_RATE": "0.5",
+                "OTEL_TRACES_SAMPLER": "traceidratio",
+                "OTEL_TRACES_SAMPLER_ARG": "0.1",
+                "DD_TRACE_ENABLED": "true",
```
zacharycmontoya, line 53: Is there a corresponding assert for DD_TRACE_ENABLED?

utils/build/docker/nodejs/parametric/server.js
zacharycmontoya: +1, I think it would be cleaner to move the translation logic into the test app so the test runner/test cases don't have that complexity

tests/parametric/test_otel_env_var.py
khanayan123: i can add one as a separate test because for node dd_trace_enabled is not exposed in the config object

tests/parametric/test_telemetry.py
```diff
@@ -164,6 +164,80 @@ def test_library_settings(self, library_env, test_agent, test_library):
...
+            ["DD_TRACE_PROPAGATION_STYLE", "OTEL_PROPAGATORS"],
+            ["DD_SERVICE", "OTEL_SERVICE_NAME"],
+            ["DD_TRACE_SAMPLE_RATE", "OTEL_TRACES_SAMPLER", "OTEL_TRACES_SAMPLER_ARG"],
+            ["DD_TRACE_ENABLED", "OTEL_TRACES_EXPORTER"],
+            ["DD_RUNTIME_METRICS_ENABLED", "OTEL_METRICS_EXPORTER"],
+            ["DD_TAGS", "OTEL_RESOURCE_ATTRIBUTES"],
+            ["DD_TRACE_OTEL_ENABLED", "OTEL_SDK_DISABLED"],
+        ]
+
+        for i, tags in enumerate(expected_tags):
```
zacharycmontoya, line 229: Are you defining the order of the `expected_tags` so that it matches up with the metrics payload you observed? If so, that seems like it could be brittle

utils/build/docker/nodejs/parametric/server.js
```diff
@@ -285,6 +285,25 @@ app.post('/trace/otel/set_attributes', (req, res) => {
   res.json({});
 });
 
+app.get('/trace/config', (req, res) => {
+  const dummyTracer = require('dd-trace').init()
+  const config = dummyTracer._tracer._config
+  res.json( { 
+    'dd_service': config ? config.service : null,
```
zacharycmontoya, line 292: Is this valid JavaScript? If so, it may be cleaner

```suggestion
    'dd_service': config?.service ?? null,
```

tests/parametric/test_otel_env_vars.py
```diff
@@ -0,0 +1,220 @@
+import pytest
+from utils import missing_feature, context, scenarios
+
+
+@scenarios.parametric
+class Test_Otel_Env_Vars:
+    @missing_feature(context.library == "dotnet", reason="Not implemented")
+    @missing_feature(context.library == "java", reason="Not implemented")
+    @missing_feature(context.library == "golang", reason="Not implemented")
+    @missing_feature(context.library < "nodejs@5.11.0", reason="Implemented in v5.11.0, v4.35.0, &v3.56.0")
```
zacharycmontoya, line 10: I think this line is irrelevant due to the `manifests/nodejs.yml` entry
```suggestion
    @missing_feature(context.library < "nodejs@5.11.0", reason="Implemented in v5.11.0, v4.35.0, &v3.56.0")
```

tests/parametric/test_otel_env_vars.py
```diff
@@ -0,0 +1,220 @@
...
+                "OTEL_LOG_LEVEL": "debug",
+                "DD_TRACE_SAMPLE_RATE": "0.5",
+                "OTEL_TRACES_SAMPLER": "traceidratio",
+                "OTEL_TRACES_SAMPLER_ARG": "0.1",
+                "DD_TRACE_ENABLED": "true",
+                "OTEL_TRACES_EXPORTER": "none",
+                "DD_RUNTIME_METRICS_ENABLED": "true",
+                "OTEL_METRICS_EXPORTER": "none",
+                "DD_TAGS": "foo:bar,baz:qux",
+                "OTEL_RESOURCE_ATTRIBUTES": "foo=bar,baz=qux",
```
zacharycmontoya, line 31: A small suggestion to differentiate between the DD tags and the OTEL tags:
```suggestion
                "OTEL_RESOURCE_ATTRIBUTES": "foo=otel_bar,baz=otel_qux",
```

tests/parametric/test_otel_env_vars.py
```diff
@@ -0,0 +1,220 @@
...
+    @missing_feature(context.library < "nodejs@5.11.0", reason="Implemented in v5.11.0, v4.35.0, &v3.56.0")
+    @missing_feature(context.library == "ruby", reason="Not implemented")
+    @missing_feature(context.library == "php", reason="Not implemented")
+    @pytest.mark.parametrize(
+        "library_env", [{"OTEL_TRACES_SAMPLER": "always_on",}],
+    )
+    def test_otel_traces_always_on(self, test_agent, test_library):
+        resp = test_library.get_tracer_config()
+        if context.library != "nodejs":
+            assert resp["dd_trace_sample_ignore_parent"] == False
```
zacharycmontoya, line 131:
```suggestion
            assert resp["dd_trace_sample_ignore_parent"] == True
```

tests/parametric/test_otel_env_vars.py
```diff
@@ -0,0 +1,220 @@
...
+    @missing_feature(context.library < "nodejs@5.11.0", reason="Implemented in v5.11.0, v4.35.0, &v3.56.0")
+    @missing_feature(context.library == "ruby", reason="Not implemented")
+    @missing_feature(context.library == "php", reason="Not implemented")
+    @pytest.mark.parametrize(
+        "library_env", [{"OTEL_TRACES_SAMPLER": "always_off",}],
+    )
+    def test_otel_traces_always_off(self, test_agent, test_library):
+        resp = test_library.get_tracer_config()
+        if context.library != "nodejs":
+            assert resp["dd_trace_sample_ignore_parent"] == False
```
zacharycmontoya, line 146:
```suggestion
            assert resp["dd_trace_sample_ignore_parent"] == True
```

tests/parametric/test_otel_env_vars.py
```diff
@@ -0,0 +1,220 @@
...
+    @missing_feature(context.library < "nodejs@5.11.0", reason="Implemented in v5.11.0, v4.35.0, &v3.56.0")
+    @missing_feature(context.library == "ruby", reason="Not implemented")
+    @missing_feature(context.library == "php", reason="Not implemented")
+    @pytest.mark.parametrize(
+        "library_env", [{"OTEL_TRACES_SAMPLER": "traceidratio", "OTEL_TRACES_SAMPLER_ARG": "0.1"}],
+    )
+    def test_otel_traces_traceidratio(self, test_agent, test_library):
+        resp = test_library.get_tracer_config()
+        if context.library != "nodejs":
+            assert resp["dd_trace_sample_ignore_parent"] == False
```
zacharycmontoya, line 161:
```suggestion
            assert resp["dd_trace_sample_ignore_parent"] == True
```

tests/parametric/test_otel_env_vars.py
```diff
@@ -0,0 +1,220 @@
...
+    @missing_feature(context.library == "golang", reason="Not implemented")
+    @missing_feature(context.library < "nodejs@5.11.0", reason="Implemented in v5.11.0, v4.35.0, &v3.56.0")
+    @missing_feature(context.library == "ruby", reason="Not implemented")
+    @missing_feature(context.library == "php", reason="Not implemented")
+    @pytest.mark.parametrize(
+        "library_env",
+        [
+            {
+                "OTEL_SERVICE_NAME": "otel_service",
+                "OTEL_LOG_LEVEL": "debug",
```
zacharycmontoya, line 68: Can you add a standalone `OTEL_LOG_LEVEL=error` and/or standalone `OTEL_LOG_LEVEL=debug` so we can assert the two different behaviors? Namely, that `debug` also sets `DD_TRACE_DEBUG=true`

zacharycmontoya (approved):
LGTM from API side

tests/parametric/test_otel_env_vars.py
```diff
@@ -0,0 +1,220 @@
+import pytest
+from utils import missing_feature, context, scenarios
+
+
+@scenarios.parametric
```
zacharycmontoya, line 5: I know we're dealing with OpenTelemetry...but all the other OTEL classes have this same features name 😆 

```suggestion
@scenarios.parametric
@features.open_tracing_api
```

tests/parametric/test_otel_env_vars.py
khanayan123: I think we want to leave it in there for best practice just to remember which release it was implemented for future reference

tests/parametric/test_telemetry.py
zacharycmontoya, line 1: I added a question in the RFC that might affect how the tags are represented on the metrics. As such, we should make sure to resolve that thread before merging these changes

tests/parametric/test_otel_env_vars.py
```diff
@@ -0,0 +1,234 @@
...
+    def test_otel_traces_exporter_none(self, test_agent, test_library):
+        resp = test_library.get_tracer_config()
+        assert resp["dd_trace_enabled"] == False
+
+    @missing_feature(context.library == "dotnet", reason="Not implemented")
+    @missing_feature(context.library == "java", reason="Not implemented")
+    @missing_feature(context.library == "golang", reason="Not implemented")
+    @missing_feature(context.library == "nodejs", reason="Not implemented")
+    @missing_feature(context.library == "ruby", reason="Not implemented")
+    @missing_feature(context.library == "php", reason="Not implemented")
```
robertomonteromiguel, line 227: If we are setting "missing_feature" in the manifests, and we are going to skip all the class methods, Do we need all these decorators?

tests/parametric/test_telemetry.py
```diff
@@ -164,6 +164,144 @@ def test_library_settings(self, library_env, test_agent, test_library):
...
+            assert metric["points"][0][1] == 1
+
+    @missing_feature(context.library == "dotnet", reason="Not implemented")
+    @missing_feature(context.library == "java", reason="Not implemented")
+    @missing_feature(context.library == "golang", reason="Not implemented")
+    @missing_feature(context.library == "nodejs", reason="Not implemented")
+    @missing_feature(context.library == "ruby", reason="Not implemented")
+    @missing_feature(context.library == "php", reason="Not implemented")
+    @missing_feature(context.library == "cpp", reason="Not implemented")
+    @missing_feature(context.library == "python", reason="Not implemented")
```
robertomonteromiguel, line 248: we are not running this test for any language?

robertomonteromiguel (changes requested):
There are tests failures on the parametric scenario for Java and Ruby

tests/parametric/test_telemetry.py
khanayan123: After an update to the RFC I had to make changes to the Node.js implementation. I have a PR up for that reimplementation in this PR https://github.com/DataDog/dd-trace-js/pull/4369 which I have tested against locally with those system tests

tests/parametric/test_otel_env_vars.py
```diff
@@ -0,0 +1,179 @@
+import pytest, os
+from utils import missing_feature, context, scenarios, features
+
+
+@scenarios.parametric
+@features.open_tracing_api
```
mabdinur, line 6:
```suggestion
```

tests/parametric/test_otel_env_vars.py
```diff
@@ -0,0 +1,179 @@
...
+        assert resp["dd_trace_enabled"] == True
+        assert resp["dd_runtime_metrics_enabled"] == True
+        tags = resp["dd_tags"]
+        assert tags.get("foo") == "bar"
+        assert tags.get("baz") == "qux"
+        assert resp["dd_trace_propagation_style"] == "b3,tracecontext"
+        assert resp["dd_trace_debug"] == False
+        if context.library != "nodejs":
+            assert resp["dd_trace_otel_enabled"] == False  # node does not expose this variable in the config
+            assert resp["dd_trace_sample_ignore_parent"] == True  # node does not implement this env variable
```
mabdinur, line 51: We should move each of these to their own test cases. I don't think any language supports `dd_trace_sample_ignore_parent` yet and and `dd_trace_otel_enabled` is not supported in ruby or go.

tests/parametric/test_otel_env_vars.py
```diff
@@ -0,0 +1,179 @@
...
+        assert resp["dd_trace_sample_rate"] == 0.1
+        assert resp["dd_trace_enabled"] == True
+        assert resp["dd_runtime_metrics_enabled"] == False
+        tags = resp["dd_tags"]
+        assert tags.get("foo") == "bar1"
+        assert tags.get("baz") == "qux1"
+        assert resp["dd_trace_propagation_style"] == "b3,datadog"
+        if context.library != "nodejs":
+            assert resp["dd_trace_otel_enabled"] == False
+            assert resp["dd_trace_sample_ignore_parent"] == True
```
mabdinur, line 83: Same thing here. We can have a common test case for the configs languages support and special cases for others.

tests/parametric/test_otel_env_vars.py
mabdinur: OR we can just assert the otel values are not set. That could be easier and would serve as a better catch all.

tests/parametric/test_otel_env_vars.py
```diff
@@ -0,0 +1,179 @@
...
+        ],
+    )
+    def test_dd_env_var_take_precedence(self, test_agent, test_library):
+        resp = test_library.get_tracer_config()
+
+        assert resp["dd_service"] == "service"
+        assert resp["dd_log_level"] == "error"
+        assert resp["dd_trace_sample_rate"] == 0.5
+        assert resp["dd_trace_enabled"] == True
+        assert resp["dd_runtime_metrics_enabled"] == True
```
mabdinur, line 43: nit: This will probably get caught by the python linter

```suggestion
        assert resp["dd_runtime_metrics_enabled"]
```

utils/parametric/_library_client.py
```diff
@@ -350,6 +353,23 @@ def http_client_request(self, method: str, url: str, headers: List[Tuple[str, st
...
+            "dd_service": resp["dd_service"],
+            "dd_log_level": resp["dd_log_level"],
+            "dd_trace_sample_rate": resp["dd_trace_sample_rate"],
+            "dd_trace_enabled": resp["dd_trace_enabled"],
+            "dd_runtime_metrics_enabled": resp["dd_runtime_metrics_enabled"],
+            "dd_tags": resp["dd_tags"],
+            "dd_trace_propagation_style": resp["dd_trace_propagation_style"],
+            "dd_trace_debug": resp["dd_trace_debug"],
+            "dd_trace_otel_enabled": resp["dd_trace_otel_enabled"],
+            "dd_trace_sample_ignore_parent": resp["dd_trace_sample_ignore_parent"],
```
mabdinur, line 368: If this key does not exist in the response this test will raise a key error. Is this acceptable? Do we want to return a default value? 

Ex:
```suggestion
            "dd_trace_sample_ignore_parent": resp.get("dd_trace_sample_ignore_parent", None),
```

utils/parametric/_library_client.py
```diff
@@ -350,6 +353,23 @@ def http_client_request(self, method: str, url: str, headers: List[Tuple[str, st
         ).json()
         return resp
 
+    def get_tracer_config(self) -> dict:
```
mabdinur, line 357: It would be nice if the type of the returned dictionary was well defined. Something like:

```suggestion
    def get_tracer_config(self) -> Dict[str, str]:
```

OR can `get_tracer_config` return an object that has well defined fields. This might be easier to maintain

utils/parametric/_library_client.py
```diff
@@ -624,6 +644,23 @@ def otel_get_span_context(self, span_id: int) -> OtelSpanContext:
     def otel_flush(self, timeout: int) -> bool:
         return self._client.OtelFlushSpans(pb.OtelFlushSpansArgs(seconds=timeout)).success
 
+    def get_tracer_config(self) -> dict:
```
mabdinur, line 650: Same here. We need better typing. Maintain dictionaries with undefined values can be painful

utils/parametric/protos/apm_test_client.proto
```diff
@@ -41,6 +41,25 @@ service APMClient {
   rpc OtelGetLinks(OtelGetLinksArgs) returns (OtelGetLinksReturn) {}
 
   rpc StopTracer(StopTracerArgs) returns (StopTracerReturn) {}
+
+  rpc GetTraceConfig(GetTraceConfigArgs) returns (GetTraceConfigReturn) {}
+}
+
+message GetTraceConfigArgs {}
+
+message GetTraceConfigReturn {
```
mabdinur, line 50: We spoke about this offline but we will convert this proto to map<string, string> config = 6;

tests/parametric/test_otel_env_vars.py
```diff
@@ -0,0 +1,190 @@
...
+        resp = test_library.get_tracer_config()
+
+        assert resp["dd_service"] == "service"
+        assert resp["dd_log_level"] == "error"
+        assert resp["dd_trace_sample_rate"] == "0.5"
+        assert resp["dd_trace_enabled"] == "true"
+        assert resp["dd_runtime_metrics_enabled"]
+        tags = resp["dd_tags"]
+        assert "foo:bar" in tags
+        assert "baz:qux" in tags
```
mabdinur, line 44: We should also assert that that the Otel resources are not in tags

tests/parametric/test_otel_env_vars.py
```diff
@@ -0,0 +1,190 @@
...
+        resp = test_library.get_tracer_config()
+        assert resp["dd_trace_sample_rate"] == "1.0" or "1"
+
+    @missing_feature(context.library < "nodejs@5.11.0", reason="Implemented in v5.11.0, v4.35.0, &v3.56.0")
+    @pytest.mark.parametrize(
+        "library_env", [{"OTEL_TRACES_SAMPLER": "always_off",}],
+    )
+    def test_otel_traces_always_off(self, test_agent, test_library):
+        resp = test_library.get_tracer_config()
+        assert resp["dd_trace_sample_rate"] == "0.0" or "0"
```
mabdinur, line 109: This will always evaluate to True 
```suggestion
        assert float(resp["dd_trace_sample_rate"]) == 0.0
```

tests/parametric/test_otel_env_vars.py
```diff
@@ -0,0 +1,190 @@
...
+
+    @missing_feature(context.library < "nodejs@5.11.0", reason="Implemented in v5.11.0, v4.35.0, &v3.56.0")
+    @pytest.mark.parametrize(
+        "library_env", [{"OTEL_TRACES_SAMPLER": "traceidratio", "OTEL_TRACES_SAMPLER_ARG": "0.1"}],
+    )
+    def test_otel_traces_traceidratio(self, test_agent, test_library):
+        resp = test_library.get_tracer_config()
+        assert resp["dd_trace_sample_rate"] == "0.1"
```
mabdinur, line 117: Same here

tests/parametric/test_otel_env_vars.py
```diff
@@ -0,0 +1,190 @@
...
+
+    @missing_feature(context.library < "nodejs@5.11.0", reason="Implemented in v5.11.0, v4.35.0, &v3.56.0")
+    @pytest.mark.parametrize(
+        "library_env", [{"OTEL_TRACES_SAMPLER": "parentbased_always_on",}],
+    )
+    def test_otel_traces_parentbased_on(self, test_agent, test_library):
+        resp = test_library.get_tracer_config()
+        assert resp["dd_trace_sample_rate"] == "1.0" or "1"
+
```
mabdinur, line 126: Same here

tests/parametric/test_otel_env_vars.py
```diff
@@ -0,0 +1,190 @@
...
+    @missing_feature(context.library < "nodejs@5.11.0", reason="Implemented in v5.11.0, v4.35.0, &v3.56.0")
+    @pytest.mark.parametrize(
+        "library_env", [{"OTEL_TRACES_SAMPLER": "parentbased_always_off",}],
+    )
+    def test_otel_traces_parentbased_off(self, test_agent, test_library):
+        resp = test_library.get_tracer_config()
+        assert resp["dd_trace_sample_rate"] == "0.0" or "0"
```
mabdinur, line 133: Same here

tests/parametric/test_otel_env_vars.py
```diff
@@ -0,0 +1,190 @@
...
+
+    @missing_feature(context.library < "nodejs@5.11.0", reason="Implemented in v5.11.0, v4.35.0, &v3.56.0")
+    @pytest.mark.parametrize(
+        "library_env", [{"OTEL_TRACES_SAMPLER": "parentbased_always_on",}],
+    )
+    def test_otel_traces_parentbased_on(self, test_agent, test_library):
+        resp = test_library.get_tracer_config()
+        assert resp["dd_trace_sample_rate"] == "1.0" or "1"
+
+    @missing_feature(context.library < "nodejs@5.11.0", reason="Implemented in v5.11.0, v4.35.0, &v3.56.0")
```
mabdinur, line 127: We can put this restriction in the node manifest file instead of annotating every test.

mabdinur (commented):
Overall looks okay to me. We can iterate on the test cases as we add support for more languages

tests/parametric/test_otel_env_vars.py
```diff
@@ -0,0 +1,190 @@
...
+        resp = test_library.get_tracer_config()
+        assert resp["dd_trace_otel_enabled"] == "true"
+
+    @missing_feature(context.library == "nodejs", reason="Not implemented")
+    @pytest.mark.parametrize(
+        "library_env", [{"OTEL_SDK_DISABLED": "true"}],
+    )
+    def test_otel_sdk_disabled_set(self, test_agent, test_library):
+        resp = test_library.get_tracer_config()
+        assert resp["dd_trace_otel_enabled"] == "true"
```
zacharycmontoya, line 174: Setting `OTEL_SDK_DISABLED` to true should turn off `DD_TRACE_OTEL_ENABLED`, right?
```suggestion
        assert resp["dd_trace_otel_enabled"] == "false"
```

tests/parametric/test_otel_env_vars.py
```diff
@@ -0,0 +1,182 @@
...
+                "OTEL_TRACES_SAMPLER": "traceidratio",
+                "OTEL_TRACES_SAMPLER_ARG": "0.1",
+                "DD_TRACE_ENABLED": "true",
+                "OTEL_TRACES_EXPORTER": "none",
+                "DD_RUNTIME_METRICS_ENABLED": "true",
+                "OTEL_METRICS_EXPORTER": "none",
+                "DD_TAGS": "foo:bar,baz:qux",
+                "OTEL_RESOURCE_ATTRIBUTES": "foo=otel_bar,baz=otel_qux",
+                "DD_TRACE_PROPAGATION_STYLE_INJECT": "b3,tracecontext",
+                "DD_TRACE_PROPAGATION_STYLE_EXTRACT": "b3,tracecontext",
```
mabdinur, line 28: Does node support DD_TRACE_PROPAGATION_STYLES? Python is planning to remove `DD_TRACE_PROPAGATION_STYLE_INJECT` and `DD_TRACE_PROPAGATION_STYLE_EXTRACT` in the next release so I would prefer to use `DD_TRACE_PROPAGATION_STYLES` here

```suggestion
                "DD_TRACE_PROPAGATION_STYLE": "b3,tracecontext",
```

tests/parametric/test_otel_env_vars.py
```diff
@@ -0,0 +1,182 @@
...
+    @missing_feature(context.library == "nodejs", reason="Not implemented")
+    @pytest.mark.parametrize(
+        "library_env", [{"OTEL_LOG_LEVEL": "debug"}],
+    )
+    def test_otel_log_level_to_debug_mapping(self, test_agent, test_library):
+        resp = test_library.get_tracer_config()
+        assert resp["dd_log_level"] == "debug"
+        assert resp["dd_trace_debug"] == "true"
+
+    @missing_feature(context.library == "nodejs", reason="Not implemented")
```
mabdinur, line 152: nit: can we give a more specific reason here. Why/what is not implemented in node

tests/parametric/test_otel_env_vars.py
```diff
@@ -0,0 +1,182 @@
...
+    @pytest.mark.parametrize(
+        "library_env", [{"OTEL_TRACES_SAMPLER": "always_on"}],
+    )
+    def test_dd_trace_sample_ignore_parent_true(self, test_agent, test_library):
+        resp = test_library.get_tracer_config()
+        assert resp["dd_trace_sample_ignore_parent"] == "true"
+
+    @missing_feature(context.library == "nodejs", reason="Not implemented")
```
mabdinur, line 176:
```suggestion
    @missing_feature(True, reason="dd_trace_sample_ignore_parent requires an RFC, this feature is not implemented in any language")
```

utils/parametric/_library_client.py
```diff
@@ -350,6 +354,25 @@ def http_client_request(self, method: str, url: str, headers: List[Tuple[str, st
         ).json()
         return resp
 
+    def get_tracer_config(self) -> Dict[str, Optional[str]]:
+        resp = self._session.get(self._url("/trace/config")).json()
+        config = resp["config"]
+        config_dict = json.loads(config)
```
mabdinur, line 360: nit:
```suggestion
        config_dict = json.loads(resp["config"])
```

mabdinur (approved):
Left some nits but overall LGTM

robertomonteromiguel (approved):

PR #2513, opened 2024-05-31
Check app-started as first message by seq_id

smola:
## Motivation

We allow skipping seq_id on failure, as long as it increases monotonically, so checking that app-started is the first message should be good as long as seq_id = min(seq_id) (rather than 1).


andrewlock (approved):
Makes sense, LGTM

paullegranddc (approved):

paullegranddc:
LGTM too

robertomonteromiguel (approved):

PR #2540, opened 2024-06-10
Fix -F option in case of bug

cbeauchesne:
## Changes

per title

## Workflow

## Reviewer checklist

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -218,7 +218,7 @@ def pytest_collection_modifyitems(session, config, items):
             for forced in config.option.force_execute:
                 if item.nodeid.startswith(forced):
                     logger.info(f"{item.nodeid} is normally skipped, but forced thanks to -F {forced}")
-                    item.own_markers = [m for m in item.own_markers if m.name not in ("skip", "skipif")]
+                    item.own_markers = [m for m in item.own_markers if m.name not in ("skip", "skipif", "xfail")]
 
         else:
             logger.debug(f"{item.nodeid} is not included in {context.scenario}")


conftest.py
```diff
@@ -218,7 +218,7 @@ def pytest_collection_modifyitems(session, config, items):
             for forced in config.option.force_execute:
                 if item.nodeid.startswith(forced):
                     logger.info(f"{item.nodeid} is normally skipped, but forced thanks to -F {forced}")
-                    item.own_markers = [m for m in item.own_markers if m.name not in ("skip", "skipif")]
+                    item.own_markers = [m for m in item.own_markers if m.name not in ("skip", "skipif", "xfail")]
```
vpellan, line 221: This will still skip tests that we want to force. Take for instance the test `tests/appsec/test_traces.py::Test_CollectDefaultRequestHeader`
It is currently set as a missing_feature in the ruby.yml manifest. But the actual test is `tests/appsec/test_traces.py::Test_CollectDefaultRequestHeader::test_collect_default_request_headers`, with item.parent set to `tests/appsec/test_traces.py::Test_CollectDefaultRequestHeader`
So when you look at its own_markers, it is empty, but when you iterate on them using iter_markers, you can still see the missing_feature marker.
That is because it also looks at the parent's own_markers, which in this case actually contains the missing_feature marker, and it most likely also looks at the entire ancestry

conftest.py
```diff
@@ -218,7 +218,7 @@ def pytest_collection_modifyitems(session, config, items):
             for forced in config.option.force_execute:
```
vpellan, line 218:
```suggestion
            for forced in config.option.force_execute:
                current_node = item
                while current_node.nodeid.startswith(forced):
                    current_node.own_markers = [m for m in current_node.own_markers if m.name not in ("skip", "skipif", "xfail")]
                    current_node = current_node.parent
```

conftest.py
vpellan, line 221:
```suggestion
```

vpellan (changes requested):

conftest.py
cbeauchesne: This code will activate other child method, which is not what user intended to do. Though, you're right that my previous solution was missing this "nested" use case. The last commit fixes that point (and I added few tests)

PR #2552, opened 2024-06-12
asm standalone propagation downstream

gnufede:
## Motivation

Extend ASM Standalone tests with downstream propagation asserts.

## Workflow

## Reviewer checklist


tests/appsec/test_asm_standalone.py
```diff
@@ -198,11 +248,19 @@ def test_upstream_appsec_propagation__no_attack__is_propagated_as_is__being_0(se
                 ["Datadog-Client-Computed-Stats", trueish,] in data["request"]["headers"] for trueish in ["yes", "true"]
             )
 
+        downstream_headers = json.loads(self.r.text)
+        assert downstream_headers["X-Datadog-Origin"] == "rum"
+        assert downstream_headers["X-Datadog-Parent-Id"] != "34343434"
+        assert "_dd.p.other=1" not in downstream_headers["X-Datadog-Tags"]
+        assert "_dd.p.appsec=1" in downstream_headers["X-Datadog-Tags"]
+        assert downstream_headers["X-Datadog-Sampling-Priority"] == "2"
```
Julio-Guerra, line 256: This test case doesn't make sense to me because appsec only sets manual_keep, so  `"x-datadog-sampling-priority": "0"` +  `"x-datadog-tags": "_dd.p.appsec=1"` doesn't exist today, and the resulting assertion where the dowstream prio becomes 2 is undefined today.

So IIUC this test case correctly, we must remove it.

tests/appsec/test_asm_standalone.py
```diff
@@ -224,11 +282,19 @@ def test_upstream_appsec_propagation__no_attack__is_propagated_as_is__being_1(se
...
```
Julio-Guerra, line 290: same comment as above: test case doesn't match any ASM specification and is undefined today

tests/appsec/test_asm_standalone.py
```diff
@@ -276,11 +350,19 @@ def test_any_upstream_propagation__with_attack__raises_priority_to_2__from_minus
                 ["Datadog-Client-Computed-Stats", trueish,] in data["request"]["headers"] for trueish in ["yes", "true"]
             )
 
+        downstream_headers = json.loads(self.r.text)
+        assert downstream_headers["X-Datadog-Origin"] == "rum"
+        assert downstream_headers["X-Datadog-Parent-Id"] != "34343434"
+        assert "_dd.p.other=1" not in downstream_headers["X-Datadog-Tags"]
```
Julio-Guerra, line 356: I don't understand this assertion about the absence of _dd.p.other=1 while it wasn't present in the weblog request. Isn't it always necessarily true? (the comment applies to the different places where this is done)

tests/appsec/test_asm_standalone.py
```diff
@@ -302,11 +384,19 @@ def test_any_upstream_propagation__with_attack__raises_priority_to_2__from_0(sel
...
+        assert "_dd.p.appsec=1" in downstream_headers["X-Datadog-Tags"]
+        assert downstream_headers["X-Datadog-Sampling-Priority"] == "2"
+        assert downstream_headers["X-Datadog-Trace-Id"] == "1212121212121212121"
+
     def setup_any_upstream_propagation__with_attack__raises_priority_to_2__from_1(self):
         trace_id = 1212121212121212121
         parent_id = 34343434
         self.r = weblog.get(
-            "/waf/",
+            "/requestdownstream/",
```
Julio-Guerra, line 399: based on the method name, it seems like `x-datadog-sampling-priority` should be set to `1` instead of `0`

Julio-Guerra (changes requested):
Really amazing - thanks a lot <3

Aside from my inline comments, I have been wondering during the review: is python throwing an expection when `span["metrics"]["_dd.apm.enabled"]` is not present? I got afraid it could yield to `0` by default and make all the assertions `span["metrics"]["_dd.apm.enabled"] == 0` succeed by accident. Can you confirm what's the behaviour please?

robertomonteromiguel (changes requested):
There are test failures related with this PR

tests/appsec/test_asm_standalone.py
gnufede: This test case should not occur in real life as if we set "_dd.p.appsec=1" in the tags, we should also set sampling priority to 2. But if at some point in the propagation this value gets lowered, this scenario makes sure that if receiving "_dd.p.appsec=1" in tags we will reset priority to 2, and propagate that value.

tests/appsec/test_asm_standalone.py
gnufede: Yes, it's redundant

tests/appsec/test_asm_standalone.py
Julio-Guerra: is it possible to have the sampling prio lowered from 2? 🤔 I was assuming not indeed but I guess you can always set force drop in a local service but not sure what happens downstream in that case.

gnufede: yes, a force drop gets propagated the same as any other sampling priority

gnufede:
> I got afraid it could yield to `0` by default and make all the assertions `span["metrics"]["_dd.apm.enabled"] == 0` succeed by accident. Can you confirm what's the behaviour please?

It would throw the following exception, failing the test:
```
__________ Test_AppSecStandalone_UpstreamPropagation.test_no_appsec_upstream__no_attack__is_kept_with_priority_1__from_minus_1 __________

self = <tests.appsec.test_asm_standalone.Test_AppSecStandalone_UpstreamPropagation object at 0x112d17d30>

    def test_no_appsec_upstream__no_attack__is_kept_with_priority_1__from_minus_1(self):
        for data, _, span in interfaces.library.get_spans(request=self.r):
            assert span["metrics"]["_sampling_priority_v1"] < 2
>           assert span["metrics"]["_dd.apm.enabled"] == 0
E           KeyError: '_dd.apm.enabled'
```

tests/appsec/test_asm_standalone.py
gnufede: same explanation as above then

robertomonteromiguel (changes requested):
There are failures related with this PR. 
Now for "uwsgi-poc" variant and "prod"

gnufede:
> Now for "uwsgi-poc" variant and "prod"

I don't see failures related to "prod" (for this scenario), can you provide the link?

robertomonteromiguel (approved):

PR #2553, opened 2024-06-12
[.NET] enable parametric tests for remote sampling rules

lucaspimentel:
## Motivation

Our .NET SDK added support for remote sampling rules in 2.53.2, so we can enable these tests.

## Changes

Enable .NET tests for sampling rules over remote configuration (RCM).

Bonus: add support for adding span tags in .NET parametric tests (will be used in another test soon)

## Workflow

## Reviewer checklist

### DIFF
--- a/tests/parametric/test_dynamic_configuration.py
+++ b/tests/parametric/test_dynamic_configuration.py
@@ -690,6 +690,7 @@ class TestDynamicConfigSamplingRules:
         reason="JSON tag format in RC differs from the JSON tag format used in DD_TRACE_SAMPLING_RULES",
     )
     @bug(context.library == "ruby", reason="RC_SAMPLING_TAGS_RULE_RATE is not respected")
+    @bug(context.library == "dotnet", reason="Not applying correct sampling rate")
     @missing_feature(library="nodejs")
     @missing_feature(library="python")
     def test_trace_sampling_rules_with_tags(self, test_agent, test_library):
--- a/manifests/dotnet.yml
+++ b/manifests/dotnet.yml
@@ -276,7 +276,7 @@ tests/:
   parametric/:
     test_dynamic_configuration.py:
       TestDynamicConfigHeaderTags: missing_feature
-      TestDynamicConfigSamplingRules: v2.54.0
+      TestDynamicConfigSamplingRules: v2.53.2
       TestDynamicConfigTracingEnabled: v2.49.0
       TestDynamicConfigV1: v2.33.0
       TestDynamicConfigV1_ServiceTargets: missing_feature
--- a/manifests/dotnet.yml
+++ b/manifests/dotnet.yml
@@ -276,7 +276,7 @@ tests/:
   parametric/:
     test_dynamic_configuration.py:
       TestDynamicConfigHeaderTags: missing_feature
-      TestDynamicConfigSamplingRules: missing_feature
+      TestDynamicConfigSamplingRules: v2.54.0
       TestDynamicConfigTracingEnabled: v2.49.0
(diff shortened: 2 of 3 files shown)


tests/parametric/test_dynamic_configuration.py
```diff
@@ -690,6 +690,7 @@ def test_trace_sampling_rules_override_rate(self, library_env, test_agent, test_
         reason="JSON tag format in RC differs from the JSON tag format used in DD_TRACE_SAMPLING_RULES",
     )
     @bug(context.library == "ruby", reason="RC_SAMPLING_TAGS_RULE_RATE is not respected")
+    @bug(context.library == "dotnet", reason="Not applying correct sampling rate")
```
zacharycmontoya, line 693: Can we remove this line since you've updated the dotnet manifest file?

lucaspimentel: Not yet 😞  In the manifest I enabled all the tests in class `TestDynamicConfigSamplingRules`, but this line still skips one of them because it is failing, `TestDynamicConfigSamplingRules::test_trace_sampling_rules_with_tags`.

I wanted to enable the rest of the tests while I investigate this particular test failure. It fails to match remote sampling rules with tags, which is still not supported by the frontend/backend, so not an urgent blocker.

lucaspimentel: Can we do this in the manifest file?

zacharycmontoya: Hmm. In that case I recommend doing something like:
```suggestion
    @bug(context.library <= "dotnet@2.53.2", reason="Not applying correct sampling rate")
```

robertomonteromiguel (changes requested):
There are test failures related with this PR

lucaspimentel:
> There are test failures related with this PR

@robertomonteromiguel, yesterday we released [v2.53.2](https://github.com/DataDog/dd-trace-dotnet/releases/tag/v2.53.2) of the .NET SDK and the test are passing now. I'm not sure why the "dev" tests didn't pass before since the code was already merged into `master`.

tests/parametric/test_dynamic_configuration.py
lucaspimentel: `context.library <= "dotnet@2.53.2"` means we fixed this test in `2.53.2`, but we haven't fixed it yet. The _other_ tests in that class start passing with `2.53.2`. This one test still fails and requires investigation.

lucaspimentel: Effectively, what I want to do is something like this in the manifest file (but I don't think this is valid?)
```yaml
tests/:
  parametric/:
    test_dynamic_configuration.py:
      TestDynamicConfigSamplingRules:
        test_trace_sampling_rules_with_tags: bug
        test_trace_sampling_rules_override_env: v2.53.2
        test_trace_sampling_rules_override_rate: v2.53.2
        test_remote_sampling_rules_retention: v2.53.2
```

robertomonteromiguel (approved):

tests/parametric/test_dynamic_configuration.py
lucaspimentel: Two things are needed to enable this one test:
- Sampling rules were not matching by tags because we were not adding tags at all in .NET. In this PR I add support in the .NET driver app used for parametric tests.
- All local sampling rules should be ignored is there are any remote rules present (even if none of them match the tracer). I have a fix for this here:
https://github.com/DataDog/dd-trace-dotnet/pull/5720
I will enable this test when that PR is merged.

PR #2597, opened 2024-06-24
enable otel env telemetry system tests

khanayan123:
## Motivation

Enable otel env var telemetry system tests for Node.js

## Changes

Enables otel env var telemetry system tests for Node.js


zacharycmontoya (approved):
LGTM

PR #2599, opened 2024-06-25
Use latest python-black to format the sources

pawelchcki:
Unpin the old version *19.x and use latest 24.x - which enables new checks and format options


cbeauchesne:
Conflict war incoming ! :D

cbeauchesne:
We switched to ruff a while ago

PR #2614, opened 2024-06-26
[nodejs] enable xpassed tests

ida613:

## Reviewer checklist


tests/test_sampling_rates.py
```diff
@@ -66,7 +66,6 @@ def setup_sampling_rates(self):
         context.library > "nodejs@3.14.1" and context.library < "nodejs@4.8.0",
         reason="_sampling_priority_v1 is missing",
     )
-    @bug(library="nodejs", reason="Unexpected amount of sampled traces")
```
simon-id, line 69: would be great to know when this was fixed to put a filter instead of removing the whole skip

tests/parametric/test_sampling_span_tags.py
```diff
@@ -122,7 +122,6 @@ def test_tags_child_kept_sst007(self, test_agent, test_library):
     @bug(library="dotnet", reason="dotnet does not set dm tag on first span")
     @bug(library="cpp", reason="unknown")
     @bug(library="php", reason="php does not set agent rate tag")
-    @bug(library="nodejs", reason="nodejs sets dm tag -0")
```
simon-id, line 125: same thing here

manifests/nodejs.yml
simon-id, line 1: did you make sure to run the system tests on the old versions you've added ? to check the tests are actually passing in those old versions

ida613: yup

ida613: CI ran after I modified the versions, that's all we need right?

tests/test_sampling_rates.py
ida613: what do you mean by "to put a filter"?

manifests/nodejs.yml
simon-id: No, the CI only run on latest main and latest npm release. So it's not verifying the tests actual pass on those old versions you've put.

tests/test_sampling_rates.py
simon-id: Look at the line just above, something like this:
`@bug(context.library < "nodejs@1.2.3"`)

ida613: Our current goal is to get as many tests to run on dd-trace-js as possible, so don't have time to dig into code archeology at the moment. Would it be okay to leave it for now?

manifests/nodejs.yml
ida613: again don't have time for this at the moment. checked with @bm1549 that we don't really care about starting from which versions of the tracer that the tests started passing, so could we just put down the latest version for everything?

bm1549: I'd vote for latest version - no need for code archeology here

robertomonteromiguel (approved):

manifests/nodejs.yml
simon-id: It's far from code archeology, it's doing a git checkout of the version you're putting and running the tests. Should take 5 mins basically

simon-id: And if you put the current latest, put a comment saying it's inaccurate

bm1549: Agreed that, if we put a version, we should confirm that version passes tests

Happy to have us put a code comment saying we suspect it's an older version, but haven't tested it

simon-id: One last note: one of the selling point of the system-test+manifest file+feature parity dashboard, is to give the release version of any feature through the different libraries, as a source of truth. If you start putting inaccurate versions in the manifest then that defeats one of the use case.
At the end of the day, you do you, it ain't my turf, but it's not that much work to do this PR properly.

tests/test_sampling_rates.py
cbeauchesne: +1 @simon-id having an "at least" version is needed. Without that, we won't be able later to run system-tests on old versions. So far, I never dig to find the real version, I always used the current prod version.

simon-id (changes requested):
Thanks for updating the version on the skip decorators, but the new versions in the manifest are also old and unverified. Please either run these tests on the specified versions of the library to make sure they pass on it, or set the released versions as the latest and add a comment saying the actual version is unknown

ida613:
> Thanks for updating the version on the skip decorators, but the new versions in the manifest are also old and unverified. Please either run these tests on the specified versions of the library to make sure they pass on it, or set the released versions as the latest and add a comment saying the actual version is unknown

tested the older versions

simon-id (approved):

PR #2623, opened 2024-06-27
fix incorrect ref to determine when to push latest tag

rochdev:
## Motivation

The branch referenced was `master` but the main branch is `main`.

## Changes

Fix incorrect ref to determine when to push latest tag.

## Workflow

## Reviewer checklist

### DIFF
--- a/.github/workflows/run-docker-mode.yml
+++ b/.github/workflows/run-docker-mode.yml
@@ -67,11 +67,11 @@ jobs:
         run: |
           docker push ${{ env.REPO }}/docker/${{ matrix.image.name }}:g${{ github.sha }}
       - name: Tag image for release
-        if: ${{ github.ref == 'refs/heads/master' }}
+        if: ${{ github.ref == 'refs/heads/master' || github.ref == 'refs/heads/main' }}
         run:
           docker tag ${{ matrix.image.internal }} ${{ env.REPO }}/docker/${{ matrix.image.name }}:latest
       - name: Push image for release
-        if: ${{ github.ref == 'refs/heads/master' }}
+        if: ${{ github.ref == 'refs/heads/master' || github.ref == 'refs/heads/main' }}
         run: |
           docker push ${{ env.REPO }}/docker/${{ matrix.image.name }}:latest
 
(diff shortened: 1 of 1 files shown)


cbeauchesne (commented):
Why keeping master in the `if` ?

rochdev:
> Why keeping master in the if ?

My understanding is that this can be used (and I think it is by Ruby?) in other repositories which may not have transitioned their main branch.

cbeauchesne:
IIUC, this checks is only about the system-tests branch. This yml file is not called anywhere else. But as this way of running is not supported (it was only something ruby team wanted, and they wrote it with the agrement of best-effort on our side), I must admit I'm not very familiar with it.

rochdev:
> But as this way of running is not supported (it was only something ruby team wanted, and they wrote it with the agrement of best-effort on our side)

Why is it not supported? Node wants this too, and I think it makes perfect sense to build this once and then use it everywhere instead of rebuilding it on every single commit of every library that uses system tests.

cbeauchesne:
> Why is it not supported? 

Because we don't have the dev bandwidth to support two different runner mechanisms.

> Node wants this too, and I think it makes perfect sense to build this once and then use it everywhere instead of rebuilding it on every single commit of every library that uses system tests.

The image that takes most of the time to build is the weblog, and you'll need to build it on each commit. The present mode offers four features : 

1. run the runner inside docker <- this is not officially supported, only best effort if there is some issue
2. build weblog image only once, then run in parallel all scenarios <- you may be interested by this one. It won't reduce CI time, (actually increase it a little bit because you need to store then load a docker image), but may reduces flakyness, and reduce CI cost.
3. build proxy images only once, same remarks as above.
4. build  agent images only once, same remarks as above.

As said, point 1 => not supported, and IIUC, it's not what you're interested in.

The point 2 is something useful, at least as an option in `.github/workflows/system-tests.yml` I want to do one day, but never did, by lack of time, and also because it does not offers better CI time.

The point 3 is definitly a win, and an easy one

The point 4 is also a win, but more difficult, because we always want to have the latest agent image. A good way to do so would be to add `ca-certificates` on this image (I don't know if agent team would agree), then use it directly without any need to build another one.

And last, if you want to implements point 2, 3, and 4 directly, you can do it directly in your CI. Actually ruby team implemented this in their CI, they do not even use this reusable workflow, which is present only to be sure that we don't break docker mode.

cbeauchesne (approved):

PR #2646, opened 2024-07-01
remove process_ignore

raphaelgavache:
Removing the process ignore option. With the new self-inject approach https://github.com/DataDog/auto_inject/pull/266 this option is not relevant anymore. The accessible cmdline is now straight the interpreter /.../bin/(java|python|ruby|dotnet..). 

Starting with the test to pass CI, and then will remove in auto_inject

### DIFF
--- a/tests/auto_inject/test_auto_inject_blocklist.py
+++ b/tests/auto_inject/test_auto_inject_blocklist.py
@@ -233,25 +233,6 @@ class TestAutoInjectBlockListInstallManualHost(_AutoInjectBlockListBaseTest):
         ],
     }
 
-    user_processes_commands = [
-        {
-            "ignored_processes": "/opt/datadog/logs_injection/*,other",
-            "command": "/opt/datadog/logs_injection/myscript.sh",
-            "skip": True,
-        },
-        {"ignored_processes": "**/myscript.sh", "command": "/opt/datadog/logs_injection/myscript.sh", "skip": True},
-        {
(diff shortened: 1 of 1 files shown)


tests/auto_inject/test_auto_inject_blocklist.py
```diff
@@ -303,33 +303,6 @@ def _create_remote_executable_script(self, ssh_client, script_path, content="#!/
...
-        reason="Block list not fully implemented ",
-    )
-    def test_user_ignored_processes(self, virtual_machine):
-        """ Check that we are not instrumenting the commands that match with patterns set by DD_IGNORED_PROCESSES env variable"""
-
-        ssh_client = virtual_machine.ssh_config.get_ssh_connection()
-        # Create a simple executable script
-        self._create_remote_executable_script(ssh_client, "/opt/datadog/logs_injection/myscript.sh")
-
-        for test_config in self.user_processes_commands:
```
robertomonteromiguel, line 318: if we remove this test, we should remove the variable "user_processes_commands"

robertomonteromiguel (changes requested):
if we remove this test, should we add a new test case testing the new approach?

raphaelgavache:
We're removing the customer configuration that this test asserts so there's no additional test to add.
The new injector will target straight the different language interpreters which shouldn't be blocked. 
DD_IGNORED_PROCESSES is not publicly documented

robertomonteromiguel (approved):

PR #2663, opened 2024-07-03
[Debugger] Expression Language - comparison operators

shurivich:
## Motivation
Test expression language comparison operators

### DIFF
--- a/tests/debugger/probes/expression_language_comparison_operators.json
+++ b/tests/debugger/probes/expression_language_comparison_operators.json
@@ -1,0 +1,418 @@
+[
+    {
+        "language": "",
+        "id": "log170aa-expr-lang-0001-1479a6method",
+        "where": {
+            "typeName": "ACTUAL_TYPE_NAME",
+            "methodName": "ExpressionComparisonOperators",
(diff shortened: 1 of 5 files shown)


tests/debugger/probes/expression_language_comparison_operators.json
dudikeleti, line 1: Do we want to test also `!` or `not` ?
Also I'm not sure if this is supported in the final version of the RFC, but I see that we have two version of each equality operator, for example `eq` => `--`, `gt` => `>` etc etc. Do we want to test it as well?

dudikeleti (approved):
LGTM. Left a comment.

jpbempel (approved):
LGTM
but we are testing only integers here.
What is missing:
- floats
- strings
- objects, `null`

tests/debugger/probes/expression_language_comparison_operators.json
shurivich: Good point! If we support it - we should. 
I will add it to upcoming PRs.

robertomonteromiguel (approved):

PR #2670, opened 2024-07-03
Run DB startup in parallel

cbeauchesne:
## Workflow

## Reviewer checklist

### DIFF
--- a/.github/actions/pull_images/action.yml
+++ b/.github/actions/pull_images/action.yml
@@ -35,4 +35,4 @@ runs:
 
     - name: Pull
       shell: bash
-      run: docker compose pull
+      run: docker compose pull -q
--- a/.github/actions/pull_images/action.yml
+++ b/.github/actions/pull_images/action.yml
@@ -9,22 +9,6 @@ inputs:
 runs:
   using: composite
   steps:
-    - name: Free some place for scenarios known to require lot of images
-      if: ${{ contains(inputs.scenarios, '"INTEGRATIONS"') || contains(inputs.scenarios, '"CROSSED_TRACING_LIBRARIES"') }}
-      shell: bash
-      run: |
-        df -h
-
-        echo "Removing docker images, dotnet, android, ghc, and CodeQL cache to free space"
-
-        sudo rm -rf /usr/local/lib/android  # 9Gb!
-        sudo rm -rf /opt/hostedtoolcache/CodeQL  # 5Gb !
-        sudo rm -rf /usr/share/dotnet # 1Gb
-
-        # sudo docker image prune --all --force  # if ever needed, but it's slow. Only 3Gb
-
-        df -h
-
     - name: Get image list
       shell: bash
       run: |
--- a/.shellcheckrc
+++ b/.shellcheckrc
@@ -2,4 +2,4 @@
 #
 # SC1091: Not following <file>
 # https://github.com/koalaman/shellcheck/wiki/SC1091
-# disable=SC1091
+disable=SC1091
--- a/pull.sh
+++ b/pull.sh
@@ -3,6 +3,6 @@
 export PYTHONPATH='.'
 
 source venv/bin/activate
-python utils/scripts/get-image-list.py '"'$1'"' > compose.yaml
+python utils/scripts/get-image-list.py '"'"$1"'"' > compose.yaml
 
 docker compose pull
--- a/pull.sh
+++ b/pull.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+export PYTHONPATH='.'
+
+source venv/bin/activate
+python utils/scripts/get-image-list.py '"'$1'"' > compose.yaml
+
+docker compose pull
--- a/utils/_context/_scenarios/__init__.py
+++ b/utils/_context/_scenarios/__init__.py
@@ -173,20 +173,21 @@ class _DockerScenario(_Scenario):
 
         if not self.replay:
             warmups.append(create_network)
-            warmups.append(self._start_containers)
+
+        warmups.append(self._start_containers)
 
         return warmups
 
     def _start_containers(self):
-        required_thread = Thread(target=start_parallel_containers, args=(self._external_containers,))
+        required_thread = Thread(target=start_parallel_containers, args=(self._external_containers, self.replay,))
         required_thread.start()
-        external_thread = Thread(target=start_sequential_containers, args=(self._required_containers,))
+        external_thread = Thread(target=start_sequential_containers, args=(self._required_containers, self.replay))
         external_thread.start()
 
         required_thread.join()
         external_thread.join()
 
-        start_parallel_containers(self._internal_containers)
+        start_parallel_containers(self._internal_containers, self.replay)
 
     # TODO: remove in parallel
     def close_targets(self):
--- a/utils/_context/_scenarios/__init__.py
+++ b/utils/_context/_scenarios/__init__.py
@@ -189,7 +189,7 @@ class _DockerScenario(_Scenario):
 
         start_parallel_containers(self._internal_containers, self.replay)
 
-    # TODO: remove in parallel
+    # TODO: remove containers in parallel
     def close_targets(self):
         for container in reversed(self.containers):
             try:
--- a/utils/_context/_scenarios/__init__.py
+++ b/utils/_context/_scenarios/__init__.py
@@ -176,18 +176,18 @@ class _DockerScenario(_Scenario):
             warmups.append(self._start_containers)
 
         return warmups
-    
+
     def _start_containers(self):
-        required_thread = Thread(target = start_parallel_containers, args = (self._external_containers,))
+        required_thread = Thread(target=start_parallel_containers, args=(self._external_containers,))
         required_thread.start()
-        external_thread = Thread(target = start_sequential_containers, args = (self._required_containers,))
+        external_thread = Thread(target=start_sequential_containers, args=(self._required_containers,))
         external_thread.start()
 
         required_thread.join()
         external_thread.join()
 
         start_parallel_containers(self._internal_containers)
-    
+
     # TODO: remove in parallel
     def close_targets(self):
         for container in reversed(self.containers):
--- a/utils/_context/_scenarios/__init__.py
+++ b/utils/_context/_scenarios/__init__.py
@@ -1,6 +1,7 @@
 import os
 import time
 import json
+from threading import Thread
 
 import pytest
 from watchdog.observers.polling import PollingObserver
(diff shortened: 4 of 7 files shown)


rochdev (approved):

.github/actions/pull_images/action.yml
```diff
@@ -9,22 +9,6 @@ inputs:
 runs:
   using: composite
   steps:
-    - name: Free some place for scenarios known to require lot of images
```
cbeauchesne, line 12: Did you checked that it was not needed anymore? we add by the past issues because the filesystem was close to be full, with at best huge increase in execution time, and worst, cryptic failures.

.github/actions/pull_images/action.yml
```diff
@@ -35,4 +19,4 @@ runs:
 
     - name: Pull
       shell: bash
-      run: docker compose pull
+      run: docker compose pull -q
```
cbeauchesne, line 22: why the `-q` ?

.shellcheckrc
```diff
@@ -2,4 +2,4 @@
 #
 # SC1091: Not following <file>
 # https://github.com/koalaman/shellcheck/wiki/SC1091
-# disable=SC1091
+disable=SC1091
```
cbeauchesne, line 5: Why disabling this check ?

pull.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
cbeauchesne, line 1: if it's at the root of the folder (which I think is a good idea), it must be a little more polished.

1. set -e
2. add an helper, with a `-h` option
3. must work without any argument (in that case, I would say print a confirmation message, then pull all images)
4. a better CLI : `./pull.sh -S NAME1,NAME2,NAME3` 

And a lazy, but totally acceptable at a first step, is to put it in `utils/scripts/`

Side question, why you didn't used it in the action ?

utils/_context/_scenarios/__init__.py
```diff
@@ -106,6 +109,8 @@ def __init__(
             raise ValueError("rc_api_enabled requires use_proxy")
 
         self._required_containers: list[TestedContainer] = []
+        self._external_containers: list[TestedContainer] = []
```
cbeauchesne, line 112: Those names does not make very clear the intention of those lists. On the other hands, I see some issue. so let's fix them before finding good names.

utils/_context/_scenarios/__init__.py
```diff
@@ -165,16 +174,24 @@ def _get_warmups(self):
...
-        for container in self._required_containers:
-            warmups.append(container.post_start)
+        warmups.append(self._start_containers)
 
         return warmups
 
+    def _start_containers(self):
+        required_thread = Thread(target=start_parallel_containers, args=(self._external_containers, self.replay,))
+        required_thread.start()
+        external_thread = Thread(target=start_sequential_containers, args=(self._required_containers, self.replay))
```
cbeauchesne, line 184: This code my have the weblog started before DB are up and healthy, some weblog interacts with DB at startup, so they must be healthy before we start weblog.

The more. I see this, the more I think we should introduce the notion of `need` in `TestedContainer`, create a compose.yaml, and let `docker compose` do the job. WDYT?

.github/actions/pull_images/action.yml
rochdev: I didn't see any issues that I can tell from removing this. I wouldn't expect a problem even with several Docker images as the images we're using are not small by any means but they're not that big either. If this is ever a problem again, I'd look into fix it in other ways as this step is extremely slow and should be avoided if at all possible.

.github/actions/pull_images/action.yml
rochdev: The output of the command unfortunately prints progress very quickly instead of in stages, which ends up filling up the log with tons and tons of lines.

.shellcheckrc
rochdev: The check is to validate that something exists on the filesystem before sourcing it, but this project by nature sources things that are dynamically generated.

pull.sh
rochdev: To be honest I just made this script by quickly extracting the logic from the action. At this point I'd rather just remove it from this PR and reintroduce it as its own PR instead.

utils/_context/_scenarios/__init__.py
rochdev: The idea was basically:

- **required**: this was the existing name and I think it still makes sense as these are the images that _must_ exist. Another name could be "system containers" maybe.
- **external**: this is basically any image that is used by tests but not owned by us, for example databases, message queues, etc. I also thought about calling this "service containers".
- **internal**: these are the containers used by the tests. I called them internal because they are created by this library, this they are not from external sources.

Basically, the distinction is that required containers are the containers for the test framework which have to be started sequentially, external containers are external service containers that can be started in parallel since they are unrelated, and internal containers are the containers for the tests that use both the required and external containers.

utils/_context/_scenarios/__init__.py
rochdev: > This code my have the weblog started before DB are up and healthy, some weblog interacts with DB at startup, so they must be healthy before we start weblog.

My main issue is that I'm not fully aware of what everything is doing and the actual interdependencies. I tried to separate things into 3 lists but that may have been incorrect or insufficient.

> The more. I see this, the more I think we should introduce the notion of need in TestedContainer, create a compose.yaml, and let docker compose do the job. WDYT?

I wish it was that simple, but unfortunately `docker compose` is unable to handle this. When using links or `depends_on`, it only makes sure to start the containers in order, but it doesn't have the notion of waiting for health checks, that is left to the applications. I usually agree with this design as applications should always be built with no specific expectation of ordering, but in this case that doesn't quite work as there is a lot of test setup involve that the applications don't know about.

With that said, I agree with the sentiment, and I was trying at first to come up with some sort of `ContainerGroup` that would allow defining containers that can be started in parallel and with a dependency system, but it ended up with too many code changes and I don't know Python well enough to make that big a change. I'll try to see if I can do something like this but while keeping changes minimum.

rochdev: Actually, would it be enough to just add `weblog` to the internal containers instead? I would expect weblog and the buddies to be able to be started in parallel since they may talk to each other and they all need to be healthy for requests to go through properly so there shouldn't be any ordering dependencies between them.

.github/actions/pull_images/action.yml
cbeauchesne: So, first thing first.

This change is not related to "Run DB startup in parallel". Can you create another PR? it will ease the process for this current PR.

Secondly, this step is not here for nothing, we had a hard time to fix that when it happened, docker errors are cryptic when disk is close to be full. I'm totally supportive to remove it if it's not necessary, but we won't go blindly into that direction. So could you :

1. get the time we save by removing it (maybe one of the three `rm` is taking most of the time?)
2. get the remaining space after all scenario have ran on each weblog before and after the change (well, only the one with the less space is important, but I don't know which weblog take the more space)

Based on that, we'll be able to have an good decision. And if it's a go, we'll keep the code commented, and raise the attention of the R&P team, because if it happen, we must act quickly.

.shellcheckrc
cbeauchesne: But it was passing before, no? Is there something in this PR that was not passing ?

utils/_context/_scenarios/__init__.py
cbeauchesne: Ok I understand. Though, I think it's not a good choice. Even if this naming describes a reality, this reality is not what justifies the presence of these three different lists. The real purpose of having three lists is the startup sequence. 

Let say one day somebody adds an "external" container, but for whatever the reason, it needs the proxy. He'll need to find another naming (after having a hard time to understand why it's not working).

The naming must reflect that startup order (and comments  must make obvious what a short name can't explain).

utils/_context/_scenarios/__init__.py
cbeauchesne: > My main issue is that I'm not fully aware of what everything is doing and the actual interdependencies. I tried to separate things into 3 lists but that may have been incorrect or insufficient.

Let me give you some help on that, I'll write a small doc asap.

> I wish it was that simple, but unfortunately docker compose is unable to handle this.

Yes it can, you have a `condition: service_healthy` option in `depends_on` parameter. And on purpose, I kept all the healthcheck definition in the format of docker compose, because I knew one day we'll need to go into that direction.

>  I would expect weblog and the buddies to be able to be started in parallel since they may talk to each other and they all need to be healthy for requests to go through properly so there shouldn't be any ordering dependencies between them.

Agree.

utils/_context/_scenarios/__init__.py
rochdev: An external container by nature would never need the proxy since the proxy is internal so an external container would never know about it, hence the way it's split.

utils/_context/_scenarios/__init__.py
rochdev: I didn't know about `service_healthy`, is it compatible with the health checks that are used in this project?

utils/_context/_scenarios/__init__.py
cbeauchesne: > An external container by nature would never need the proxy since the proxy is internal so an external container would never know about it

I disagree, we don't know what will be the future. We can totally imagine an external container sending telemetry metrics we'd like to observe. But whatever, it's only an example, it  changes nothing to the underlying reason : the only reason for those list is the startup order, this point must be reflected in their names.

utils/_context/_scenarios/__init__.py
cbeauchesne: > I didn't know about service_healthy

https://docs.docker.com/compose/startup-order/

> is it compatible with the health checks that are used in this project?

As said, I kept healthcheck definitions in the exact format of docker compose healthchecks : [Example](https://github.com/DataDog/system-tests/blob/main/utils/_context/containers.py#L634)

cbeauchesne: > Let me give you some help on that, I'll write a small doc asap.

https://github.com/DataDog/system-tests/blob/main/docs/internals/end-to-end-life-cycle.md

rochdev: Any pointers on how I could switch this to Docker Compose?

cbeauchesne: 1. Add an optional `need:list[TestedContainer]` parameter to `TestedContainer`'s `__init__`. If not filled, it's initialized with an empty list.
2. add a method on `Scenario` class that build a  `compose.yml` base on `required_container`, with the good `need` value
3. on scenario warmup, spawn a blocking process, calling `docker compose` with the result of the function above as argument.
4. if it fails, print stdout/stderr

cbeauchesne: And as the current code adds container in startup order, it'll require a small revamp to change this order, as weblog will mostly be in first.

cbeauchesne:
replaced by #2732

PR #2720, opened 2024-07-10
make disk cleanup optional when pulling images

rochdev:
## Motivation

The cleanup step takes most of the time of the job and is not needed when using larger runners which have more disk space ([600GB](https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners/about-larger-runners#specifications-for-general-larger-runners) vs [14GB](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories)).

With the cleanup, the `pull_images` job takes 5-15 minutes for `INTEGRATIONS`. Without the cleanup, it takes ~30 seconds.

## Changes

Make disk cleanup optional when pulling images.

## Workflow

## Reviewer checklist

### DIFF
--- a/.github/actions/pull_images/action.yml
+++ b/.github/actions/pull_images/action.yml
@@ -20,7 +20,7 @@ runs:
   using: composite
   steps:
     - name: Free some place for scenarios known to require lot of images
-      if: ${{ inputs.cleanup && (contains(inputs.scenarios, '"INTEGRATIONS"') || contains(inputs.scenarios, '"CROSSED_TRACING_LIBRARIES"')) }}
+      if: ${{ inputs.cleanup == true && (contains(inputs.scenarios, '"INTEGRATIONS"') || contains(inputs.scenarios, '"CROSSED_TRACING_LIBRARIES"')) }}
       shell: bash
       run: |
         df -h
--- a/.github/actions/pull_images/action.yml
+++ b/.github/actions/pull_images/action.yml
@@ -20,7 +20,7 @@ runs:
   using: composite
   steps:
     - name: Free some place for scenarios known to require lot of images
-      if: ${{ inputs.cleanup == true && (contains(inputs.scenarios, '"INTEGRATIONS"') || contains(inputs.scenarios, '"CROSSED_TRACING_LIBRARIES"')) }}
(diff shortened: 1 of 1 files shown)


cbeauchesne (changes requested):
with the default value set to true, it'll be uses on this present CI, which cannot be made with a little bit more visibility : 

* either we continue using default runner => get the current left space at the end of end-to-end process for all weblogs, to see if we are close to the limit or not
* or we use large runner => what are the avaibility of those large runner ? we spawn thousand of jobs every day, and we already have some wait time for a runner to be present. We must not loose here what we gain by removing the `rm`. And I also don't want to exhaust the DD large runners grid, just to save 2mn of CI time.

Another easy solution is to set the default value to `False`, or to add `cleanup=False` in this CI.

But, actually, I'd see a better solution :  be to just get the free space on the disk, and if it's above a limit (let says 20Gb), run the cleanup : 

- no downside
- no new inputs
- no cognitive effort for user 

WDYT ?

rochdev:
> with the default value set to true, it'll be uses on this present CI, which cannot be made with a little bit more visibility

It's the reverse. A value of `true` is equivalent to today's behaviour, so there won't be any change on the current CI. It has to be set to `false` to trigger the change.

> But, actually, I'd see a better solution : be to just get the free space on the disk, and if it's above a limit (let says 20Gb), run the cleanup :

I prefer to be explicit, otherwise builds won't be deterministic as GitHub only guarantees a _minimum_ size, meaning that a rerun could get a machine with more or less disk space.

cbeauchesne:
But it's currently not executed ? -> https://github.com/DataDog/system-tests/actions/runs/9880334639/job/27336822494?pr=2720

Or did I missed something ? 

I prefer to be explicit, otherwise builds won't be deterministic as GitHub only guarantees a minimum size, meaning that a rerun could get a machine with more or less disk space.

Agree

rochdev:
> But it's currently not executed ?

Hmm let me look into it. Originally the condition didn't work so I changed it and assumed it was now working, but it may still be broken in the other way where it just never runs regardless of the config.

rochdev:
Looks like default values are only supported for strings. I can either switch to a string or change the name to something like `skip-cleanup` which will then automatically default to `false` without having to be explicit.

cbeauchesne (approved):

PR #2732, opened 2024-07-11
start containers in parallel

rochdev:
## Motivation

Containers that don't depend on each other shouldn't have to wait for each other before starting.

## Changes

Start containers in parallel. This PR is a replacement for #2670 with less changes and more flexible code to manage interdependencies.

## Workflow

## Reviewer checklist

### DIFF
--- a/utils/_context/_scenarios/core.py
+++ b/utils/_context/_scenarios/core.py
@@ -3,6 +3,7 @@ from logging import FileHandler
 import os
 from pathlib import Path
 import shutil
+from threading import Thread
 
 import pytest
 from watchdog.observers.polling import PollingObserver
@@ -25,6 +26,7 @@ from utils._context.containers import (
     create_network,
     BuddyContainer,
     TestedContainer,
+    start_container,
 )
 
 from utils._context.library_version import LibraryVersion
@@ -309,14 +311,25 @@ class DockerScenario(Scenario):
 
         if not self.replay:
             warmups.append(create_network)
-
-            for container in self._required_containers:
-                warmups.append(container.start)
-
-        for container in self._required_containers:
-            warmups.append(container.post_start)
+            
+        warmups.append(self._start_containers)
 
         return warmups
+    
+    def _start_containers(self):
+        threads = {}
+
+        for container in self._required_containers:
+            for name in container.depends_on:
+                if name in threads:
+                    threads[name].join()
+
+            thread = Thread(target=start_container, args=(container, self.replay,))
+            thread.start()
+            threads[container.name] = thread
+
+        for name in threads:
+            threads[name].join()
 
     def close_targets(self):
         for container in reversed(self._required_containers):
--- a/utils/_context/containers.py
+++ b/utils/_context/containers.py
@@ -63,6 +63,13 @@ def create_inject_volume():
     _get_client().volumes.create(_VOLUME_INJECTOR_NAME)
 
 
+def start_container(container, replay=False):
+    if not replay:
+        container.start()
+
+    container.post_start()
+
+
 class TestedContainer:
 
     # https://docker-py.readthedocs.io/en/stable/containers.html
@@ -75,6 +82,7 @@ class TestedContainer:
         allow_old_container=False,
         healthcheck=None,
         stdout_interface=None,
+        depends_on=[],
         **kwargs,
     ) -> None:
         self.name = name
@@ -89,6 +97,7 @@ class TestedContainer:
         self.kwargs = kwargs
         self._container = None
         self.stdout_interface = stdout_interface
+        self.depends_on = depends_on
 
     def get_image_list(self, library: str, weblog: str) -> list[str]:
         """ returns the image list that will be loaded to be able to run/build the container """
@@ -401,6 +410,7 @@ class AgentContainer(TestedContainer):
             },
             ports={self.agent_port: f"{self.agent_port}/tcp"},
             stdout_interface=interfaces.agent_stdout,
+            depends_on=["proxy"],
         )
 
         self.agent_version = None
@@ -493,6 +503,23 @@ class WeblogContainer(TestedContainer):
             healthcheck={"test": f"curl --fail --silent --show-error localhost:{self.port}", "retries": 60},
             ports={"7777/tcp": self.port, "7778/tcp": weblog._grpc_port},
             stdout_interface=interfaces.library_stdout,
+            depends_on=[
+                "agent",
+                "cassandra_db",
+                "elasticmq",
+                "go_buddy",
+                "java_buddy",
+                "kafka",
+                "localstack-main",
+                "mongodb",
+                "mssql",
+                "mysqldb",
+                "nodejs_buddy",
+                "postgres",
+                "python_buddy",
+                "rabbitmq",
+                "ruby_buddy",
+            ],
         )
 
         self.tracer_sampling_rate = tracer_sampling_rate
@@ -684,6 +711,7 @@ class KafkaContainer(TestedContainer):
                 "timeout": 2 * 1_000_000_000,
                 "retries": 25,
             },
+            depends_on=["zookeeper"]
         )
 
     def warmup(self):
--- a/utils/_context/_scenarios/__init__.py
+++ b/utils/_context/_scenarios/__init__.py
@@ -3,6 +3,7 @@ import json
 
 import pytest
 from utils._context.library_version import LibraryVersion
+from threading import Thread
 
 from utils._context.header_tag_vars import VALID_CONFIGS, INVALID_CONFIGS
 
--- a/utils/_context/_scenarios/__init__.py
+++ b/utils/_context/_scenarios/__init__.py
@@ -3,7 +3,6 @@ import json
 
 import pytest
 from utils._context.library_version import LibraryVersion
-from threading import Thread
 
 from utils._context.header_tag_vars import VALID_CONFIGS, INVALID_CONFIGS
 


utils/_context/_scenarios/core.py
```diff
@@ -310,13 +312,24 @@ def _get_warmups(self):
         if not self.replay:
             warmups.append(create_network)
 
-            for container in self._required_containers:
-                warmups.append(container.start)
+        warmups.append(self._start_containers)
```
cbeauchesne, line 315: Containers must not be started in replay mode

utils/_context/_scenarios/core.py
```diff
@@ -310,13 +312,24 @@ def _get_warmups(self):
...
+
+        return warmups
+
+    def _start_containers(self):
+        threads = {}
 
         for container in self._required_containers:
-            warmups.append(container.post_start)
+            for name in container.depends_on:
```
cbeauchesne, line 323: Let says we have this list for _required_containers: 

```
A (depends_on B)
B
```

In the loop, A is processed. `threads["B"]` does not exists, so A is started even if it depends on B -> race condition.

utils/_context/containers.py
```diff
@@ -75,6 +82,7 @@ def __init__(
         allow_old_container=False,
         healthcheck=None,
         stdout_interface=None,
+        depends_on=[],
```
cbeauchesne, line 85: Do not use mutable values for default args

```suggestion
        depends_on=None,
```

utils/_context/containers.py
```diff
@@ -89,6 +97,7 @@ def __init__(
         self.kwargs = kwargs
         self._container = None
         self.stdout_interface = stdout_interface
+        self.depends_on = depends_on
```
cbeauchesne, line 100:
```suggestion
        self.depends_on = [] if depends_on is None else depends_on
```

utils/_context/containers.py
```diff
@@ -401,6 +410,7 @@ def __init__(self, host_log_folder, use_proxy=True) -> None:
             },
             ports={self.agent_port: f"{self.agent_port}/tcp"},
             stdout_interface=interfaces.agent_stdout,
+            depends_on=["proxy"],
```
cbeauchesne, line 413: Could you make `depends_on` a list of TestedContainer, rather than a list of str? spreading container name in the code will make it harder to maintain.

cbeauchesne (changes requested):
If I'm not wrong, buddies also require the proxy (and possibly the agent?)

cbeauchesne:
> This PR is a replacement for https://github.com/DataDog/system-tests/pull/2670 with less changes and more flexible code to manage interdependencies.

Can I close this PR ?

utils/_context/_scenarios/core.py
rochdev: Yes, the order of the list is still important with the new logic, but that was already the case before.

utils/_context/_scenarios/core.py
rochdev: They aren't, there is a check in `start_container`.

utils/_context/_scenarios/core.py
rochdev: I'm working on a new logic to rely on classes instead of names and it will also address dependency ordering.

cbeauchesne: But we can very easily use only the depends-on graph, and free of cost a circular dependency check. Having two things to consider (depends_on and the order of the list) is hard to maintain.

Here is the algo : 

```
started_containers = []

while len(containers) > 0:
    containers_layer = []

    # get the list of conatiner that have no unstarted dependencies
    for container in containers:
        depends_on = [dep for dep in container.depends_on if dep not in started_containers]
        if len(depends_on) == 0:
            containers_layer.append(container)

    if len(containers_layer) == 0:
        raise Exception("Circular dependency detected")

    # this will start each container in containers_layer in parallel
    start_containers(containers_layer)

    # move containers in containers_layer from containers[] to started_containers[]
    started_containers.extends(containers_layer)
    containers = [container for container in containers if container not in containers_layer]
```

utils/_context/_scenarios/core.py
cbeauchesne: Ok by reading the code, I understood that you also want to have the `post_start` in the threads, am I right ? 

Could you just add a comment explaining why, it'll save some cognitive effort for future maintainers 🙏 .

utils/_context/containers.py
```diff
@@ -63,6 +63,13 @@ def create_inject_volume():
     _get_client().volumes.create(_VOLUME_INJECTOR_NAME)
 
 
+def start_container(container, replay=False):
```
cbeauchesne, line 66: Could you just change the prototype of `TestedContainer.start` by adding `replay` argument, and calling `post_start` at the end?

cbeauchesne (changes requested):

utils/_context/containers.py
cbeauchesne: You can even change `post_start` name to `_post_start` (python convention for private methods)

utils/_context/_scenarios/core.py
rochdev: I can add a comment but I'm not sure I understand the part that needs a comment. Can you clarify what needs to be clarified? 😆

cbeauchesne: The purpose of `replay` mode is to no start anything. By just reading this code, we cannot understand why `_start_containers` is called on replay mode.

```suggestion
        # on replay mode, _start_containers() won't actually start containers, 
        # but will only call the necessary post_start method
        warmups.append(self._start_containers)
```

rochdev: I tend to prefer more descriptive code to comment, maybe I could name the method something better 🤔

utils/_context/containers.py
rochdev: I made the change 👍

rochdev:
> If I'm not wrong, buddies also require the proxy (and possibly the agent?)

Just saw this and added a dependency on the agent just in case.

> Can I close this PR ?

Yes.

utils/_context/containers.py
```diff
@@ -162,10 +164,13 @@ def start(self) -> Container:
         self.wait_for_health()
         self.warmup()
 
+        if not replay:
```
cbeauchesne, line 167: I don't get your code. You always starts the container (in both normal mode and replay mode), but you execute `_post_start` only on replay mode?  🤔

cbeauchesne (changes requested):

cbeauchesne:
There are still lot of work on this PR, I'm putting it in draft mode to prevent spam in our notifs.

utils/_context/containers.py
rochdev: Looks like I did that in reverse, pushing a fix now.

rochdev:
> There are still lot of work on this PR, I'm putting it in draft mode to prevent spam in our notifs.

Can you clarify? Assuming the tests start passing it's actually ready for review.

utils/_context/containers.py
rochdev: Looks like there are many places that call `.start` directly, so to avoid too many code changes I'll leave the `post_start` out of `start` and out of the threads as it was before.

utils/_context/containers.py
```diff
@@ -493,6 +500,19 @@ def __init__(
             healthcheck={"test": f"curl --fail --silent --show-error localhost:{self.port}", "retries": 60},
             ports={"7777/tcp": self.port, "7778/tcp": weblog._grpc_port},
             stdout_interface=interfaces.library_stdout,
+            depends_on=[
```
cbeauchesne, line 500: `depends_on` must be a list of `Testedcontainer` instances, not class. Other wise, if you have several container that are instance of the same class, but with different dependancies, it'll break.

cbeauchesne (changes requested):

utils/_context/containers.py
rochdev: The only way to do that would be to define this externally, so when objects have been instantiated. This pushes the dependency management to the user of the class instead of having it right on the class itself. Since these are instantiated all over the place, it would pollute a lot of the code. I also disagree that it needs to be instances, as if the dependencies are well-defined, it should be possible to add them to the classes. For example, I would expect every BuddyContainer to have the same dependencies, and if that is no longer true, then I would expect to have something like a NodeBuddyContainer. If a class describes a container, then its container should always have the same dependencies.

utils/_context/containers.py
rochdev: Ended up just reverting the change to `post_start` as the main goal of this PR was addressing `start`.

utils/_context/_scenarios/core.py
```diff
@@ -310,13 +311,62 @@ def _get_warmups(self):
...
-            for container in self._required_containers:
-                warmups.append(container.start)
+        # on replay mode, _start_containers() won't actually start containers,
+        # but will only call the necessary post_start method
+        warmups.append(self._start_containers)
+
+        return warmups
+
+    def _start_containers(self):
+        dependencies = self._get_dependencies()
```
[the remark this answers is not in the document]
cbeauchesne: The `dependencies` name is definitly not self-explanatory. We cannot understand what is this data structure without reading the code that produces it. 

`_get_dependencies` misleading me, as by its name, I initially understood it was a list of deps, where it's more a map.

`_start_containers` is not as simple as "start container". It recursively crawls the dependency tree. In particular, if I'm not wrong, the algo is  sensible to the order we add container.

utils/_context/containers.py
cbeauchesne: It's introducing a hidden constraint, it must not be done without a good reason. 

Actually, It's not needed, as the Testcontainer.depends_on is a public property, it does not have to be existing and complete at container creation, we can definitly use that : 

```
    my_container = SomeConatainer()
    self.weblog. depends_on.append(my_container)
```

rochdev: Yes but then you'd have to know about all the containers added to the array all over the code, keep track of that, and make sure references are either passed properly, or keep looping through the array to get the instance. It's just overall messy.

On the other hand, by having it on the class, it's centralized so it's very clean and easy to understand, and it allows doing things like saying weblog depends on all buddies by putting a dependency on BuddyContainer instead of on individual containers.

cbeauchesne: > Yes but then you'd have to know about all the containers added to the array all over the code, keep track of that, and make sure references are either passed properly, or keep looping through the array to get the instance, and keep repeating that logic everywhere that dependencies exist. It's just overall messy.

Which is also true right now, you can do `self.weblog.depends_on.append(SomeContainer)`. In theory, at the end of `DockerScenario.__init__()`, the dependency tree is done, and should not be updated. We can enforce that. But again, what you describe is true for both solutions.

> by having it on the class, it's centralized

It'll be evenly centralized, in `__init__` method. If it's done correctly, you don't even need to use `depends_on.append`, it only needs to create containers in the right order. See the code above as an example.

> it's very clean and easy to understand

I do not agree. The link instance <-> class is very very very weak, and one could easily add another container, or change a container class, and have a hard time to understand why the dependency tree is brokken


Before
```python
class EndToEndScenario(DockerScenario):
    def __init__(self):
        super().__init__() # DockerScenario create the proxy
        self.weblog_container = WeblogContainer()
        self.agent_container = AgentContainer()
```

After, the only thing to change is that if A needs B, then B must be created before A in order to avoid a call to `TestedContainer.depends_on.append`

```python
class EndToEndScenario(DockerScenario):
    def __init__(self):
        super().__init__() # DockerScenario create the proxy
        self.agent_container = AgentContainer(needs=[self.poxy_container])
        self.weblog_container = WeblogContainer(use_proxy=use_proxy, needs=[self.agent_container])
```

cbeauchesne: Another way of telling it : if you want to use the paradigm where its classes that ship info, and not  instances, you must use it everywhere, or nowhere. But the middle road where we use them sometimes as instance, sometimes as class, is not sustainable.

utils/_context/_scenarios/core.py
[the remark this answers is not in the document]
rochdev: I added clarifying comments, let me know if that works!

> In particular, if I'm not wrong, the algo is sensible to the order we add container.

This is no longer the case, it can be in any order.

utils/_context/_scenarios/core.py
```diff
@@ -317,6 +317,8 @@ def _get_warmups(self):
 
         return warmups
 
+    # Build a tree of dependencies between containers and then start the
+    # containers in the right order and in parallel if possible.
     def _start_containers(self):
```
cbeauchesne, line 322: Python stuff here 

```
    def _start_containers(self):
    """ 
      Build a tree of dependencies between containers and then start the
      containers in the right order and in parallel if possible.
    """
```

it's what's called docstrings, it'll be attached to the method, and for IDE that supports python, will. be displayed on hoverging the function calls.

Same review for other comments

utils/_context/_scenarios/core.py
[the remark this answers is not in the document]
cbeauchesne: Let say you have 

```yml
parent_1:
   depends_on: [child_A]
parent_2:
   depends_on: [child_B]
parent_3:
   depends_on: [child_C]
parent_4:
   depends_on: [child_A, child_B, child_C]
```

It'll start sequentially `child_A`, `parent_1`, `child_B`, `parent_2`, `child_C`, `parent_3`, then `parent_4`.

But if you have : 


```yml
parent_4:
   depends_on: [child_A, child_B, child_C]
parent_1:
   depends_on: [child_A]
parent_2:
   depends_on: [child_B]
parent_3:
   depends_on: [child_C]
```

if I'm not wrong, it'll start in parallel `child_A`, `child_B`, `child_C`, then in parallel `parent_4`, `parent_1`, `parent_2` and `parent_3`, right ? 

So far, I'm fine with that, let's not do over-engineering. But if it's true, I think having a snall comment somewhere explaining that for performance reason, containers with lot of common depens should be placed in first will help a lot.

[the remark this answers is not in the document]
rochdev: > if I'm not wrong, it'll start in parallel child_A, child_B, child_C, then in parallel parent_4, parent_1, parent_2 and parent_3, right ?

No, both cases will work as your first example. This is why there are 2 loops, one to build the dependency tree, and another one to start the containers according to the dependencies that have been calculated by looking at every container definition first.

utils/_context/containers.py
rochdev: It's never as instance with my approach, it's always by class. Maybe my mistake was adding it as an instance property. Would it work for you if I move `depends_on` to the class instead? My reasoning is that a class represents a _container_ and not an _image_. So a _container_ would always have the same dependencies.

cbeauchesne: > It's never as instance with my approach, it's always by class.

In the current code, it's always by instances. We won't mix boths approach. I also add that having a list of `depends_on` that always contains stuff that are not started will make the data structure harder to understand.

utils/_context/_scenarios/core.py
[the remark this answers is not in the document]
cbeauchesne: > No, both cases will work as your first example. 

As my **first** example, so the worst case whatever the order in the list ? 🤔 


> This is why there are 2 loops, one to build the dependency tree,

Then could you add more comments about the data structure produced by `_get_dependencies`?

utils/_context/containers.py
rochdev: > In the current code, it's always by instances. We won't mix boths approach.

Yes that's what I'm asking, would you be fine with me moving to _always_ by class and _never_ by instance?

rochdev: Reasoning is that the only way to do it by container without using their names would be to have explicit references to every container, for example with `self._kafka_container` and so on because there are interdependencies across multiple methods. While doable, I think the class approach is simpler.

cbeauchesne: No I'm not : 

* using instances is the natural way of doing. 
* Using classes will requires more advanced python tricks to do it cleanly, for a code harder to understand and maintain.
* it'll requires way more efforts to move everything to classes, compared to use the [code above](https://github.com/DataDog/system-tests/pull/2732#discussion_r1678201838).

I see absolutely no reason for using classes.

rochdev: > it'll requires way more efforts to move everything to classes,

No, it's literally the same as the current code for this PR, but with static properties instead.

> compared to use the https://github.com/DataDog/system-tests/pull/2732#discussion_r1678201838.

That code doesn't work, and making it work will require way more code and trickeries that this PR currently has.

rochdev: I think I have an idea to make this work with an external map. Let me try something.

rochdev: Nevermind that didn't work, the only way I see to make this work with containers is to add properties for each containers on the scenario, and then add dependencies everywhere as the container objects are being instantiated.

rochdev: I moved the whole logic to the caller so that it now works on the instances directly. This also removed the need to generate a dependency map as dependencies are now added directly by their dependents. Let me know if that's more in line with what you had in mind!

cbeauchesne: > I moved the whole logic to the caller so that it now works on the instances directly.

Yes, it's  exactly what I meant, it's going into the right direction, and no need for any trickery.

> This also removed the need to generate a dependency map as dependencies are now added directly by their dependents.

Yes, one complexity removed.

Now, we need a very last thing to have a very clear and self explanatory data structure : first add back a `depends_on` property in `TestedContainer` then, every time you have 

```python
self._container_dependencies[container_a].append(container_b)
```

Your replace it by : 

```python
container_a.depends_on.append(container_b)
```

(On this example, `container_b` needs to be started before `container_a`)

And then, `_container_dependencies` can be removed.

We can also add it in the `TestedContainer` constructor, but using it would require a little bit more of work, so I'm fine with that for now.

Finally,  in _start_dependancies:

```python
for dependency in self._container_dependencies.get(container, []):
```

becomes

```
for dependency in container.depends_on:
```

That's exactly the same logic, but dependencies are shipped directly in the container instance, rather than in a separated dictionnary, which will make the code easier to understand an maintain.

rochdev: At that point I'm thinking of even moving the parallel start logic to `TestedContainer` itself as well, WDYT?

rochdev: It wouldn't work for multiple root containers, but at the end of the day the goal is to start weblog, so weblog can start all of its dependencies itself at that point.

rochdev: Actually it would work even then since at that point `start` would return its thread that could then be joined.

cbeauchesne: > At that point I'm thinking of even moving the parallel start logic to TestedContainer itself as well, WDYT?

but 

> It wouldn't work for multiple root containers

The code that works on all situations is the one to go. If one day we have two root containers, we'll need to revert.

rochdev: The original comment has now been addressed, and the code has been updated to reflect our latest discussions.

utils/_context/containers.py
```diff
@@ -132,7 +136,39 @@ def stop_previous_container(self):
...
+            pytest.exit(f"Circular dependency detected between containers: {dependencies}", 1)
+
+        seen.append(self)
+
+        with self._starting_lock:
+            if self._starting_thread is None:
+                self._starting_thread = Thread(target=self._start_with_dependencies, args=(seen,))
+                self._starting_thread.start()
+
+            return self._starting_thread
```
cbeauchesne, line 155: Even if it literally the same execution flow, it's cleaner to leave the function outside the context manager 

```suggestion
        return self._starting_thread
```

utils/_context/containers.py
```diff
@@ -162,6 +198,9 @@ def start(self) -> Container:
         self.wait_for_health()
         self.warmup()
 
+    def depend_on(self, container):
```
cbeauchesne, line 201: `TestedContainer.depends_on.append()` is fine, and I don't think you use `TestedContainer.depend_on` anywhere. =>  you can remove it.

utils/_context/containers.py
```diff
@@ -132,7 +136,39 @@ def stop_previous_container(self):
...
+
+    def _start_async(self, seen: list):
+        """ Start the container and its dependencies in a thread with circular dependency detection """
+        if self in seen:
+            dependencies = " -> ".join([s.name for s in seen])
+            pytest.exit(f"Circular dependency detected between containers: {dependencies}", 1)
+
+        seen.append(self)
+
+        with self._starting_lock:
```
cbeauchesne, line 150: Small comment won't hurt anybody 

```suggestion
        with self._starting_lock:
            # This lock will prevent any race condition, as this function can be called on the same container in 
            # different threads.
```

utils/_context/containers.py
```diff
@@ -132,7 +136,39 @@ def stop_previous_container(self):
...
+    def start(self) -> Thread:
+        return self._start_async([])
+
+    def _start_async(self, seen: list):
+        """ Start the container and its dependencies in a thread with circular dependency detection """
+        if self in seen:
+            dependencies = " -> ".join([s.name for s in seen])
+            pytest.exit(f"Circular dependency detected between containers: {dependencies}", 1)
+
+        seen.append(self)
```
cbeauchesne, line 148: You must move this block inside the `lock`, as you have a race condition between the `if` and the `append`

cbeauchesne (changes requested):
LGTM. Just need to fix the CI crashes, and some minor changes, and we're good to merge !

cbeauchesne (changes requested):
To fix Ci issue on Lib injection : 

* `WeblogInjectionScenario` must join threads
* `OpenTelemetryCollectorContainer.start` has a wrong return type

cbeauchesne:
@rochdev i've added two commits : 

* A first one that put the circular check in the main thread : having a exception raising inside a thread created an output hard to understand (or would have required to propagate the exception in the main thread). With that, it's the same logic, but easier to understand without extra code.
* And some test cases

I'm waiting for the CI, and if it's all green (or not related, PHP failures expected ... 😭 ) -> merge !

rochdev:
@cbeauchesne Is everything good? I'm not sure what is going on with all the failures but it doesn't seem like it would be related?

cbeauchesne:
No it's should not be. Just for my peace of mind, I'd like to have an almost-all-green CI before merging it. Don't worry, I'm taking care of it.

cbeauchesne (approved):

PR #2733, opened 2024-07-11
Change AWS messaging propagation & DSM tests to use actual AWS

wconti27:
## Motivation

This PR initially began to simply enable a few more Java AWS SNS propagation tests, now that dd-trace-java has the functionality to propagate APM and DSM context through SNS topics. In enabling these tests, it was discovered that Localstack, the 3rd party Docker image used to mock AWS services, does not provide the functionality needed to properly test these new features. 

It was decided to move all tests to AWS, which should boost our confidence in these propagation tests.

### **General Changes**
- Remove Localstack container image
- Change AWS clients to reference our Datadog Sandbox Eng account
- Add AWS auth tokens via environment variables provided through Github secrets
- A distinct hash `UNIQUE_ID` created that will be appended to all AWS resource names (queue names, topic names, etc) to avoid conflicting resources between different system-tests CI runs.

### **Test Specific Changes**
 **Cross Propagation Messaging Integration Tests:**
- Add distinct id as AWS SQS Queue, SNS Topic, and Kinesis stream names. Prevents test resources conflicting. 
- Add functions to cleanup test resources used in AWS after setup functions have run
- Add distinct id to the message being sent, and have consumer services stop consuming once exact message is found

**DSM Test Updates:**
- Add distinct id to AWS queue / topic names.
- Add distinct id to message sent through queues
- Since queue names are now dynamic, we need to calculate the DSM hash for each test case. The DSM hashing function from dd-trace-py has been added to test utils.

## Workflow

## Reviewer checklist


tests/integrations/utils.py
```diff
@@ -144,3 +148,39 @@ def get_span_from_agent(weblog_request):
                 return span_child
 
         raise ValueError(f"Span is not found for {weblog_request.request.url}")
+
+
+def delete_sqs_queue(queue_name):
+    queue_url = f"https://sqs.us-east-1.amazonaws.com/601427279990/{queue_name}"
+    sqs_client = boto3.client("sqs")
```
cbeauchesne, line 155: It rely on a third party service ?

cbeauchesne:
I set it as draft, you can set it back to normal when you need a review

pawelchcki:
Thanks William. Awesome PR!

One question - is it at all possible to spilt the testing into one part working with "localstack" and the second "extended" part to work with AWS ? 

I'm asking mostly from 3 perspectives: 
1) avoiding engineers locally to setup aws credentials to run at least some basic tests
2) avoiding flakiness of actuall AWS
3) and most importantly - we might actually have to put some of those tests in GitLab to avoid exposing secrets to our AWS accounts on public CIs

utils/build/docker/python/flask/app.py
```diff
@@ -531,7 +531,8 @@ def consume_kafka_message():
 @app.route("/sqs/produce")
 def produce_sqs_message():
     queue = flask_request.args.get("queue", "DistributedTracing")
-    message = "Hello from Python SQS"
+    message = flask_request.args.get("message", "Hello from Python SQS")
```
cbeauchesne, line 534: Could you update `docs/weblog` ?

utils/build/docker/python/install_ddtrace.sh
```diff
@@ -7,8 +7,8 @@ cd /binaries
 if [ -e "dd-trace-py" ]; then
     echo "Install from local folder /binaries/dd-trace-py"
     pip install /binaries/dd-trace-py
-elif [ "$(ls *.whl *.tar.gz | wc -l)" = "1" ]; then
-    path=$(readlink -f $(ls *.whl *.tar.gz))
+elif [ "$(ls *.whl | wc -l)" = "1" ]; then
```
cbeauchesne, line 10: why removing `.tar.gz` ?

cbeauchesne (changes requested):
And CI is failing

.github/workflows/run-end-to-end.yml
```diff
@@ -100,249 +121,265 @@ jobs:
...
-      if: steps.build.outcome == 'success' && contains(inputs.scenarios, '"DEFAULT"')
-      run: ./run.sh DEFAULT
       env:
-        DD_API_KEY: ${{ secrets.DD_API_KEY }}
+        AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+        AWS_DEFAULT_REGION: ${{ secrets.AWS_DEFAULT_REGION }}
+        AWS_REGION: ${{ secrets.AWS_DEFAULT_REGION }}
+        UNIQUE_ID: ${{ needs.generate-id.outputs.unique-id }}
+    # - name: Run DEFAULT scenario
```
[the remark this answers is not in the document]
wconti27: this was done to speed up ci runs of the affected tests. I will revert before my final commit

.github/workflows/run-end-to-end.yml
```diff
@@ -52,8 +52,19 @@ env:
   REGISTRY: ghcr.io
 
 jobs:
+  # generates a unique ID used for Cross Tracer Propagation Tests (for naming AWS resource uniquely). We need this ID to be saved and reused for replay scenario.
```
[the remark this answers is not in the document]
wconti27: understood, i will change how this works.

utils/build/docker/python/install_ddtrace.sh
wconti27: fixed this

tests/integrations/utils.py
wconti27: I don't understand the question here. This code cleans up AWS resources that were used for this test, in this case AWS SQS queues

wconti27: So it relies on AWS

utils/build/docker/python/install_ddtrace.sh
wconti27: actually, the tests were failing on docker pull for PHP and .NET. The problem is because we are building the tracer buddies, however, these repos have `.tar` files that end up causing the check on line 10 to pass, which is incorrect, and tries to build using the tar file. I've added a whitelist of these file names but maybe this should be fixed some other way.

.github/workflows/run-end-to-end.yml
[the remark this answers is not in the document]
wconti27: this has been updated and should always be unique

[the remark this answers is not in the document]
wconti27: ... without waiting for a new runner

.github/workflows/run-end-to-end.yml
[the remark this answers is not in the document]
wconti27: uncommented these tests

.github/workflows/run-end-to-end.yml
```diff
@@ -74,13 +77,25 @@ jobs:
       with:
         name: ${{ inputs.binaries_artifact }}
         path: binaries/
+    - name: Generate Unique ID to be used for all AWS Test Resource names
+    # generates a unique ID used for Cross Tracer Propagation Tests (for naming AWS resource uniquely). We need this ID to be saved and reused for replay scenario.
+      id: generate_unique_id
+      run: |
+        HASH=$(uuidgen | tr -d '-' | head -c 10)
+        UNIQUE_ID=${HASH}_${{ inputs.library }}_${{ matrix.weblog }}
```
cbeauchesne, line 85: I'd be more confortable with a more unique name for this en var. `SYSTEM_TESTS_INTEGRATION_UNIQUE_ID` ?

cbeauchesne: Though, why this unique Id can't be computed inside the test process ?

utils/_context/containers.py
```diff
@@ -658,6 +669,13 @@ def get_image_list(self, library: str, weblog: str) -> list[str]:
 
     def configure(self, replay):
         super().configure(replay)
+
```
cbeauchesne, line 672: leftover :)

cbeauchesne (changes requested):
Few changes requested.

BTW, may I ask for later PRs, to cut the task in smaller items? big PR are quite hard to review and merge 🙏

utils/build/docker/python/flask/app.py
wconti27: Done

.github/workflows/run-end-to-end.yml
wconti27: Thanks for fixing this!

utils/build/docker/python/flask/app.py
wconti27: Ive added all missing endpoints to the docs

.github/workflows/run-end-to-end.yml
```diff
@@ -81,6 +81,11 @@ jobs:
     - name: Build buddies weblog images
       if: inputs.build_buddies_images
       run: ./utils/build/build_tracer_buddies.sh
+      env:
+        AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
```
cbeauchesne, line 85: Leftover? Secrets must not be necessary to build  images.

.github/workflows/run-end-to-end.yml
```diff
@@ -108,11 +113,19 @@ jobs:
       run: ./run.sh DEFAULT
       env:
         DD_API_KEY: ${{ secrets.DD_API_KEY }}
+        AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
```
cbeauchesne, line 116: The default scenario does not run anything with AWS, it should not be necessary

.github/workflows/run-end-to-end.yml
```diff
@@ -128,6 +141,10 @@ jobs:
       run: ./run.sh INTEGRATIONS
       env:
         DD_API_KEY: ${{ secrets.DD_API_KEY }}
+        AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
```
cbeauchesne, line 144: is it needed for this scenario? or only for the CROSSED_TRACING_LIBRARIES ?

cbeauchesne (changes requested):
in `utils/_context/containers.py`, in function `collect_logs`, there is a guard that prevent leaking keys, can you add `AWS_ACCESS_KEY_ID` here if present ?

utils/_context/containers.py
```diff
@@ -330,7 +330,7 @@ def collect_logs(self):
         SEP = "=" * 30
 
         keys = [
-            bytearray(os.environ["DD_API_KEY"], "utf-8"),
+            bytearray(os.environ["DD_API_KEY", "AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY"], "utf-8"),
```
cbeauchesne, line 333: It's more this : 

```suggestion
            bytearray(os.environ["DD_API_KEY"], "utf-8"),
            bytearray(os.environ["AWS_ACCESS_KEY_ID"], "utf-8"),
            bytearray(os.environ["AWS_SECRET_ACCESS_KEY"], "utf-8"),
```

Your syntax was pythonic, but it was not achieving what you expect: `some_dict[a, b, c]` is looking for the tuple `(a, b, c)` in `some_dict`. And obviously, `os.environ` contains only strings as keys.

cbeauchesne (changes requested):

utils/_context/containers.py
cbeauchesne: PS : this was not poping in the CI, because exception in that function are ignored.

wconti27: Ah makes sense, i made the change quickly, thanks!

cbeauchesne: ah, my suggestion was crap, because the value is not necessarly here. You'll need to check if they are here before adding

wconti27: ok, i changed it to check if set, then add. this should be all set

cbeauchesne (approved):
Ok, so far LGTM.

**BUT** :) 

Every time we rely on an external service in testing, we are prone to this service healthiness. I think we can give it a try, but at the first sign of flakiness, I think it's better to move those tests in a dedicated scenario, to not change the signal quality of CROSSED_INTEGRATION and INTEGRATION. 

Does it make sense for you ?

wconti27:
> Ok, so far LGTM.
> 
> **BUT** :)
> 
> Every time we rely on an external service in testing, we are prone to this service healthiness. I think we can give it a try, but at the first sign of flakiness, I think it's better to move those tests in a dedicated scenario, to not change the signal quality of CROSSED_INTEGRATION and INTEGRATION.
> 
> Does it make sense for you ?

Sounds good to me!

tabgok (approved):

amarziali (approved):
looks ok from a cursory read. Perhaps I would have put topic names and other arn in a file and referenced them. It will be easier in case we need to change them in the future

nayeem-kamal (approved):

tlhunter (approved):

PR #2777, opened 2024-07-19
[nodejs] allow mounting local dd-trace-js as weblog volume

rochdev:
## Motivation

Right now we have to `npm pack` the local code, copy the resulting archive to `system-tests/binaries`, and rebuild weblog for every single code change before we can rerun the tests.

By allowing to instead mount the local dd-trace-js folder as a volume on the container, we can simply rerun the tests on code changes without having to go through any of the above.

## Changes

Allow mounting local dd-trace-js as weblog volume.

## Workflow

## Reviewer checklist


utils/build/docker/nodejs/express4-typescript.Dockerfile
```diff
@@ -23,8 +23,9 @@ ENV PGPORT=5433
 ENV DD_DATA_STREAMS_ENABLED=true
 
 # docker startup
-RUN printf '#!/bin/bash\nnode dist/app.js' > app.sh
-RUN chmod +x app.sh
+COPY utils/build/docker/nodejs/app.sh app.sh
+RUN printf '#!/bin/bash\nnode dist/app.js' > server.sh
+RUN chmod +x server.sh
 CMD ./app.sh
```
robertomonteromiguel, line 29: Sometimes I confuse between app.js, app.sh, server.js and server.sh

robertomonteromiguel (changes requested):
I like the idea of assembling the volumes, but I also wonder how much we gain. I mean looking for the balance of keeping the code easy to read, its complexity and homogeneity (we run apps in different languages).

rochdev:
> I like the idea of assembling the volumes, but I also wonder how much we gain. I mean looking for the balance of keeping the code easy to read, its complexity and homogeneity (we run apps in different languages).

This is by far the biggest pain point of working with system tests, so I'd say it's warranted especially since it doesn't really increase complexity all that much. As for homogeneity, I would agree but only in the sense that this should also be done for other languages if possible, although this is out of scope of this first PR.

utils/build/docker/nodejs/app.sh
```diff
@@ -1,7 +1,10 @@
 #!/bin/bash
 
-if [ ${UDS_WEBLOG:-} = "1" ]; then
-    ./set-uds-transport.sh
```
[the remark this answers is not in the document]
rochdev: It's not removed, it's moved to a different file. This is because `app.sh` is a hardcoded entry point, but its code changes depending on the image as some of them generate the file. So there is now a new `app.sh` that is always the same and checks for the volume, and calls out to a separate file that is now generated instead of `app.sh`.

utils/_context/containers.py
```diff
@@ -481,12 +481,22 @@ def __init__(
 
         self.port = weblog.port
 
+        volumes = {
+            f"./{host_log_folder}/docker/weblog/logs/": {"bind": "/var/log/system-tests", "mode": "rw",},
+        }
+
+        if os.path.exists("./binaries/nodejs-load-from-local"):
```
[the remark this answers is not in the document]
rochdev: This is already the case for `nodejs-load-from-npm`, I simply followed the pattern. Discussing this pattern is out of the scope of this PR. It's worth noting that having a separate file per language has benefits, for example when I work locally I can use my own dd-trace-js but the latest release for dd-trace-java.

docs/scenarios/parametric.md
```diff
@@ -144,10 +144,13 @@ TEST_LIBRARY=python PYTHON_DDTRACE_PACKAGE=git+https://github.com/Datadog/dd-tra
 
 #### NodeJS
 
-There is two ways for running the NodeJS tests with a custom tracer:
+There is three ways for running the NodeJS tests with a custom tracer:
 1. Create a file `nodejs-load-from-npm` in `binaries/`, the content will be installed by `npm install`. Content example:
     * `DataDog/dd-trace-js#master`
 2. Clone the dd-trace-js repo inside `binaries`
+3. Create a file `nodejs-load-from-local` in `binaries/`, this will disable installing with `npm install dd-trace` and
```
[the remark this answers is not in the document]
rochdev: This is literally the problem that this PR is trying to solve.

utils/build/docker/nodejs/parametric/package.json
```diff
@@ -10,6 +10,8 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
+    "@opentelemetry/api": "^1.9.0",
+    "@opentelemetry/core": "^1.25.1",
```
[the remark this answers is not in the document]
rochdev: Because the app uses it, but it previously used the ones installed transitively by `dd-trace`, which is an anti-pattern. An app should always depend on all the modules it uses directly. This is especially true now that `dd-trace` can be linked, meaning its dependencies are no longer flattened at the root.

utils/build/docker/nodejs/express4-typescript.Dockerfile
rochdev: Not sure what action item is asked for here.

utils/build/docker/nodejs/install_ddtrace.sh
```diff
@@ -4,25 +4,34 @@ set -eu
...
+
+    elif [ -e /binaries/dd-trace-js ]; then
+        target=$(npm pack /binaries/dd-trace-js)
+        echo "install from local folder /binaries/dd-trace-js"
+        npm install $target
+
+    else
+        target="dd-trace"
+        echo "install from NPM"
+        npm install $target
```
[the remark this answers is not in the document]
rochdev: I did the change but will wait for all review feedback to be addressed before pushing since CI takes 2h.

cbeauchesne:
> I addressed his comments and he just went on vacation for a month

Don't worry, he just sent me all his pending reviews.

docs/scenarios/parametric.md
```diff
@@ -144,10 +144,13 @@ TEST_LIBRARY=python PYTHON_DDTRACE_PACKAGE=git+https://github.com/Datadog/dd-tra
...
+   will instead mount the `../dd-trace-js` directory in the container and then `npm link` to it. This also removes the
```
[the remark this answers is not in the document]
cbeauchesne: A universal way is to get the content of `nodejs-load-from-local`, and use it to have the location of the repo. 

```suggestion
   will instead get the content of the folder, and use it as a location of the `dd-trace-js` repo and then `npm link` to it. For instance, if this repo is at the location, you can set the content of this file to `../dd-trace-js`. This also removes the
```

utils/build/docker/nodejs/app.sh
```diff
@@ -1,7 +1,10 @@
...
+if [ -e /volumes/dd-trace-js ]; then
```
cbeauchesne, line 3: If any error in `npm link`, this will make the error easier to see in the output 

```suggestion
set -e

if [ -e /volumes/dd-trace-js ]; then
```

utils/build/docker/nodejs/app.sh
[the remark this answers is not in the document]
cbeauchesne: Not very important, but is it possible to merge `app.sh` and `server.sh` ?

utils/build/docker/nodejs/install_ddtrace.sh
```diff
@@ -4,25 +4,34 @@ set -eu
...
-    target=$(</binaries/nodejs-load-from-npm)
-    echo "install from: $target"
-
-elif [ -e /binaries/dd-trace-js ]; then
-    target=$(npm pack /binaries/dd-trace-js)
-    echo "install from local folder /binaries/dd-trace-js"
+if [ -e /binaries/nodejs-load-from-local ]; then
+    echo "using local version that will be mounted at runtime"
 
+    echo "0.0.0" > SYSTEM_TESTS_LIBRARY_VERSION
```
cbeauchesne, line 10: As it'll be always the `dev` version, using a high version number is preferable : otherwise, any test activated "starting" a version will be deactivated locally.

And better if it's possible : get the real version from the local folder.

cbeauchesne (changes requested):

utils/build/docker/nodejs/app.sh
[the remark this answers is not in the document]
rochdev: In order to merge them, the Dockerfiles would have to append the dynamic code to the existing static file instead of creating a new one. Definitely doable, but I'm not sure it would be much cleaner. I can definitely do that if that's your preference.

docs/scenarios/parametric.md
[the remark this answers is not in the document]
simon-id: obviously have `../dd-trace-js` be the default value as i think most devs use it too

utils/build/docker/nodejs/install_ddtrace.sh
cbeauchesne: > do they need to be moved to app.sh maybe ?

Unfortunately not possible, it must be done at build step.

utils/build/docker/nodejs/app.sh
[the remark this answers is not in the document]
cbeauchesne: Definitly no strong opinion here, I let you choose what you prefer.

utils/build/docker/nodejs/install_ddtrace.sh
rochdev: Good point on getting the real version, which should also be possible for the dependencies I think. However, this does mean that the weblog will need to be rebuilt whenever there is a version change. That is probably an acceptable limitation though, and I prefer that trade-off to the trade-off of using a high version number which could fail in an older branch.

docs/scenarios/parametric.md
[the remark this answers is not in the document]
rochdev: @simon-id I just used what's in the file without a default since the file needs to be created anyway. If you really think it would make it easier to use to have a default I can add it as well.

utils/build/docker/nodejs/app.sh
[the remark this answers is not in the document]
rochdev: I ended up moving everything to `app.sh` and appending only the call out to `node`. It's actually cleaner and doesn't require a new file 👍

utils/build/docker/nodejs/install_ddtrace.sh
rochdev: #2801 has now landed and I've adapted this PR to use the new health check, along with adding the AppSec event rules versions to the endpoint. So the version is now collected entirely at runtime, no weblog rebuild needed even when changing dd-trace-js branch locally.

utils/_context/containers.py
```diff
@@ -555,12 +555,23 @@ def __init__(
...
+            path = open("./binaries/nodejs-load-from-local").read().strip(" \r\n")
```
cbeauchesne, line 563: use a context manager to ensure that the file is closed in any situation (and `strip`, by default, strips any spaces/newline chars) : 

```suggestion
            with open("./binaries/nodejs-load-from-local", encoding="utf-8") as f:
                path = f.read().strip()
```

docs/scenarios/parametric.md
```diff
@@ -144,10 +144,14 @@ TEST_LIBRARY=python PYTHON_DDTRACE_PACKAGE=git+https://github.com/Datadog/dd-tra
 
 #### NodeJS
 
-There is two ways for running the NodeJS tests with a custom tracer:
+There is three ways for running the NodeJS tests with a custom tracer:
```
cbeauchesne, line 147: Could you also update `docs/execute/binaries.md` ?

utils/_context/_scenarios/parametric.py
```diff
@@ -353,16 +353,25 @@ def node_library_factory() -> APMLibraryTestServer:
     nodejs_appdir = os.path.join("utils", "build", "docker", "nodejs", "parametric")
     nodejs_absolute_appdir = os.path.join(_get_base_directory(), nodejs_appdir)
     nodejs_reldir = nodejs_appdir.replace("\\", "/")
+    volumes = {}
+
+    if os.path.exists("./binaries/nodejs-load-from-local"):
+        path = open("./binaries/nodejs-load-from-local").read().strip(" \r\n")
```
cbeauchesne, line 359: Use a context manager (see comment below)

utils/_context/containers.py
```diff
@@ -555,12 +555,23 @@ def __init__(
...
+            volumes[os.path.abspath(path)] = {
+                "bind": f"/volumes/dd-trace-js",
```
cbeauchesne, line 565:
```suggestion
                "bind": "/volumes/dd-trace-js",
```

utils/_context/containers.py
```diff
@@ -684,10 +695,11 @@ def post_start(self):
         # https://github.com/DataDog/system-tests/issues/2799
         if self.library == "nodejs":
             with open(self.healthcheck_log_file, mode="r", encoding="utf-8") as f:
-                data = json.load(f)
+                lib = json.load(f)["library"]
```
cbeauchesne, line 698: Could you let it as `data`. in a near future we'll need other info that will be at the root of `data`.

If repeated `data["library"]` hurts, you can add `library_data=data["library"]` after.

utils/build/docker/nodejs/express4-typescript/app.ts
```diff
@@ -36,7 +36,8 @@ app.get('/healthcheck', (req: Request, res: Response) => {
     library: {
       language: 'nodejs',
       version: require('dd-trace/package.json').version,
-      libddwaf_version: require('@datadog/native-appsec/package.json').libddwaf_version
+      libddwaf_version: require('dd-trace/node_modules/@datadog/native-appsec/package.json').libddwaf_version,
+      appsec_event_rules_version: require('dd-trace/packages/dd-trace/src/appsec/recommended.json').metadata.rules_version
```
cbeauchesne, line 40: It's unfortunately a little bit more complex. This code must get the version of the rule file actually used by appsec extension, not the version of the file shipped by default : 

This rules file can be a custom one, specified by `DD_APPSEC_RULES` env var. In that situation, the version returned here must be the version of this file (see ` utils/build/docker/overwrite_waf_rules.Dockerfile`).

@simon-id do you know if there is an easy solution to do so ?

cbeauchesne (changes requested):

docs/scenarios/parametric.md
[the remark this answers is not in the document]
simon-id: actually nvm you're right

utils/build/docker/nodejs/express4-typescript/app.ts
simon-id: Oh is this a new requirement ? because previously it was like this:
```npm explore dd-trace -- cat packages/dd-trace/src/appsec/recommended.json | jq -r '.metadata.rules_version // "1.2.5"' > SYSTEM_TESTS_APPSEC_EVENT_RULES_VERSION```
which was only getting the default version, not a custom one.

```suggestion
      appsec_event_rules_version: require('dd-trace/packages/dd-trace/src/appsec/waf').wafManager?.rulesVersion
```

cbeauchesne: That's the first layer. Then another logic that overwrite the value is executed here : https://github.com/DataDog/system-tests/blob/main/utils/build/build.sh#L234-L246

simon-id: well this new suggestion i'm proposing would get the version used by the waf at any time. but i don't think it would take RC changes into account

cbeauchesne: That's perfect

utils/_context/containers.py
rochdev: I was mostly trying to save horizontal space as I don't know how to break a line of code in multiple lines in Python 😅

utils/_context/containers.py
rochdev: Are you sure about newlines? I see other occurrences of this type of stripping in the codebase, and the documentation is also unclear on that.

cbeauchesne: I'm pretty sure, and I even tested before commenting :) 

It removes `whitespaces`, see this explanation : https://stackoverflow.com/questions/77258198/full-list-symbols-stripped-with-str-strip-by-default

The [doc](https://docs.python.org/3.4/library/stdtypes.html#str.strip) is a little bit misleading, as it does not precise that "whitespace" is this list of chars.

image

But nevermind, that'a a detail.

cbeauchesne: Though, I'm curious about the test you perform, is there a situation where it does not work ? 🤔

utils/_context/containers.py
cbeauchesne: > I was mostly trying to save horizontal space as I don't know how to break a line of code in multiple lines in Python

Don't hurt yourself, let `./format.sh` take care of this 😄

utils/build/docker/nodejs/express4-typescript/app.ts
rochdev: @simon-id `wafManager` is `null` when dd-trace is not initialized.

rochdev: Actually never mind it should be initialized at that point, so not sure why it's still `null` 🤔

rochdev: Looks like some tests are testing missing rules, so of course there is no version reported in that case.

@simon-id Any ideas to work around that?

rochdev: I think using `SYSTEM_TESTS_APPSEC_EVENT_RULES_VERSION` when present and otherwise falling back to the previous logic of checking `.metadata.rulesVersion` would be equivalent to how it worked before.

rochdev: Ended up keeping `wafManager.rulesVersion` since that would always be more accurate when present, and then falling back to `SYSTEM_TESTS_APPSEC_EVENT_RULES_VERSION` when it's not set.

Qard (approved):

utils/build/docker/nodejs/express4-typescript/app.ts
rochdev: The above didn't work, so I tried to fallback to `require('dd-trace/packages/dd-trace/src/appsec/recommended.json').metadata.rules_version` only when the environment variable isn't set as it was before but at runtime, and it still doesn't work.

@simon-id Any idea what is the correct ordering of things here?

utils/build/docker/nodejs/express4/.npmrc
simon-id, line 1: why this whole nested thing ?

utils/build/docker/nodejs/express4/app.js
```diff
@@ -44,7 +44,8 @@ app.get('/healthcheck', (req, res) => {
     library: {
       language: 'nodejs',
       version: require('dd-trace/package.json').version,
-      libddwaf_version: require('@datadog/native-appsec/package.json').libddwaf_version
+      libddwaf_version: require('dd-trace/node_modules/@datadog/native-appsec/package.json').libddwaf_version,
```
simon-id, line 47: why change from the direct require ?

utils/build/docker/nodejs/express4-typescript/app.ts
simon-id: wdym it doesn't work ?

simon-id: no idea about build.sh, but the js suggestion i made should definitely work no ?

docs/execute/binaries.md
```diff
@@ -36,6 +36,11 @@ But, obviously, testing validated versions of components is not really interesti
     * `DataDog/dd-trace-js#master`
 2. Clone the dd-trace-js repo inside `binaries`
 
+OR
+
+1. Create a file `nodejs-load-from-load` in `binaries/`, the content will be installed by `npm link` to the local folder
```
simon-id, line 41:
```suggestion
3. Create a file `nodejs-load-from-load` in `binaries/`, the content will be installed by `npm link` to the local folder
```

simon-id: also `nodejs-load-from-what???`

rochdev: This step is not a step that comes after 2, it's a step that replaces 2.

> also nodejs-load-from-what???

Ah yes it should be `from-local`, will fix.

utils/build/docker/nodejs/express4/app.js
rochdev: That didn't work, but reverting doesn't work either. I'm looking into it.

utils/build/docker/nodejs/express4/.npmrc
rochdev: Because when you `npm link` the dependencies are always nested, npm cannot flatten them. But when you `npm install` npm flattens dependencies by default. This means that `@datadog/native-appsec` can end up either at the root or inside of `dd-trace`. By telling npm to always nest, the folder structure stays the same with both `install` or `link`.

utils/build/docker/nodejs/express4/app.js
rochdev: I misunderstood the question actually, the reason is because when dd-trace is linked the `@datadog/native-appsec` is not flattened at the root, so we need to require it from under `dd-trace`. This is also why the `.npmrc` files were added with a `nested` configuration.

docs/execute/binaries.md
rochdev: Fixed the incorrect name, but I disagree with the change as it's not an additional step, it replaces the previous steps entirely.

rochdev:
It looks like the CI failure is also happening on other branches and so it's unrelated to this PR.

utils/build/docker/nodejs/nextjs/next.config.js
```diff
@@ -1,6 +1,6 @@
 /** @type {import('next').NextConfig} */
 const nextConfig = {
-  output: 'standalone'
```
simon-id, line 3: can you explain this change ? and should we remove the file if it's empty now ?

docs/execute/binaries.md
simon-id: But Roch I don't think they're steps ? I think they're already alternatives no ?

utils/build/docker/nodejs/express4/app.js
```diff
@@ -44,7 +44,8 @@ app.get('/healthcheck', (req, res) => {
...
+      appsec_event_rules_version: require('dd-trace/packages/dd-trace/src/appsec/waf').wafManager?.rulesVersion
```
simon-id, line 48: btw i just realized, both infos are on the same object already, you can require it once and just use both props
```suggestion
      libddwaf_version: require('dd-trace/packages/dd-trace/src/appsec/waf').wafManager?.ddwafVersion,
      appsec_event_rules_version: require('dd-trace/packages/dd-trace/src/appsec/waf').wafManager?.rulesVersion
```

utils/build/docker/nodejs/nextjs/next.config.js
rochdev: Standalone mode doesn't work with anything external, so linking dd-trace after the Next build is not possible.

rochdev: Not sure if the file can be removed, I kept it to keep the project structure as close as possible to what the CLI generates (the same is true for a few other files that are probably useless for our purpose).

simon-id: maybe it's worth having 
```javascript
  // disabled because not compatible with linking
  // output: 'standalone'
```

utils/build/docker/nodejs/express4/app.js
rochdev: With this change the `.npmrc` files are not longer needed too which is a nice plus.

simon-id: Niiice

utils/build/docker/nodejs/parametric/.npmrc
simon-id, line 1: forgot to remove this ?

docs/execute/binaries.md
simon-id: just like in parametric.md btw, same principle

simon-id: we should just copy those two files tbh:
```markdown
3. Create a file `nodejs-load-from-local` in `binaries/`, this will disable installing with `npm install dd-trace` and
   will instead get the content of the file, and use it as a location of the `dd-trace-js` repo and then mount it as a
   volume and `npm link` to it. For instance, if this repo is at the location, you can set the content of this file to
   `../dd-trace-js`. This also removes the need to rebuild the weblog image since the code is mounted at runtime.
```

rochdev: Oh right, I misread that entirely. Will change it.

utils/build/docker/nodejs/parametric/.npmrc
rochdev: Can no longer remove as we're accessing `@datadog/native-appsec` again.

simon-id (approved):
Don't know much about the python internals but otherwise LGTM

rochdev:
@cbeauchesne PR is ready for one last look!

cbeauchesne (approved):
All good, thank you for tackling this.

PR #2789, opened 2024-07-22
parametric: adds a consistent interface for retrieving traces and spans

mabdinur:
## Motivation

Provides a consistent API for getting spans, traces, and chunk root spans (first span in a trace payload) in system tests. This will provide a standard for writing parametric tests that do not rely on how spans and trace level tags are serialized and packed into trace payloads.

## Changes

- Removes Span and Trace similarity helpers that were previously used to search for spans. Also removes instances were traces captured by  `_TestAgentAPI.wait_for_num_traces()`  were indexed directly (ex: traces[0][0], traces[1][0]).  The order of traces/spans in a payload are not guaranteed and chunk root spans (spans with trace level tags) are NOT always the root span of the trace. To provide a consistent API for retrieving spans this PR introduces 3 methods and ensures they are used by all parametric tests:
  - `find_trace(traces, trace_id)`: retrieves a trace by id. So the order traces were received in should not matter.
  - `find_span(trace, span_id)`: retrieves a span  from a trace by span id. With this helper the order spans are serialized will not impact tests. 
  - `find_span_in_traces(traces, trace_id, span_id)`: retrieves a span from a list of traces. This is a shortcut for calling `find_trace` and then `find_span` to get a span from a list of traces.
  - `find_chunk_root_span(trace)`: retrieves the first span in the captured trace payload (NOT ALWAYS the root span).  Raises an error if the first span in the trace does not have trace level tags (ex: this can occur if the spans in the trace were sorted by start time).

- Documents new helper functions: https://github.com/DataDog/system-tests/pull/2789/commits/12fc7e41bb7ae9fcfa3768571d48fbc55356394f 

- Adds `sort_by_start` argument to `_TestAgentAPI.wait_for_num_traces()`. 
   - Some languages serialize spans in the order they are created (ex: python), other tracers serialize spans in the order they are finished (ex: dotnet), and other do a mix of both (partial flushing). The agent is not guaranteed to receive spans/traces in any order. By adding the `sort_by_start` argument to `_TestAgentAPI` we provide a mechanism to capture equivalent trace payloads across tracers. 
   - Most tests sort traces and spans by start time (sort_by_start=True by default). However in other tests the order spans in a trace payload matter. For example when capturing trace level tags from the first span in a trace payload. In these tests  `sort_by_start` is set to `False`.
 
- Updates go and php parametric apps to ensure integer span_ids  and trace_ids are returned by start_span endpoints. This will allow us to use `find_span` and `find_trace` helpers in all tests.
   - Updates the `otel_start_span` endpoint in the `go parametric app` to return valid uint64 trace_ids. Previously this endpoint returned invalid trace_ids. Since this value was never used in tests this wasn't a problem. Now it is a problem so I fix it!
   - Updates the `otel_start_span` endpoint in the `php parametric app` to return integer trace_ids and span_ids (instead of hexadeciamal). This will align the php endpoint with other parametric apps.
   - There are still inconsistencies in how parametric start_span endpoints return the  trace_id and span_ids of newly created spans (some apps return 128bit trace ids as big nums, some truncate trace_ids and return just the lower order uint64 component, some return span_ids/trace_ids as base 10 strings). These inconsistencies will be resolved in a future PR. `find_span` and `find_trace` were written in a way to normalize all these edge cases.

## Workflow

## Reviewer checklist


utils/parametric/spec/trace.py
```diff
@@ -154,107 +154,37 @@ def decode_v06_stats(data: bytes) -> V06StatsPayload:
...
-
-def find_trace_by_root(traces: List[Trace], span: Span) -> Trace:
+def find_trace(traces: List[Trace], trace_id: int) -> Optional[Trace]:
     """Return the trace from `traces` with root span matching all fields of `span`."""
-    trace = find_similar_trace_by_root(traces, span)
-    _assert_span_match(span, root_span(trace))
-    return trace
+    trace_id = trace_id & (2^64 - 1) # Use 64-bit trace id
+    for trace in traces:
+        # This check ignroes the high bits of the trace id
```
lucaspimentel, line 161: small typo
```suggestion
        # This check ignores the high bits of the trace id
```

utils/parametric/spec/trace.py
```diff
@@ -154,107 +154,37 @@ def decode_v06_stats(data: bytes) -> V06StatsPayload:
...
-        similarity = _span_similarity(root, span)
-        if similarity > max_similarity:
-            max_score_trace = trace
-            max_similarity = similarity
-    return max_score_trace
-
-
-def find_trace_by_root(traces: List[Trace], span: Span) -> Trace:
+def find_trace(traces: List[Trace], trace_id: int) -> Optional[Trace]:
     """Return the trace from `traces` with root span matching all fields of `span`."""
```
lucaspimentel, line 158: I think this comment doesn't match the function anymore.
```suggestion
    """Return the trace from `traces` with trace id matching `trace_id`."""
```

utils/parametric/spec/trace.py
```diff
@@ -154,107 +154,37 @@ def decode_v06_stats(data: bytes) -> V06StatsPayload:
...
+        # This check ignroes the high bits of the trace id
+        # TODO: Check _dd.p.tid 
+        if trace and trace[0].get("trace_id") == trace_id:
+            return trace
 
 
-def find_similar_span(trace: Trace, span: Span) -> Span:
-    """Return a span from the trace which most closely matches `span`."""
+def find_span(trace: Trace, span_id: int) -> Optional[Span]:
+    """Return a span from the trace matches all fields in `span`."""
```
lucaspimentel, line 168:
```suggestion
    """Return a span from the trace that matches `span_id`."""
```

utils/parametric/spec/trace.py
```diff
@@ -154,107 +154,37 @@ def decode_v06_stats(data: bytes) -> V06StatsPayload:
...
-        similarity = _span_similarity(span, other_span)
-        if similarity > max_similarity:
-            max_similarity = similarity
-            max_similarity_span = other_span
-    return max_similarity_span
-
-
-def find_span(trace: Trace, span: Span) -> Span:
+def find_span_by_name(trace: Trace, span_name: str) -> Optional[Span]:
     """Return a span from the trace matches all fields in `span`."""
```
lucaspimentel, line 175:
```suggestion
    """Return a span from the trace that matches `span_name`."""
```

lucaspimentel: What if more than one span in the traces matches the given span name?

utils/parametric/spec/trace.py
```diff
@@ -154,107 +154,37 @@ def decode_v06_stats(data: bytes) -> V06StatsPayload:
...
-                if var == "service":
-                    meta = similar.get("meta")
-                    if meta is not None:
-                        s_val = meta.get(var)
-                assert (
-                    val == s_val
-                ), f"Span field '{var}' mismatch '{val}' != '{s_val}'\nSpan   : {span}\nSimilar: {similar}"
-    return similar
+def find_span_by_resource(trace: Trace, resource: str) -> Optional[Span]:
+    """Return a span from the trace matches all fields in `span`."""
```
lucaspimentel, line 183:
```suggestion
    """Return a span from the trace that matches `resource`."""
```

mabdinur: Honestly I am not a fan of retrieving spans by name. We will always have access to the SpanID, get spans by id is a lot more reliable than retrieving spans by names. I added this helper to avoid refactoring existing tests. On second thought I will just remove this method. We should always use `get_span()`.

lucaspimentel: Agreed!

tests/parametric/test_128_bit_traceids.py
```diff
@@ -438,12 +438,13 @@ def test_w3c_128_bit_propagation_tid_in_parent(self, test_agent, test_library):
...
                     pass
 
         traces = test_agent.wait_for_num_traces(1, clear=True)
-        parent = find_span_in_traces(traces, Span(name="parent", service="service"))
-        child = find_span_in_traces(traces, Span(name="child", service="service"))
+        trace = find_trace(traces, parent.trace_id)
+        parent = find_span(trace, parent.span_id)
+        child = find_span(trace, child.span_id)
 
         parent_tid = parent["meta"].get("_dd.p.tid")
```
lucaspimentel, line 449: This looks like one of the cases we were talking about. The spec says the tag `_dd.p.tid` (and any other `_dd.p.*` propagated tag) should be placed in the first span of every trace chunk sent to the agent (meaning the first span in the payload, not necessarily the first span created or the first span closed). Whether that first span is the parent or the child span is undefined/unpredictable and that relationship should not be used for this test.

mabdinur: Yup. The initial goal of this PR is to standardize around retrieving spans by ids using find_trace and find_span. Once this change is passing for all enabled tests, then we can move away from sorting spans by start_time (by default) and/or introduce helpers to capture the chunk root spans / trace level tags.

darccio (approved):

tests/parametric/test_128_bit_traceids.py
mabdinur: I updated this test. There are a few more tests with the same issue

tests/parametric/test_headers_tracecontext.py
```diff
@@ -742,13 +743,25 @@ def test_tracestate_w3c_p_extract(self, test_agent, test_library):
...
+            ) as s2:
+                pass
+
+            with test_library.start_span(
+                name="p_not_propagated_valid_dd_tracestate",
+                http_headers=[
+                    ["traceparent", "00-12345678901234567890123456789015-1234567890123459-00"],
+                    ["tracestate", "key1=value1,dd=s:2;t.dm:-4"],
+                ],
+            ) as s3:
```
mabdinur, line 755:
```suggestion
            ) as s2:
```

cbeauchesne (commented):
Parametric tests are currently in a very bad shape, that's unfortunately not he best moment to change their internals, because it's hard to ensure that we're not adding some flakiness source to an already complicated situation 😭 

I'm re-running failing job, hoping to see some light.

cbeauchesne (approved):

tests/parametric/conftest.py
```diff
@@ -222,12 +222,14 @@ def snapshot_context(self, token, ignores=None):
...
-    def wait_for_num_traces(self, num: int, clear: bool = False, wait_loops: int = 30) -> List[Trace]:
+    def wait_for_num_traces(
+        self, num: int, clear: bool = False, wait_loops: int = 30, sort_by_start: bool = True
+    ) -> List[Trace]:
         """Wait for `num` traces to be received from the test agent.
 
         Returns after the number of traces has been received or raises otherwise after 2 seconds of polling.
 
-        Returned traces are sorted by the first span start time to simplify assertions for more than one trace by knowing that returned traces are in the same order as they have been created.
+        When sort_by_start=True returned traces are sorted by the span start time to simplify assertions by knowing that returned traces are in the same order as they have been created.
```
lucaspimentel, line 232:
```suggestion
        When sort_by_start=True the spans in each trace chunk are sorted by the span start time to simplify assertions by knowing that returned traces are in the same order as they have been created.
```

tests/parametric/test_128_bit_traceids.py
```diff
@@ -427,30 +429,31 @@ def test_w3c_128_bit_propagation_tid_consistent(self, test_agent, test_library):
...
                     pass
 
-        traces = test_agent.wait_for_num_traces(1, clear=True)
-        parent = find_span_in_traces(traces, Span(name="parent", service="service"))
-        child = find_span_in_traces(traces, Span(name="child", service="service"))
+        traces = test_agent.wait_for_num_traces(1, clear=True, sort_by_start=False)
+        trace = find_trace(traces, parent.trace_id)
+        assert len(trace) == 2
+        chunk_root = find_chunk_root_span(trace)
+        non_chunk_root = trace[1]
```
lucaspimentel, line 449: It looks like we're expected the child span to be the second span in the payload, but that is not guaranteed. We should find the span with `parent_id > 0`.

tests/parametric/test_128_bit_traceids.py
```diff
@@ -427,30 +429,31 @@ def test_w3c_128_bit_propagation_tid_consistent(self, test_agent, test_library):
...
 
-        parent_tid = parent["meta"].get("_dd.p.tid")
-        child_tid = (child.get("meta") or {}).get("_dd.p.tid")
-        propagation_error = parent["meta"].get("_dd.propagation_error")
+        tid_chunk_root = chunk_root["meta"].get("_dd.p.tid")
+        tid_non_chunk_root = (non_chunk_root.get("meta") or {}).get("_dd.p.tid")
```
lucaspimentel, line 452: Just curious (not a Python expert), why access `meta` in different ways here?
```
chunk_root["meta"]
non_chunk_root.get("meta")
```

tests/parametric/test_128_bit_traceids.py
lucaspimentel, line 449: There's an inconsistency here with trying to find the root span with `find_chunk_root_span()`, but using the span position to find the child span. The ordering of parent and child spans is not guaranteed. 

The `_dd.p.tid` span tag should always be added to the first span in the payload. This first span is not guaranteed to be the root span, so we should rely on span position only to test this tag.

I think this test do something like (consider this pseudo-code)
```python
# assert that "_dd.p.tid" is added only to first span in payload
assert trace[0]["meta"].get("_dd.p.tid") is not None
assert trace[1]["meta"].get("_dd.p.tid") is None
```

----

The `_dd.propagation_error` tag is another story. It is an optional tag that may be added to the root span while decoding headers. I don't think we should test it in this test. We already have a test above (`test_datadog_128_bit_propagation_tid_malformed_optional_tag`) that checks for its presence when a header is malformed. We could add another test to test for its absence when a header is valid, and this test would assert that the tag is missing on every span in the trace.

utils/parametric/spec/trace.py
```diff
@@ -154,107 +152,49 @@ def decode_v06_stats(data: bytes) -> V06StatsPayload:
...
-                # Right now, nodejs sets a tag named "service" in meta with the correct value.
-                if var == "service":
-                    meta = similar.get("meta")
-                    if meta is not None:
-                        s_val = meta.get(var)
-                assert (
-                    val == s_val
-                ), f"Span field '{var}' mismatch '{val}' != '{s_val}'\nSpan   : {span}\nSimilar: {similar}"
-    return similar
+def find_chunk_root_span(trace: Trace) -> Span:
```
lucaspimentel, line 191: This `find_chunk_root_span` function is currently mixing up a few concepts (root spans and trace-level tags) that do not have a consistent correlation that we can rely on.

### Root span

The root span of a trace is the span with no (local or remote) parent span. The correct way to find the root span of a trace is to find the span where `parent_id is null or 0`. A local trace must have exactly one root span. If a local trace is split into multiple chunks, only one of the chunks will contain the root span. In other words, a trace chunk can only ever have zero or one root spans.

### Trace-level tags

For several reasons, we are unfortunately very inconsistent with how we treat "trace-level" tags. We cannot rely on these tags to find the root span.
- `_dd.origin` tag is added to all spans in a trace.
- `language` tag is added to multiple spans in a trace. Originally, we added it to all spans except spans that represent an external service call. Today we can safely add it to all spans. It was never added only to the root span.
- `runtime-id` is added to all top-level spans in a trace. The root span will always have this tag, since it is always a top-level span, but other spans in the same trace can also have this tag.
- `_dd.p.tid` and other propagated tags (`_dd.p.*`) are added to the first span in every trace chunk. It is not guaranteed that this first span will be the root span, hence propagated tags have no relationship to the root span.

lucaspimentel: There are other examples:
- `process_id` is added only to root spans, so technically is could be used to find the root span. but really, there's no reason to rely on span tags when we can use the `parent_id` which is the definition of roto span and always works. (for example, the behavior of `process_id` could change in the future) 
- "priority sampling rate" tags (`_dd.*_psr`) are added to the root span as well
- `_sampling_priority_v1` has changed over time and is pretty complicated. Let's just say that at a minimum, it needs to show up in at least one span (any span) in every trace chunk. But for the Trace Agent, placing it in some spans is more efficient than others.
- `env` is added to all spans
- `version` is added to all spans whose service name is the same as the default service name

tests/parametric/test_128_bit_traceids.py
mabdinur: chunk_root["meta"] will raise a KeyError Exception if meta is not found in map 

non_chunk_root.get("meta") will return None if meta is not found in the map.

We are pretty inconsistent in system tests. We should stick to one convention. In my opinion we use .get for safe access

manifests/dotnet.yml
```diff
@@ -305,10 +305,7 @@ tests/:
       Test_Trace_Sampling_Tags_Feb2024_Revision: v2.48.0
       Test_Trace_Sampling_With_W3C: v2.46.0
     test_tracer.py:
-      # The dotnet parametric app generates trace chunks in a way that differs from the other languages
-      # This causes the sci tests to fail. Before enabling these tests we must ensure that the first span in the
-      # trace chunk is always the local root span.
-      Test_TracerSCITagging: missing_feature
+      Test_TracerSCITagging: v2.48.0
```
lucaspimentel, line 308: Thank you!

tests/parametric/test_partial_flushing.py
```diff
@@ -79,14 +78,17 @@ def do_partial_flush_test(self, test_agent, test_library):
...
                 pass
-            partial_trace = test_agent.wait_for_num_traces(1, clear=True, wait_loops=30)
-            child_span = find_span_in_traces(partial_trace, Span(name="child1"))
+            partial_traces = test_agent.wait_for_num_traces(1, clear=True, wait_loops=30)
+            partial_trace = find_trace(partial_traces, parent_span.trace_id)
             assert len(partial_trace) == 1
+            child_span = find_span(partial_trace, child1.span_id)
             assert child_span["name"] == "child1"
+
     traces = test_agent.wait_for_num_traces(1, clear=True)
```
lucaspimentel, line 89: I don't understand this test 😅 . Why is this test calling `test_agent.wait_for_num_traces(1, ...)` twice? It is because we expect two trace chunks? Why not use `test_agent.wait_for_num_traces(2, ...)` instead?

And what is `partial_trace` vs `full_trace`? Does `test_agent` return the entire trace (the two trace chunks combined) in the second call to `test_agent.wait_for_num_traces()`? That seems weird. How can we test that each span was sent in it own trace chunk?

lucaspimentel: Ideally, a test for partial flushing with `DD_TRACE_PARTIAL_FLUSH_MIN_SPANS=1` and two spans should:
- assert that the tracer sent exactly two trace chunks
- assert that each trace chunk has exactly one span
- assert that the span from one chunk is the parent and the span from the other chunk is the child

The order of the chunks or the spans is not defined, so the test shouldn't which trace chunk contains the parent span or the child span.

tests/parametric/test_trace_sampling.py
```diff
@@ -49,9 +48,9 @@ class Test_Trace_Sampling_Basic:
     def test_trace_sampled_by_trace_sampling_rule_exact_match(self, test_agent, test_library):
         """Test that a trace is sampled by the exact matching trace sampling rule"""
         with test_library:
-            with test_library.start_span(name="web.request", service="webserver"):
+            with test_library.start_span(name="web.request", service="webserver") as span:
                 pass
-        span = find_span_in_traces(test_agent.wait_for_num_traces(1), Span(name="web.request", service="webserver"))
+        span = find_span_in_traces(test_agent.wait_for_num_traces(1), span.trace_id, span.span_id)
```
lucaspimentel, line 53: Could we use `find_only_span()` here (and other tests below) since we expect only one span?

tests/parametric/test_128_bit_traceids.py
mabdinur: Ahhh makes sense. I thought the goal of this test was to validate the first span in the trace payload, not necessarily the chunk root span. I just updated all instances were traces were accessed directly. All tests that validate the first span in a trace now use `find_chunk_root`.

tests/parametric/test_trace_sampling.py
mabdinur: I added find_only_span after making this change and I was too lazy to revert it. I will fix this

utils/parametric/spec/trace.py
mabdinur: With this change, I will go over tests that use `find_trace_with_trace_level_tags` and make sure this helper is used correctly.

tests/parametric/test_128_bit_traceids.py
lucaspimentel: I think we're getting a few concepts mixed up here. Sorry if I wasn't clear before 😓 
 
> I thought the goal of this test was to validate the first span in the trace payload, not necessarily the chunk root span

Yes, that was correct: `_dd.p.tid` should be added to the first span in each trace chunk, which may or may not be the trace's root span. To check for this tag, we should rely on the order in which the tracer sent the spans in each chunk, not the relationship between spans (root, parent ,child).

>... not necessarily the chunk root span.
> All tests that validate the first span in a trace now use `find_chunk_root`.

What do you mean by "chunk root"? That's not a thing. The spans found in a trace chunk have no implied relationship (root, parent, child, etc) except that they all belong to the same trace.

utils/parametric/spec/trace.py
lucaspimentel: `find_trace_with_trace_level_tags` still doesn't make sense to me. First, shouldn't it be `find_span...`? You're trying to find and return a single span, not a trace.

But more importantly, different "trace-level" tags have different rules about which spans we add them to. I listed 8 examples in my comment above. So you cannot find a single spans that contains all the trace-level tags. Tests for each individual tag should know where to look for the span and assert that they were added to the correct span.

Maybe we have different definitions of what a "trave-level" tags is, since that's more of an implicit concept that we haven't defined anywhere in APM.

utils/parametric/spec/trace.py
```diff
@@ -154,107 +152,57 @@ def decode_v06_stats(data: bytes) -> V06StatsPayload:
...
+    """Return the only span in a list of traces. Raises an error if there are no traces or more than one span."""
+    assert len(traces) == 1, traces
+    assert len(traces[0]) == 1, traces[0]
+    return traces[0][0]
 
-def find_span(trace: Trace, span: Span) -> Span:
-    """Return a span from the trace matches all fields in `span`."""
-    return _assert_span_match(span, find_similar_span(trace, span))
 
+def find_span_with_trace_level_tags(trace: Trace) -> Span:
```
mabdinur, line 191: rename to propagated tags

utils/parametric/spec/trace.py
lucaspimentel: After talking over this on zoom I understand better now, thanks! `find_trace_with_trace_level_tags` was looking for the span with `_dd.p.*` tags (aka propagated tags).

We decided it would be better to grab the first span in the trace chunk, since that's where those tags are supposed to be. Searching for any span with the tag can result in test passing that should fail (if the tag is on the wrong span).

tests/parametric/test_128_bit_traceids.py
lucaspimentel: Talked on zoom and cleared up my confusion. Thanks! (resolving this thread)

tests/parametric/test_partial_flushing.py
mabdinur: Yup, the test agent combines all trace chunks into one list and returns it: https://github.com/DataDog/dd-apm-test-agent/blob/master/ddapm_test_agent/agent.py#L425. The easiest way to test partial flushing is to call `wait_for_num_traces` twice 😭 (or we can check whether the spans are marked as top level but that's less reliable).

lucaspimentel: Oh, I see 😬  Thanks for clarifying!

lucaspimentel: We may need to change that someday to allow testing "chunk tags" like `_dd.p.tid`, but not here, not now 😅

PR #2792, opened 2024-07-23
Add PHP to live debugger system tests

bwoebi:
Prepare system tests for live debugger in dd-trace-php version 1.4.

I've verified locally that these system tests pass with the artifact from https://output.circle-artifacts.com/output/job/59686ec3-ad57-47e9-8202-09fb9a298447/artifacts/0/dd-library-php-1.4.0+457f890a449fc458b3b2955d3952e56a129c3eae-aarch64-linux-gnu.tar.gz.

This PR fixes a couple mistakes in existing tests, adds a bit more debug logging to the PHP tracer outputs and makes the probes conform to the current debugging schema (PHP looks at the type rather than the name of the probe).

Note that, by design of the RC client in PHP, no probes are received if nothing is running. Hence there are calls to `await_probes.php`, which actively waits for the probes to be installed, and will also emit diagnostics, which the system-tests can interpret as signal of all probes having been successfully installed.

Further, it's not trivial as in PHP to define routes with wildcards (either you say "all args after first slash are function args" or "route is always static". I've opted to replace a couple slashes in paths by underscores. Not just as nice, but works perfectly fine and is easier.

## Reviewer checklist


robertomonteromiguel:
I move this PR to draft until we can execute the CI build without failures (until the PHP PR is merged )

cbeauchesne:
Seen on zoom, Bob still want to work on this PR.

bwoebi:
Superseded by #6760.

PR #2805, opened 2024-07-23
lower library interface timeout for remote config tests

rochdev:
## Motivation

Waiting 100 seconds seems like a lot, especially since it should be possible to get the remote config from client libraries in milliseconds.

## Changes

Lower library interface timeout for remote config tests.

## Workflow

## Reviewer checklist

### DIFF
--- a/utils/_context/_scenarios/__init__.py
+++ b/utils/_context/_scenarios/__init__.py
@@ -584,7 +584,7 @@ class scenarios:
         rc_api_enabled=True,
         appsec_enabled=False,
         weblog_env={"DD_REMOTE_CONFIGURATION_ENABLED": "true"},
-        library_interface_timeout=100,
+        library_interface_timeout=15,
         doc="",
         scenario_groups=[ScenarioGroup.APPSEC],
     )
@@ -598,7 +598,7 @@ class scenarios:
             "DD_REMOTE_CONFIG_ENABLED": "true",
             "DD_INTERNAL_RCM_POLL_INTERVAL": "1000",
         },
-        library_interface_timeout=100,
+        library_interface_timeout=15,
         doc="",
     )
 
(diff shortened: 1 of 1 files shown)


rochdev:
I'm not sure what this is actually waiting for. It seems that waiting for 20s doesn't work with a poll interval of 2s, but it works with a polling interval of 1s, and the previous values were timeout of 100s with a polling interval of 5s. Does the timeout need to be 20x the polling interval? Is this checking for 20 remote config request? It's very unclear to me why it needs to wait for so long given that we can configure the polling interval all the way down to 1ms if needed.

cc @cbeauchesne

rochdev:
Failures are on tests that are not affected by this change.

cbeauchesne (commented):
Could you compute the time gain with the change of `DD_REMOTE_CONFIG_POLL_INTERVAL_SECONDS` ? 

Because it's far away from what we have in real condition, and the purpose of end-to-end is to be as close as possible from real conditions. We can deviate a little bit from this ideal line, but it must be worthy.

I've updated your branch to fix failures

cbeauchesne (commented):
So far, the CI seems happy.

But I'd really like to know what is the gain of `DD_REMOTE_CONFIG_POLL_INTERVAL_SECONDS=0.25` before approving.

The change on `library_interface_timeout` is totally legit IMO.

rochdev:
> Could you compute the time gain with the change of DD_REMOTE_CONFIG_POLL_INTERVAL_SECONDS ?

Time gain is pretty easy to calculate as it's literally the timeout, so it goes from 100 seconds to 5 seconds.

> the purpose of end-to-end is to be as close as possible from real conditions

It's still real condition, just faster. The fact that we're waiting 5 seconds every time in real environments is only so that we don't overload the backend, but in an isolated test there is no reason not to do this faster.

cbeauchesne:
> Time gain is pretty easy to calculate as it's literally the timeout, so it goes from 100 seconds to 5 seconds.

I was speaking about `DD_REMOTE_CONFIG_POLL_INTERVAL_SECONDS`, not `library_interface_timeout`

rochdev:
> I was speaking about DD_REMOTE_CONFIG_POLL_INTERVAL_SECONDS, not library_interface_timeout

Oh, well without lowering the poll interval, the remote config won't have been fetched before the timeout. As to why it needs to be ~20x lower than the timeout, I have no idea and the commit history was lost for that file so I don't know who could answer it. At the end of the day, the ratio is the same as before and everything is passing 🤷‍♂️

simon-id (approved):
LGTM but not I'm not part of system-tests core team

PR #2883, opened 2024-08-14
[Debugger] Enable python tests

shurivich:
## Motivation
Enable python tests


utils/build/docker/python/flask/debugger_controller.py
```diff
@@ -0,0 +1,128 @@
...
+# dummy line
+# dummy line
+# dummy line
+debugger_blueprint = Blueprint("debugger", __name__, url_prefix="/debugger")
+
+intLocal = 0
+intMixLocal = 0
+
+
+@debugger_blueprint.route("/log", methods=["GET"])
```
cbeauchesne, line 18: could you document in `docs/weblog` what those route are supposed to perform ?

cbeauchesne (changes requested):
Once the doc is done, AGTM

utils/build/docker/python/flask/debugger_controller.py
shurivich: It's actually not the new endpoint. It was not implemented for `python` but already exist for `dotnet` and`java`.
All the documentation already there.
https://github.com/DataDog/system-tests/tree/main/docs/weblog#get-debugger

cbeauchesne (approved):

PR #2912, opened 2024-08-23
Print weblog crash logs

simon-id:
## Changes

When weblog container crashes in the middle of the test, print stdout and sterr logs, instead of showing test failures.

- do not raise an exception when explicit flushing fails, just print a warning and mark as unhealthy
- when stopping a container, if it's already stopped, mark is as unhealthy
- when collecting logs, if unhealthy, print the last 50 lines for stdout and sterr
- swap order of stdout and stderr to have stderr last for more clarity in error messages 

## Reviewer checklist


utils/_context/containers.py
```diff
@@ -326,12 +330,19 @@ def collect_logs(self):
...
-            logger.stdout(f"\n{sep} {self.name} STDERR {sep}")
-            logger.stdout(stderr.decode("utf-8"))
-            logger.stdout(f"\n{sep} {self.name} STDOUT {sep}")
-            logger.stdout(stdout.decode("utf-8"))
-            logger.stdout("")
+            logger.stdout(f"\n{sep} {self.name} STDOUT last {str(tail)} lines {sep}")
+            logger.stdout(stdout)
+            logger.stdout(f"\n{sep} {self.name} STDERR last {str(tail)} lines {sep}")
+            logger.stdout(stderr)
+            logger.stdout(f"\n{sep}============================{sep}")
```
simon-id, line 345: FYI I swapped the order of STDERR and STDOUT to put STDERR last as it's usually more useful than the huge amount of debug logs from STDOUT

cbeauchesne (approved):

PR #2916, opened 2024-08-26
[Debugger] Fix java exception replay inconsistency

shurivich:
## Motivation
Fix java exception replay inconsistency by adding `wait for snapshot` functionality.

### DIFF
--- a/tests/debugger/test_debugger_exception_replay.py
+++ b/tests/debugger/test_debugger_exception_replay.py
@@ -38,14 +38,14 @@ class Test_Debugger_Exception_Replay(base._Base_Debugger_Test):
             if contents:
                 for content in contents:
                     snapshot = content.get("debugger", {}).get("snapshot") or content.get("debugger.snapshot")
-                    if snapshot and snapshot["probe"]["location"]["method"].lower().replace('_', '') == self.method:
+                    if snapshot and snapshot["probe"]["location"]["method"].lower().replace("_", "") == self.method:
                         self.snapshot = snapshot
 
                         logger.debug("Snapshot received")
                         return True
-                        
+
         return False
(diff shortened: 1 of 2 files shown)


tests/debugger/test_debugger_exception_replay.py
```diff
@@ -3,26 +3,55 @@
...
+            if contents:
+                for content in contents:
+                    snapshot = content.get("debugger", {}).get("snapshot") or content.get("debugger.snapshot")
+                    if snapshot and snapshot["probe"]["location"]["method"].lower().replace("_", "") == self.method:
+                        self.snapshot = snapshot
+
+                        logger.debug("Snapshot received")
+                        return True
+
+        return False
```
P403n1x87, line 47:
```suggestion
        def snapshot_received(data):
            if data["path"] == base._LOGS_PATH:

                contents = data["request"].get("content", []) or []
                if contents:
                    for content in contents:
                        snapshot = content.get("debugger", {}).get("snapshot") or content.get("debugger.snapshot")
                        if snapshot and snapshot["probe"]["location"]["method"].lower().replace("_", "") == self.method:
                            self.snapshot = snapshot

                            logger.debug("Snapshot received")
                            return True

            return False


        logger.debug(f"Waiting for snapshot, retry #{retries}")

        self.weblog_responses.add(weblog.get(request_path))
        interfaces.agent.wait_for(snapshot_received, timeout=60)
```

shurivich: It will work great for python, but not for dotnet and java, since we have to send at least two requests to make exception replay work.

robertomonteromiguel (approved):

PR #2958, opened 2024-09-02
[Debugger] Test exception that thrown in recursion method

shurivich:
## Motivation
Test exception that thrown in recursion method


jpbempel (approved):

utils/build/docker/dotnet/weblog/Controllers/DebuggerController.cs
```diff
@@ -119,5 +119,40 @@ public async Task<IActionResult> Nulls(int? intValue = null, string strValue = n
...
+        {
+            if (depth > 0)
+            {
+                return ExceptionReplayRecursion20(depth - 1);
+            }
+            else
+            {
+                throw new System.Exception("Recursion exception");
+            }
+        }
```
GreenMatan, line 156: Can't you simply pass the `depth` via url query param? - it will allow us to test exciting things, such as, querying varying deepness (10, 9, 8, ...) and check that each of them 'finalizes' upon second execution, and do so in various ordering. E.g requesting `exceptionreplay/recursion/10`, `exceptionreplay/recursion/9`, ... and then requesting the same endpoints in various orders to see that we submit snapshots for each of them individually and at correct times (upon second request, and not capturing any snapshots when hitting those endpoints with varying recursion deepness, etc)?

GreenMatan (commented):
Left a small comment

utils/build/docker/dotnet/weblog/Controllers/DebuggerController.cs
shurivich: I can, but it will be hard to find related snapshots due to system test limitations. Right now, the only way I can identify related snapshots is by the unique method name. The system tests are very robust but don't allow much flexibility.

cbeauchesne: Would it be possible to inject identifiers in the query ?

shurivich: It possible, but again makes things even more complicated to identify. For now I created a simplest way to test this functionality. If we want to make things more complicated and cover more cases, we can do it in future PRs.

GreenMatan: We've been discussed in office and concluded that we will postpone it for a later phase, as @shurivich states that it would be very complicated to implement it. We will get System Tests session in the coming weeks from @shurivich that will explain in detail how the infrastructure is built. I envision that after this session we will recommend a few tweaks for the System Tests infra to support the use-case I presented above, alongside similar complex scenarios.
Resolving for now.

GreenMatan (approved):

cbeauchesne (changes requested):
Could you rebase? it'll solve the .NET failures

shurivich:
> Could you rebase? it'll solve the .NET failures
@cbeauchesne done

cbeauchesne (approved):

utils/build/docker/java/spring-boot/src/main/java/com/datadoghq/system_tests/springboot/debugger/DebuggerController.java
```diff
@@ -121,4 +121,27 @@ public String nulls(
                 ". intValue is null " + (intValue == null) +
                 ". strValue is null " + (strValue == null) + ".";
     }
+
+    @GetMapping("/exceptionreplay/simple")
```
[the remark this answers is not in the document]
shurivich: Done 2fe9b8e5423b6a9f178a78804bc7145c334005d6

PR #2979, opened 2024-09-04
Test telemetry report for first batch of "Consistent Configs" for tracers. 

mtoffl01:
## Motivation
[APMAPI-283](https://datadoghq.atlassian.net/browse/APMAPI-283)
Ensure all tracing libraries report telemetry for the configuration options in scope for the above RFC.

## Changes
Adds a new test class Test_Consistent_Configs with logic for asserting telemetry clients report on all "consistent" configs.

## Workflow

## Reviewer checklist

[APMAPI-283]: https://datadoghq.atlassian.net/browse/APMAPI-283?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


manifests/dotnet.yml
```diff
@@ -179,6 +179,7 @@ tests/:
         Test_Scanners: v1.28.6
       test_telemetry.py:
         Test_TelemetryMetrics: missing_feature
+        Test_Consistent_Configs: missing_feature
```
zacharycmontoya, line 182: It looks like this is in the wrong subdirectory. This needs to be moved to the `parametric/` subsection

manifests/ruby.yml
```diff
@@ -345,6 +345,7 @@ tests/:
       Test_Environment: missing_feature
       Test_TelemetryInstallSignature: missing_feature
       Test_TelemetrySCAEnvVar: v2.1.0
+      Test_Consistent_Configs: missing_feature
```
zacharycmontoya, line 348: CI is complaining about this not being sorted alphabetically, so you'll need to move this to be the first entry of the `test_telemetry.py` grouping here

tests/parametric/test_telemetry.py
```diff
@@ -89,6 +89,93 @@ def test_library_settings(self, library_env, test_agent, test_library):
...
+            pass
+        event = test_agent.wait_for_telemetry_event("app-started", wait_loops=400)
+        configuration = event["payload"]["configuration"]
+
+        configuration_by_name = {item["name"]: item for item in configuration}
+        for apm_telemetry_name, value in [
+            ("trace_rate_limit", "100"),
+            ("trace_header_tags", "header:tag"),
+            ("trace_enabled", ("true", True)),
+            ("trace_obfuscation_query_string_regexp", "^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$"),
```
zacharycmontoya, line 135: Should this match the configuration above?

```suggestion
            ("trace_obfuscation_query_string_regexp", "^[a-zA-Z]$"),
```

tests/parametric/test_telemetry.py
```diff
@@ -89,6 +89,93 @@ def test_library_settings(self, library_env, test_agent, test_library):
...
+                "DD_TRACE_HEADER_TAGS": "header:tag",
+                "DD_TRACE_ENABLED": "true",
+                "DD_TRACE_OBFUSCATION_QUERY_STRING_REGEXP": "^[a-zA-Z]$",
+                "DD_TRACE_LOG_DIRECTORY": "/some/temporary/directory",
+                "DD_VERSION": "123",
+                "DD_HTTP_CLIENT_ERROR_STATUSES": "400",
+                "DD_HTTP_SERVER_ERROR_STATUSES": "500",
+                "DD_TRACE_HTTP_CLIENT_TAG_QUERY_STRING": "true",
+                "DD_TRACE_CLIENT_IP_HEADER": "X-Forwarded-For",
+                "DD_TRACE_SERVICE_MAPPING": "plugin:custom",
```
zacharycmontoya, line 119: Can we remove DD_TRACE_SERVICE_MAPPING for now since that's now out-of-scope for this push?

tests/parametric/test_telemetry.py
```diff
@@ -89,6 +89,93 @@ def test_library_settings(self, library_env, test_agent, test_library):
...
+            ("trace_enabled", ("true", True)),
+            ("trace_obfuscation_query_string_regexp", "^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$"),
+            ("trace_log_directory", "/some/temporary/directory"),
+            ("version", "123"),
+            ("trace_http_client_error_statuses", "400"),
+            ("trace_http_server_error_statuses", "500"),
+            ("trace_http_client_tag_query_string", ("true", True)),
+            (
+                "trace_client_ip_header",
+                "x-forwarded-for",
```
zacharycmontoya, line 143: Although HTTP header names are case-insensitive, the tracers are likely going to store and report this value in their configuration with whatever the case-sensitive input was. As a result, let's assert the same string as we set:

```suggestion
                "X-Forwarded-For",
```

tests/parametric/test_telemetry.py
```diff
@@ -89,6 +89,93 @@ def test_library_settings(self, library_env, test_agent, test_library):
...
+            or context.library == "python"
+            or context.library == "ruby"
+        ):
+            assert cfg_item is not None, "Missing telemetry config item for '{}'".format(apm_telemetry_name)
+            assert cfg_item.get("value") is "grpc"
+        if context.library == "php":
+            assert cfg_item is not None, "Missing telemetry config item for '{}'".format(apm_telemetry_name)
+            assert cfg_item.get("value") is "phpredis"
+        # The trace_agent_url is a container address -- don't know the value, but we can assert its not empty (i.e, that it reports)
+        apm_telemetry_name = _mapped_telemetry_name(context, "trace_agent_url")
```
zacharycmontoya, line 174: This one is weird because there are multiple configurations that may combine to set the agent url (agent_host + agent_port OR agent_url). The telemetry behavior when nothing is set may be different for each language, so perhaps that is something we need to standardize on...

However, for the immediate term, may I propose that ONLY assert the configuration key `trace_agent_url` when we are actually explicitly setting environment variable `DD_TRACE_AGENT_URL`? WDYT?

mtoffl01: @link04 said there were issues with .NET when setting trace_agent_url in this test (that is, if it's not the actual agent address, the data is never received). Hence why I didn't set it/tried to assert over default behavior.

But in this test, I'm just asserting the telemetry exists, not asserting over its value. So this shouldn't be a concern?
>  there are multiple configurations that may combine to set the agent url (agent_host + agent_port OR agent_url)

Btw, setting DD_TRACE_AGENT_URL it to a random address still worked in Go, though (i.e, telemetry data was received).

zacharycmontoya: > And in this test, I'm just asserting the telemetry exists, not asserting over its value. So this shouldn't be a concern?

My worry is that not all languages may emit the trace_agent_url telemetry by default (maybe they report the host name and port?), but maybe we could leave this as-is and see

zacharycmontoya (approved):
LGTM

cbeauchesne (changes requested):
CI is failing on the new test

tests/parametric/test_telemetry.py
```diff
@@ -89,6 +89,91 @@ def test_library_settings(self, library_env, test_agent, test_library):
             assert cfg_item.get("origin") == "default", "Unexpected origin for '{}'".format(apm_telemetry_name)
 
 
+@scenarios.parametric
+@rfc("https://docs.google.com/document/d/1kI-gTAKghfcwI7YzKhqRv2ExUstcHqADIWA4-TZ387o")
+@features.telemetry_app_started_event
```
cbeauchesne, line 94: is it really this feature? 🤔

tests/parametric/test_telemetry.py
mtoffl01: Marking this as resolved since we decided all tracers should report telemetry for all "consistent config" items, regardless of whether they've been configured.

tests/parametric/test_telemetry.py
mtoffl01: @cbeauchesne I blindly copied that over from the Test_Environment class below 😄
Technically this behavior is expected when app has started... and all these env vars have been set. But this is also part of the tracing_configuration_consistency effort that we've been working on, and that @zacharycmontoya added a `feature` for in https://github.com/DataDog/system-tests/pull/2980. Should I mark this with `@features.tracing_configuration_consistency` instead, or does that *not* make sense because this is telemetry-specific? WDYT?

cbeauchesne: The RFC filled for this feature is https://docs.google.com/document/d/1kI-gTAKghfcwI7YzKhqRv2ExUstcHqADIWA4-TZ387o/edit#heading=h.ixldy9ty2xy5

If the behavior your testing is described in this RFC, it's definitly the one to use. otherwise, you can either create a new feature, or still use `tracing_configuration_consistency`, up to you/your team to define how you want to define  this feature.

cbeauchesne (changes requested):
The Ci is failing on the new test

smola (commented):
Does this pass on _any_ language? Ideally we would merge together with at least one language enabled, to make sure the test runs correctly.

cbeauchesne (approved):
Ok for me, even with no passing lib. But it'll require some follow up to ensure that the test is ok, otherwise it'll be either not used, or worse, leads to a bad implem on all languages

PR #2976, opened 2024-09-04
Config Consistency DD_TRACE_ENABLED system test.

mhlidd:
## Motivation

Writing Config Consistency system tests based upon the following [RFC](https://docs.google.com/document/d/1kI-gTAKghfcwI7YzKhqRv2ExUstcHqADIWA4-TZ387o/edit#heading=h.o9qoojeu3llz).

## Changes

Created file to hold all config consistency [system tests](https://docs.google.com/document/d/1kI-gTAKghfcwI7YzKhqRv2ExUstcHqADIWA4-TZ387o/edit#heading=h.m3s3qx29yhmw). 
Wrote system test to check for `DD_TRACE_ENABLED` when enabled and disabled. 
Ticket Link: [APMAPI-288](https://datadoghq.atlassian.net/browse/APMAPI-288)

## Workflow

1. ⚠️ Create your PR as draft ⚠️
2. Work on you PR until the CI passes (if something not related to your task is failing, you can ignore it)
3. Mark it as ready for review
    * Test logic is modified? -> Get a review from RFC owner. We're working on refining the `codeowners` file quickly.
    * Framework is modified, or non obvious usage of it -> get a review from [R&P team](https://dd.enterprise.slack.com/archives/C025TJ4RZ8X)

:rocket: Once your PR is reviewed, you can merge it!

🛟 [#apm-shared-testing](https://dd.enterprise.slack.com/archives/C025TJ4RZ8X) 🛟

## Reviewer checklist

* [ ] If PR title starts with `[<language>]`, double-check that only `<language>` is impacted by the change
* [ ] No system-tests internal is modified. Otherwise, I have the approval from [R&P team](https://dd.enterprise.slack.com/archives/C025TJ4RZ8X)
* [ ] CI is green, or failing jobs are not related to this change (and you are 100% sure about this statement)
* [ ] A docker base image is modified?
    * [ ] the relevant `build-XXX-image` label is present
* [ ] A scenario is added (or removed)?
    * [ ] Get a review from [R&P team](https://dd.enterprise.slack.com/archives/C025TJ4RZ8X)

[APMAPI-288]: https://datadoghq.atlassian.net/browse/APMAPI-288?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3010, opened 2024-09-10
Add config consistency tests for DD_VERSION

mhlidd:
## Motivation

Adding tests for tracing configs to make sure implementation across all languages are consistent, based upon the following [RFC](https://docs.google.com/document/d/1kI-gTAKghfcwI7YzKhqRv2ExUstcHqADIWA4-TZ387o/edit#heading=h.o9qoojeu3llz).

## Changes

- Add test cases for when `DD_SERVICE` and `DD_VERSION` are specified, as well as when neither are included.
- Included in existing `tracing_configuration_consistency` feature.
- All tests are currently marked as `missing_feature` in the manifest files
- Small change made to pass in `None` instead of `""` in testing handler call to the python tracer library
- Ticket Link: [APMAPI-347](https://datadoghq.atlassian.net/browse/APMAPI-347)

This PR is stacked on top of #2976 

## Workflow

## Reviewer checklist

[APMAPI-347]: https://datadoghq.atlassian.net/browse/APMAPI-347?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/parametric/test_config_consistency.py
+++ b/tests/parametric/test_config_consistency.py
@@ -95,7 +95,7 @@ class Test_Config_UnifiedServiceTagging:
 
         span1 = find_span_in_traces(traces, s1.trace_id, s1.span_id)
         assert span1["service"] == "version_test"
-        assert span1['meta']["version"] == "5.2.0"
+        assert span1["meta"]["version"] == "5.2.0"
 
         span2 = find_span_in_traces(traces, s2.trace_id, s2.span_id)
         assert span2["service"] != "version_test"
--- a/tests/parametric/test_config_consistency.py
+++ b/tests/parametric/test_config_consistency.py
@@ -3,6 +3,7 @@ Test configuration consistency for features across supported APM SDKs.
 """
 import pytest
 from utils import scenarios, features
+from utils.parametric.spec.trace import find_span_in_traces
 
 parametrize = pytest.mark.parametrize
 
@@ -68,23 +69,34 @@ def set_service_version_tags():
 @scenarios.parametric
 @features.tracing_configuration_consistency
 class Test_Config_UnifiedServiceTagging:
-    @set_service_version_tags()
-    def test_version_tag(self, library_env, test_agent, test_library):
-        assert library_env.get("DD_SERVICE", "test_service") == "test_service"
-        assert library_env.get("DD_VERSION", "5.2.0") == "5.2.0"
-
+    @parametrize("library_env", [{}])
+    def test_default_version(self, library_env, test_agent, test_library):
         with test_library:
-            with test_library.start_span(name="s1"):
+            with test_library.start_span(name="s1") as s1:
                 pass
-            with test_library.start_span(name="s2", service="no dd_service"):
+
+        traces = test_agent.wait_for_num_traces(1)
+        assert len(traces) == 1
+
+        span = find_span_in_traces(traces, s1.trace_id, s1.span_id)
+        assert span["service"] != "version_test"
+        assert "version" not in span["meta"]
+
+    @parametrize("library_env", [{"DD_SERVICE": "version_test", "DD_VERSION": "5.2.0"}])
+    def test_specific_version(self, library_env, test_agent, test_library):
+        with test_library:
+            with test_library.start_span(name="s1") as s1:
+                pass
+            with test_library.start_span(name="s2", service="no dd_service") as s2:
                 pass
 
         traces = test_agent.wait_for_num_traces(2)
         assert len(traces) == 2
 
-        for trace in traces:
-            for span in trace:
-                if span["service"] == "test_service":
-                    assert span["meta"]["version"] == "5.2.0"
-                else:
-                    assert "version" not in span['meta']
+        span1 = find_span_in_traces(traces, s1.trace_id, s1.span_id)
+        assert span1["service"] == "version_test"
+        assert span1['meta']["version"] == "5.2.0"
+
+        span2 = find_span_in_traces(traces, s2.trace_id, s2.span_id)
+        assert span2["service"] != "version_test"
+        assert "version" not in span2["meta"]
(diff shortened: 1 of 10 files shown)


cbeauchesne (approved):

tests/parametric/test_config_consistency.py
```diff
@@ -43,3 +43,34 @@ def test_tracing_disabled(self, library_env, test_agent, test_library):
...
+
+def set_service_version_tags():
+    env1 = {}
+    env2 = {"DD_SERVICE": "test_service", "DD_VERSION": "5.2.0"}
+    return parametrize("library_env", [env1, env2])
+
+
+@scenarios.parametric
+@features.tracing_configuration_consistency
+class Test_Config_UnifiedServiceTagging:
```
mtoffl01, line 56: Cool! Is the approach to contain all service, env and version test functions inside this class?

mhlidd: Yes, as each individual environment setup should not affect the others.

tests/parametric/test_config_consistency.py
```diff
@@ -43,3 +43,34 @@ def test_tracing_disabled(self, library_env, test_agent, test_library):
...
+    @set_service_version_tags()
+    def test_version_tag(self, library_env, test_agent, test_library):
+        assert library_env.get("DD_SERVICE", "test_service") == "test_service"
+        assert library_env.get("DD_VERSION", "5.2.0") == "5.2.0"
```
zacharycmontoya, line 60: This line looks like it will fail for env1 because it has an empty set of environment variables. Have you tested this locally?

mhlidd: It doesn't fail because it the `.get(key, val)` defaults the return value to the `val` if it is not found in `library_env`.

mtoffl01: So you're saying lines 59 & 60 will pass even with `env1`? That feels **wrong** 😆

mhlidd: Hmm, this is done elsewhere as well, such as `tests/parametric/test_dynamic_configuration.py::TestDynamicConfigTracingEnabled::test_tracing_client_tracing_enabled`. 

I could turn it into an if statement to check both conditions and assert them. Do you think that is a more sound way to implement this check?

mtoffl01: Hmm.. maybe you don't even need these assertions -- like they're just checking that the env vars (that you're setting above) were set. But we can assume they were set because you set them 😆 .

The issue with keeping it is it makes the expected behavior when DD_SERVICE and DD_VERSION are not set ambiguous -- is it that val was just returned because of the get() function's default return, or was val returned because we really determined that's what the env var is set to?

Ultimately, this test may not be clear enough with the distinct behaviors we expect. The `env1` condition could theoretically pass on line 73/74, even though we'd want it to pass on 75/76 instead. You've made it all nice and succinct, but maybe the test will be clearer and more robust if you create separate functions for each condition.

mhlidd: I split the two cases into separate functions and directly check their conditions in the new commit.

tests/parametric/test_config_consistency.py
```diff
@@ -43,3 +44,39 @@ def test_tracing_disabled(self, library_env, test_agent, test_library):
...
+                pass
+
+        traces = test_agent.wait_for_num_traces(1)
+        assert len(traces) == 1
+
+        span = find_span_in_traces(traces, s1.trace_id, s1.span_id)
+        assert span["service"] != "version_test"
+        assert "version" not in span["meta"]
+
+    @parametrize("library_env", [{"DD_SERVICE": "version_test", "DD_VERSION": "5.2.0"}])
```
mtoffl01, line 65:
```suggestion
    # Assert that iff a span has service name set by DD_SERVICE, it also gets the version specified in DD_VERSION
    @parametrize("library_env", [{"DD_SERVICE": "version_test", "DD_VERSION": "5.2.0"}])
```

mtoffl01 (approved):

tests/parametric/test_config_consistency.py
```diff
@@ -43,3 +44,39 @@ def test_tracing_disabled(self, library_env, test_agent, test_library):
...
+
+        traces = test_agent.wait_for_num_traces(2)
+        assert len(traces) == 2
+
+        span1 = find_span_in_traces(traces, s1.trace_id, s1.span_id)
+        assert span1["service"] == "version_test"
+        assert span1["meta"]["version"] == "5.2.0"
+
+        span2 = find_span_in_traces(traces, s2.trace_id, s2.span_id)
+        assert span2["service"] != "version_test"
```
erikayasuda, line 81: nit:
```suggestion
        assert span2["service"] != "no dd_service"
```

tests/parametric/test_config_consistency.py
erikayasuda: @mtoffl01 Looks like there's a typo with `Assert that iff a span has ...`

erikayasuda (approved):

tests/parametric/test_config_consistency.py
mhlidd: If anything, shouldn't this be checking `assert span2["service"] == "no dd_service"`

erikayasuda: Oops, yes you're right! Let me fix it

tests/parametric/test_config_consistency.py
mtoffl01: The "iff"? I was using it on purpose, like [this](https://en.wikipedia.org/wiki/If_and_only_if) 😆

PR #3057, opened 2024-09-17
Configure Kind in Gitlab to allow kubernetes injection tests to run

randomanderson:
## Motivation

Previously, the kubernetes injection tests ran in Github Actions. With the move to *one-pipeline*, these tests need to run in Gitlab. Running Kind in gitlab doesn't work without the right incantations.

## Changes

* Using the correct ports/ip addresses for the `build` container to talk to the `kind` container
* Running kind in the `bridge` network instead of the `kind` network. Only the `bridge` network can communicate with the internet
* Changing `execute_command` to use `shlex` because the previous `command.split()` broke on any complicated command.

**There were _hundreds_ of variations and commits attempted to get this final solution. While there are only ~100 lines changed in this PR, there were months of work behind it**

## Reviewer checklist


robertomonteromiguel (approved):
Thanks Laplie!! Update your branch before merge, because there are some changes in the k8s lib injection that are not present on this PR. Probably you need to make some tweaks on the new tests (configure port and host)

PR #3060, opened 2024-09-17
Config Consistency Tests for DD_TRACE_<INTEGRATION>_ENABLED

link04:
## Motivation

Our APMAPI-284 Jira card where we need to write a test that validates the outcome of using the setting with both `true` and `false` values so that we can test if the auto instrumented library behaves as expected.

## Changes

- Included the MongoDB container to the `tracing_config_nondefault` scenario and added the env var with the value of `false` and  to `tracing_config_nondefault_2`with the value off `true`.
-  Added a new endpoint for .NET where I run a MongoDB query to check for the version of Mongo.
- Added two tests cases where I check for spans that belong to the expected Integration by checking the operation name for now(most consistent tag based on my research), and then assert on them depending on the test.

## Workflow

## Reviewer checklist


tests/test_config_consistency.py
```diff
@@ -71,3 +71,39 @@ def test_status_code_202(self):
...
         assert spans[0]["meta"]["http.status_code"] == "202"
         assert spans[0]["error"] == 1
+
+@scenarios.tracing_config_nondefault
+@features.tracing_configuration_consistency
+class Test_Config_IntegrationEnabled_False:
+    """ Verify behavior of integrations automatic spans """
+
+    def setup_mongodb_integration_enabled_false(self):
+        self.r = weblog.get("/integration_enabled_config")
```
zacharycmontoya, line 81: Two small comments here:
1. Can you rename this to something more explicit about what the endpoint is doing, like `/mongodb/getdatabase`?
2. Can you document the endpoint in `docs/weblog/README.md`?

link04: Addressed in: https://github.com/DataDog/system-tests/pull/3060/commits/957ea233b8fd4a3bdd64a6294a241eab6fb9f8eb

tests/test_config_consistency.py
```diff
@@ -237,3 +237,41 @@ def test_default_service_name(self):
...
+
+    def test_mongodb_integration_enabled_true(self):
+        assert self.r.status_code == 200
+
+        mongodbSpans = []
+        # We do not use get_spans: the span we look for is not directly the span that carry the request information
+        for data, trace in interfaces.library.get_traces(request=self.r):
+            mongodbSpans += [(data, span) for span in trace if span.get("name") == "mongodb.query"]
+
+        assert len(mongodbSpans) >= 1
```
khanayan123, line 277: same here, we should assert that non mongodb spans are created

khanayan123 (changes requested):

docs/weblog/README.md
```diff
@@ -695,3 +695,10 @@ Query parameters required in the `GET` method:
 
 Examples:
 - `GET`: `/session/user?sdk_user=sdkUser`
+
+### \[GET\] /mongodb/getdatabase
+
+This endpoint is used to test if automatic instrumentation creates or not MongoDB spans depending of the configuration passed to the service.
```
mabdinur, line 701: Do we need to add a new endpoint? Can we just enable/disable an integration and hit an existing endpoint

mabdinur: Adding this new endpoint to all weblog apps would be a pretty time consuming task

link04: Updated [here](https://github.com/DataDog/system-tests/pull/3060/commits/5c5392d93919668df67e9ff4e921ff22aad14395) to make use of the existing `/kafka/produce` endpoint all languages seem to have but PHP(will need to figure out something else for it later but I already got a plan. 

The rest we should address as people start enabling if there are irregularities between languages in the test/spans.

tests/test_config_consistency.py
link04: Added here: https://github.com/DataDog/system-tests/pull/3060/commits/5c5392d93919668df67e9ff4e921ff22aad14395

tests/test_config_consistency.py
```diff
@@ -237,3 +237,41 @@ def test_default_service_name(self):
...
+
+    def test_mongodb_integration_enabled_false(self):
+        assert self.r.status_code == 200
+
+        mongodbSpans = []
+        # We do not use get_spans: the span we look for is not directly the span that carry the request information
+        for data, trace in interfaces.library.get_traces(request=self.r):
+            mongodbSpans += [(data, span) for span in trace if span.get("name") == "mongodb.query"]
+
+        assert len(mongodbSpans) == 0
```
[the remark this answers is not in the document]
link04: Added here: https://github.com/DataDog/system-tests/pull/3060/commits/5c5392d93919668df67e9ff4e921ff22aad14395

khanayan123 (approved):

tests/test_config_consistency.py
```diff
@@ -276,3 +276,47 @@ def test_default_service_name(self):
...
+
+    def test_kafka_integration_enabled_false(self):
+        assert self.r.status_code == 200
+
+        nonKafkaSpans = []
+        kafkaSpans = []
+        # We do not use get_spans: the span we look for is not directly the span that carry the request information
+        for data, trace in interfaces.library.get_traces(request=self.r):
+            nonKafkaSpans += [(data, span) for span in trace if span.get("name") != "kafka.produce"]
+            kafkaSpans += [(data, span) for span in trace if span.get("name") == "kafka.produce"]
```
mabdinur, line 297: Nit: we can use `get_spans` we just need to set `full_trace` to `True` (it's false by default)
```suggestion
        for _,  _, span in interfaces.library.get_spans(request=self.r, full_trace=True):
            if if span.get("name") != "kafka.produce"
                 nonKafkaSpans.append(span)
           else:
                 kafkaSpans.append(span)
```

utils/_context/_scenarios/__init__.py
```diff
@@ -453,13 +453,18 @@ def all_endtoend_scenarios(test_object):
...
         },
+        include_kafka=True,
         doc="",
         scenario_groups=[ScenarioGroup.ESSENTIALS],
     )
 
     tracing_config_nondefault_3 = EndToEndScenario(
-        "TRACING_CONFIG_NONDEFAULT_3", weblog_env={"DD_TRACE_HTTP_CLIENT_TAG_QUERY_STRING": "false"}, doc="",
+        "TRACING_CONFIG_NONDEFAULT_3",
+        weblog_env={"DD_TRACE_HTTP_CLIENT_TAG_QUERY_STRING": "false", "DD_TRACE_Kafka_ENABLED": "true"},
```
mabdinur, line 465: nit: casing

```suggestion
        weblog_env={"DD_TRACE_HTTP_CLIENT_TAG_QUERY_STRING": "false", "DD_TRACE_KAFKA_ENABLED": "true"},
```

mabdinur, line 465: We should leave a comment stating that kafka is being used to validate  DD_TRACE_<INTEGRATION>_ENABLED configurations. There's nothing special about kafka, it just an integration that is consistent across all libraries (🤞).

manifests/cpp.yml
```diff
@@ -182,6 +182,8 @@ tests/:
     Test_Config_HttpClientErrorStatuses_FeatureFlagCustom: missing_feature
     Test_Config_HttpServerErrorStatuses_Default: missing_feature
     Test_Config_HttpServerErrorStatuses_FeatureFlagCustom: missing_feature
+    Test_Config_IntegrationEnabled_False: missing_feature
```
mabdinur, line 185: Can we look through the ci runs and confirm whether some of these tests are actually passing for some libraries.  It we be nice if we could add a descriptive message next to missing_feature 

ex: 
```suggestion
    # example reason, this is not accurate, please don't merge this suggestion
    Test_Config_IntegrationEnabled_False: missing_feature (integration name is confluent_kafka and not kafka)
```

mabdinur (approved):
left some nits but overall looks good to me

manifests/cpp.yml
link04: I think that would be counter productive for this PR, I thought the goal was that the person owning the  language should test both **telemetry** and the related **config** test run and then enable it for that language, that way they would find what is missing for the config they may need to change and they can lead their language procces.

utils/_context/_scenarios/__init__.py
link04: Indeed so the other languages need to add the env var as their tracer expects it as they enable the test(some languages may even have different integrations names).

bwoebi: @zacharycmontoya Well, the casing is required for the PHP extension to recognize the flag on Linux, but in uppercase.
If I read it correctly at least Java and Ruby are using full uppercase too. From a quick glance at docs I couldn't determine what exactly Python and node are doing.
So... we have to change this in any case.
I'd propose to universally move to fully uppercase.

bwoebi: It's not present in PHP though.

tests/test_config_consistency.py
```diff
@@ -303,3 +303,45 @@ def test_default_service_name(self):
         assert (
             spans[0]["service"] != "service_test"
         )  # in default scenario, DD_SERVICE is set to "weblog" in the dockerfile; this is a temp fix to test that it is not the value we manually set in the specific scenario
+
+
+@scenarios.tracing_config_nondefault
+@features.tracing_configuration_consistency
+class Test_Config_IntegrationEnabled_False:
```
cbeauchesne, line 310: Just a small cherry on the cake, if you could add `@rfc` with the RFC link to decorate the class. It's optionnal, and does not change anything for the execution, but it helps a lots other users

utils/_context/_scenarios/__init__.py
```diff
@@ -456,7 +456,9 @@ def all_endtoend_scenarios(test_object):
             "DD_APPSEC_ENABLED": "false",
             "DD_TRACE_HTTP_CLIENT_ERROR_STATUSES": "200-201,202",
             "DD_SERVICE": "service_test",
+            "DD_TRACE_Kafka_ENABLED": "false",  # Using Kafka as is the most common endpoint and integration.
```
cbeauchesne, line 459: Really lowercase ? 

```suggestion
            "DD_TRACE_KAFKA_ENABLED": "false",  # Using Kafka as is the most common endpoint and integration.
```

utils/_context/_scenarios/__init__.py
cbeauchesne: > I'd propose to universally move to fully uppercase. I find uppercase also more intuitive, in the midst of everything else being uppercase.

Agree, all our configuration env var are upper case. It looks weird to have one that mix both. By the why, is there a RFC for that ?

cbeauchesne (commented):
Testing : ok
CI: ok (failures are not related)

There is only the question of lowecase/uppercase. Not my decision here (even if I have a strong opinion 😉 ) So I let @bwoebi the charge to approve the PR.

utils/_context/_scenarios/__init__.py
link04: Yeah the plan was to use a different existing endpoint and env var for PHP since is the only one left out.

link04: So it was previously implemented like that, not sure I could find the RFC for that since is so old, but we have agreed that for these efforts of the [Config Consistency RFC](https://docs.google.com/document/d/1kI-gTAKghfcwI7YzKhqRv2ExUstcHqADIWA4-TZ387o/edit#heading=h.8v16cioi7qxp) it should be added that we always expected a all Upper cased env var.

tests/test_config_consistency.py
link04: Addressed in latest commit.

PR #3073, opened 2024-09-19
Khanayan123/implement dd trace agent url system test

khanayan123:
## Motivation

config consistency initiative

## Changes

implement dd trace agent url system test

### DIFF
--- a/manifests/cpp.yml
+++ b/manifests/cpp.yml
@@ -140,6 +140,7 @@ tests/:
       Test_Config_TraceEnabled: missing_feature
       Test_Config_TraceLogDirectory: missing_feature
       Test_Config_UnifiedServiceTagging: missing_feature
+      Test_Config_DDTraceAgentURL: missing_feature
     test_dynamic_configuration.py:
       TestDynamicConfigHeaderTags: missing_feature
     test_otel_api_interoperability.py: irrelevant (library does not implement OpenTelemetry)
--- a/manifests/cpp.yml
+++ b/manifests/cpp.yml
@@ -137,7 +137,7 @@ tests/:
       Test_Dsm_Manual_Checkpoint_Intra_Process: missing_feature
   parametric/:
     test_config_consistency.py:
-      Test_Config_DDTraceAgentURL: missing_feature
+      Test_Config_TraceAgentURL: missing_feature
       Test_Config_TraceEnabled: missing_feature
       Test_Config_TraceLogDirectory: missing_feature
       Test_Config_UnifiedServiceTagging: missing_feature
--- a/manifests/cpp.yml
+++ b/manifests/cpp.yml
@@ -137,10 +137,10 @@ tests/:
       Test_Dsm_Manual_Checkpoint_Intra_Process: missing_feature
   parametric/:
     test_config_consistency.py:
+      Test_Config_DDTraceAgentURL: missing_feature
       Test_Config_TraceEnabled: missing_feature
       Test_Config_TraceLogDirectory: missing_feature
       Test_Config_UnifiedServiceTagging: missing_feature
-      Test_Config_DDTraceAgentURL: missing_feature
     test_dynamic_configuration.py:
       TestDynamicConfigHeaderTags: missing_feature
     test_otel_api_interoperability.py: irrelevant (library does not implement OpenTelemetry)
--- a/manifests/nodejs.yml
+++ b/manifests/nodejs.yml
@@ -521,7 +521,7 @@ tests/:
       TestAdmisionControllerProfiling: *ref_5_22_0
   parametric/:
     test_config_consistency.py:
-      Test_Config_DDTraceAgentURL: *ref_5_23_0
+      Test_Config_TraceAgentURL: *ref_5_22_0
       Test_Config_TraceEnabled: missing_feature
       Test_Config_TraceLogDirectory: missing_feature
       Test_Config_UnifiedServiceTagging: missing_feature
--- a/manifests/nodejs.yml
+++ b/manifests/nodejs.yml
@@ -524,6 +524,7 @@ tests/:
       Test_Config_TraceEnabled: missing_feature
       Test_Config_TraceLogDirectory: missing_feature
       Test_Config_UnifiedServiceTagging: missing_feature
+      Test_Config_DDTraceAgentURL: *ref_5_23_0
     test_dynamic_configuration.py:
       TestDynamicConfigHeaderTags: missing_feature
       TestDynamicConfigSamplingRules: *ref_5_16_0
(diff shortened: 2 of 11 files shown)


tests/parametric/test_config_consistency.py
```diff
@@ -118,3 +118,43 @@ def test_specific_env(self, library_env, test_agent, test_library):
 
         span = find_span_in_traces(traces, s1.trace_id, s1.span_id)
         assert span["meta"]["env"] == "dev"
+
+
+@scenarios.parametric
+@features.tracing_configuration_consistency
+class Test_Config_DDTraceAgentURL:
```
zacharycmontoya, line 125: As a nit: I think removing the DD would fit the convention we've been assigning these names

```suggestion
class Test_Config_TraceAgentURL:
```

tests/parametric/test_config_consistency.py
```diff
@@ -118,3 +118,43 @@ def test_specific_env(self, library_env, test_agent, test_library):
...
+        "library_env",
+        [
+            {
+                "DD_TRACE_AGENT_URL": "unix:///var/run/datadog/apm.socket",
+                "DD_AGENT_HOST": "localhost",
+                "DD_AGENT_PORT": "8126",
+            }
+        ],
+    )
+    def test_dd_trace_agent_unix_url(self, library_env, test_agent, test_library):
```
zacharycmontoya, line 137: Nit: what do you think of appending to the test name so it's clear this is testing the case where the agent can't be found? That would make it more obvious why we're expecting failure in the test code. This same feedback applies to the http test case as well

```suggestion
    def test_dd_trace_agent_unix_url_nonexistent(self, library_env, test_agent, test_library):
```

zacharycmontoya (approved):
LGTM with some really minor nitpicks

link04 (changes requested):
Need to update all manifests to `Test_Config_TraceAgentURL` instead of `Test_Config_DDTraceAgentURL`, after that LGTM(passes for .NET).

manifests/cpp.yml
```diff
@@ -137,6 +137,7 @@ tests/:
       Test_Dsm_Manual_Checkpoint_Intra_Process: missing_feature
   parametric/:
     test_config_consistency.py:
+      Test_Config_DDTraceAgentURL: missing_feature
```
mabdinur, line 140: Do we need to set this to missing_feature? Can we just get this test to pass in this PR OR leave a descriptive message on why this test is failing (ex: `missing_feauture (cpp does not support uds)`

tests/parametric/test_config_consistency.py
```diff
@@ -118,3 +118,43 @@ def test_specific_env(self, library_env, test_agent, test_library):
...
+    )
+    def test_dd_trace_agent_unix_url_nonexistent(self, library_env, test_agent, test_library):
+        with test_library as t:
+            resp = t.get_tracer_config()
+        assert resp["dd_trace_agent_url"] == "unix:///var/run/datadog/apm.socket"
+        with pytest.raises(ValueError):
+            test_agent.wait_for_num_traces(num=1, clear=True)
+        assert (
+            True
+        ), "wait_for_num_traces raises an exception after waiting for 1 trace."  # wait_for_num_traces will throw an error if not received within 2 sec, so we expect to see an exception
```
mabdinur, line 145: You don't need this assertion. If a ValueError is not raised on line 141 then this test will fail. If you want to add an assertion you can check the error message in the value error and ensure it's do to a timeout/missing spans and not some other failure.

```suggestion
```

tests/parametric/test_config_consistency.py
```diff
@@ -118,3 +118,43 @@ def test_specific_env(self, library_env, test_agent, test_library):
...
+    )
+    def test_dd_trace_agent_http_url_nonexistent(self, library_env, test_agent, test_library):
+        with test_library as t:
+            resp = t.get_tracer_config()
+        assert resp["dd_trace_agent_url"] == "http://random-host:9999/"
+        with pytest.raises(ValueError):
+            test_agent.wait_for_num_traces(num=1, clear=True)
+        assert (
+            True
+        ), "wait_for_num_traces raises an exception after waiting for 1 trace."  # wait_for_num_traces will throw an error if not received within 2 sec, so we expect to see an exception
```
mabdinur, line 160: same comment as above

utils/parametric/_library_client.py
```diff
@@ -729,6 +730,7 @@ def get_tracer_config(self) -> Dict[str, Optional[str]]:
             "dd_trace_sample_ignore_parent": config_dict.get("dd_trace_sample_ignore_parent", None),
             "dd_env": config_dict.get("dd_env", None),
             "dd_version": config_dict.get("dd_version", None),
+            "dd_trace_agent_url": config_dict.get("dd_trace_agent_url", None),
```
mabdinur, line 733: Should we also modify the protos and grpc endpoints in this PR?

Even if this test isn't enabled for all libraries, we should write this test in a way where it can easily be enabled.

manifests/cpp.yml
cbeauchesne, line 140: Typo in naming:

```suggestion
      Test_Config_TraceAgentURL: missing_feature
```

cbeauchesne (changes requested):
A small typo to fix in all manifests

manifests/cpp.yml
cbeauchesne: You don't necessarily need to give a reason for `missing_feature`, but it may help later on, so if you have one that is not as obvious as "*we didn't started yet to work on this*" , feel free to add it.

manifests/nodejs.yml
```diff
@@ -517,6 +517,7 @@ tests/:
       TestAdmisionControllerProfiling: *ref_5_22_0
   parametric/:
     test_config_consistency.py:
+      Test_Config_DDTraceAgentURL: *ref_5_23_0
```
simon-id, line 520: was this feature just merged to master ? this version hasn't been released yet

utils/parametric/_library_client.py
khanayan123: What needs to be modified? for the existing proto:

rpc GetTraceConfig(GetTraceConfigArgs) returns (GetTraceConfigReturn) {}

GetTraceConfigReturn just returns a map<string, string> config 

so to enabled the test on the grpc end, parametrics apps just need to be updated to pass in the dd_trace_agent_url key in the map

manifests/cpp.yml
khanayan123: @mabdinur setting it to missing_feature as of now because the parametric client apps need to be updated to send dd_trace_agent_url

utils/parametric/_library_client.py
mabdinur: You're right we don't need to modify protos for this change.

link04 (approved):

cbeauchesne (approved):

PR #3086, opened 2024-09-20
[java] Enabled Java profiling for native-image binaries (tracing shared container)

MattAlp:
## Motivation

There is no E2E testing for the recently-added native-image-mode Java profiler.
## Changes

This modifies the `spring-boot-3-native` container (note- used by trace scenarios as well) to bake in and enable the Java profiler, as well as enabling the profiling scenarios for this container for `dd-trace-java v1.39.0+`.

It may be more correct to create two separate environments- one with the profiler enabled and one without, to test tracing and profiling functionality separately with different generated binaries.

## Workflow

## Reviewer checklist

### DIFF
--- a/utils/build/docker/java/spring-boot-3-native/pom.xml
+++ b/utils/build/docker/java/spring-boot-3-native/pom.xml
@@ -70,7 +70,7 @@
                         <artifactId>native-maven-plugin</artifactId>
                         <configuration>
                              <buildArgs>
-                                -J-javaagent:${agent.path} -J-Ddd.profiling.enabled=true --enable-monitoring=jfr
+                                -J-javaagent:${agent.path} -J-Ddd.profiling.enabled=true --enable-monitoring=jfr -Ob
                              </buildArgs>
                         </configuration>
                     </plugin>
@@ -91,7 +91,7 @@
                         <artifactId>native-maven-plugin</artifactId>
                         <configuration>
                              <buildArgs>
-                                -J-javaagent:${agent.path}
+                                -J-javaagent:${agent.path} -Ob
                              </buildArgs>
                         </configuration>
                     </plugin>


utils/build/docker/java/spring-boot-3-native/pom.xml
```diff
@@ -67,7 +67,7 @@
                 <artifactId>native-maven-plugin</artifactId>
                 <configuration>
                      <buildArgs>
-                        <buildArg>-J-javaagent:${agent.path}</buildArg>
+                        -J-javaagent:${agent.path} -J-Ddd.profiling.enabled=true --enable-monitoring=jfr
```
robertomonteromiguel, line 70: Do we really want to activate profiling all the time, and could it impact other scenarios?

robertomonteromiguel (changes requested):
Is there any assertion for profiling?

robertomonteromiguel (changes requested):
I'm not sure if I would advocate this strategy. We are launching the build twice, every time, whether we run this scenario or not. Looking at the times, it used to take 3 minutes to build this weblog, now it takes 6. What do you think @cbeauchesne?

cbeauchesne:
Well, the issue here is that we can't activate profiling at run time, it must be done at build time.

As these two steps are totally separated in system tests, there is not good solution.

1. Either we do a totally separated weblog, call `java-native-profiling`, with sole purpose of being tested on the PROFILING scenario
  * pro : in theory, does not add extra time, as it runs it parallel
  * con : CI cost
2. or we build a second jar with profiling  activated, and we use it at runtime based on the value of the env var
  * pro : totally transparant in the system-tests model
  * con : +3 mn in build time

Well, it's not like the native weblog is fast. `2.` is not perfect, but I'm in favor of that compromise.

Do you know if we could run the compilation in two parallel process to leverage the fact we have multiple core on all our machine ? 

```
RUN /opt/apache-maven-3.8.6/bin/mvn -Pnative,with-profiling native:compile && /opt/apache-maven-3.8.6/bin/mvn -Pnative,without-profiling native:compile
```

cbeauchesne:
> If a container can be built in a different way based on the associated scenario, we can avoid doing the unnecessary compilation.

We could imagine that, by adding an argument to `build.sh`. But as it's the only use case, I prefer to not open this pandora box 😄

robertomonteromiguel (approved):

PR #3105, opened 2024-09-23
implement dd_trace_rate_limit system test

khanayan123:
## Motivation

config consistency initiative

## Changes

implements rate limiting system tests for dd_trace_rate_limit environment variable

### DIFF
--- a/manifests/python.yml
+++ b/manifests/python.yml
@@ -685,7 +685,7 @@ tests/:
     test_config_consistency.py:
       Test_Config_RateLimit: v2.6.0
       Test_Config_TraceAgentURL: missing_feature
-      Test_Config_TraceEnabled: v2.6.0
+      Test_Config_TraceEnabled: missing_feature
       Test_Config_TraceLogDirectory: missing_feature
       Test_Config_UnifiedServiceTagging: missing_feature
     test_crashtracking.py:
--- a/manifests/python.yml
+++ b/manifests/python.yml
@@ -683,8 +683,9 @@ tests/:
     test_128_bit_traceids.py:
       Test_128_Bit_Traceids: v2.6.0
     test_config_consistency.py:
+      Test_Config_RateLimit: v2.6.0
       Test_Config_TraceAgentURL: missing_feature
-      Test_Config_TraceEnabled: missing_feature
+      Test_Config_TraceEnabled: v2.6.0
       Test_Config_TraceLogDirectory: missing_feature
       Test_Config_UnifiedServiceTagging: missing_feature
     test_crashtracking.py:
--- a/tests/parametric/test_config_consistency.py
+++ b/tests/parametric/test_config_consistency.py
@@ -2,7 +2,7 @@
 Test configuration consistency for features across supported APM SDKs.
 """
 import pytest
-from utils import scenarios, features, context, bug
+from utils import scenarios, features, context, bug, missing_feature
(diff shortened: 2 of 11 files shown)


tests/parametric/test_config_consistency.py
```diff
@@ -152,3 +152,31 @@ def test_dd_trace_agent_http_url_nonexistent(self, library_env, test_agent, test
...
+            resp = t.get_tracer_config()
+        assert resp["dd_trace_rate_limit"] == "100"
+
+    @parametrize("library_env", [{"DD_TRACE_RATE_LIMIT": "1"}])
+    def test_setting_trace_rate_limit(self, library_env, test_agent, test_library):
+        with test_library:
+            with test_library.start_span(name="s1") as s1:
+                pass
+            with test_library.start_span(name="s2") as s2:
+                pass
```
mtoffl01, line 176: Is it safe to assume these spans will both be generated in <1s?

mtoffl01 (approved):
lgtm, just some comments

manifests/python.yml
```diff
@@ -683,8 +683,9 @@ tests/:
     test_128_bit_traceids.py:
       Test_128_Bit_Traceids: v2.6.0
     test_config_consistency.py:
+      Test_Config_RateLimit: v2.6.0
       Test_Config_TraceAgentURL: missing_feature
-      Test_Config_TraceEnabled: missing_feature
+      Test_Config_TraceEnabled: v2.6.0
```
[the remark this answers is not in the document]
khanayan123: It was since the test was passing, but probably best to remove it from this PR and do a separate enabling one

tests/parametric/test_config_consistency.py
khanayan123: Same response as in the thread:

"Yeah so it might be flaky? i guess we'll see how other libraries handle it. Seems to be fine for Python, but I don't really know how else to test it since the accepted input for the variable is a positive integer and 1 is the lowest accepted value"

manifests/python.yml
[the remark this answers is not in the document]
mabdinur: I can enable it my PR
```suggestion
```

utils/build/docker/python/parametric/apm_test_client/server.py
```diff
@@ -136,6 +136,7 @@ def trace_config() -> TraceConfigReturn:
             "dd_trace_sample_ignore_parent": None,
             "dd_env": config.env,
             "dd_version": config.version,
+            "dd_trace_rate_limit": str(config._trace_rate_limit),
```
mabdinur, line 139: Can we make this change to all the parametric apps. That way we can easily see which libraries pass this test and which don't. It should be a simple change in most libraries.

khanayan123: Let me do it in a separate PR. I want to keep this PR scoped to the system test

mabdinur: We should confirm whether this system test is written in a way where it is only compatible with python. It would be nice to know now if all/most libraries store the traceagent url in this format `http://random-host:9999/`. 

This isn't a blocker for this PR but moving forward, we should try to confirm the validity of the shared tests by running the test against 1 or more libraries

mabdinur (approved):

PR #3194, opened 2024-10-08
Test introduction demo

colin-higgins:
Overview demo of introducing tests and the capabilities of the framework.

### DIFF
--- a/tests/test_demo.py
+++ b/tests/test_demo.py
@@ -1,0 +1,38 @@
+# Unless explicitly stated otherwise all files in this repository are licensed under the the Apache License Version 2.0.
+# This product includes software developed at Datadog (https://www.datadoghq.com/).
+# Copyright 2022 Datadog, Inc.
+
+from utils import weblog, interfaces, features, scenarios
+
+
+#If your test verifies a feature, make sure you represent it
+# @features.unix_domain_sockets_support_for_traces
+@features.not_reported # This hides your feature from the jaws of the Feature Parity Dashboard
+@scenarios.my_nice_scenario # Do you need this? Maybe? Only you can decide.
+class Test_Demo:
+    """ This is a place to describe the purpose of the test """
+
+    def setup_very_nice_system_test_has_web_span(self):
+        self.r = weblog.get("/my-cool-variable")
+
+    def test_very_nice_system_test_has_web_span(self):
+        # You can define multiple tests, but make sure they each have a setup :)
+        interfaces.library.assert_trace_exists(self.r, span_type="web")
+
+    def setup_very_nice_system_test_has_cool_tag_on_span(self):
+        self.r = weblog.get("/my-cool-variable")
+        
+    def test_very_nice_system_test_has_cool_tag_on_span(self):
+        span_count = 0
+        for _, _, span in interfaces.library.get_spans(request=self.r):
+            print(" !Span inspection! ")
+            print(span)
+            span_count += 1
+
+        if span_count > 1:
+            raise ValueError(f"Oh no, this endpoint should only have one span.")
+
+        cool_tag = span["meta"]["DD_WOW_WOW"]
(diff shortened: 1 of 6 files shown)


tests/test_demo.py
```diff
@@ -0,0 +1,38 @@
...
+        # You can define multiple tests, but make sure they each have a setup :)
+        interfaces.library.assert_trace_exists(self.r, span_type="web")
+
+    def setup_very_nice_system_test_has_cool_tag_on_span(self):
+        self.r = weblog.get("/my-cool-variable")
+
+    def test_very_nice_system_test_has_cool_tag_on_span(self):
+        span_count = 0
+        for _, _, span in interfaces.library.get_spans(request=self.r):
+            print(" !Span inspection! ")
```
cbeauchesne, line 28: At the import section of the file : 

```python
from utils.tools import logger
```

Then here : 

```suggestion
            logger.info(" !Span inspection! ")  # can be info, debug, error... This will be printed out in case of failure of the test. 
```

tests/test_demo.py
```diff
@@ -0,0 +1,38 @@
...
+            print(span)
+            span_count += 1
+
+        if span_count > 1:
+            raise ValueError(f"Oh no, this endpoint should only have one span.")
```
cbeauchesne, line 33: pytest has a very fancy tooling around assertion that display in the output more details (in that case, it will shows the value of `span_count`. For more complex assertion, it really does some magic.

```suggestion
        assert span_count <= 1, "Oh no, this endpoint should only have one span."
```

tests/test_demo.py
```diff
@@ -0,0 +1,38 @@
...
+
+        cool_tag = span["meta"]["DD_WOW_WOW"]
+        cool_tag_expectation = "wow wow wee wow"
+        if cool_tag != cool_tag_expectation:
+            raise ValueError(f"Expected {cool_tag_expectation} but got {cool_tag}")
```
cbeauchesne, line 38:
```suggestion
        assert cool_tag == cool_tag_expectation,  "Expected {cool_tag_expectation} but got {cool_tag}"
```

utils/_context/_scenarios/__init__.py
```diff
@@ -693,6 +693,14 @@ def all_endtoend_scenarios(test_object):
 
     external_processing = ExternalProcessingScenario("EXTERNAL_PROCESSING")
 
+    # It's VERY important to match the exact name of the property and the name
+    my_nice_scenario = EndToEndScenario(
+        name="MY_NICE_SCENARIO",
+        weblog_env={"DD_VERY_NICE_FEATURE": "wow wow wee wow",},
+        scenario_groups=[ScenarioGroup.VERY_COOL_SCENARIO_GROUP],
```
cbeauchesne, line 700: This feature is optional, and probably too secondary regarding other point you'll present, I suggest not having that.

colin-higgins: Removing scenario groups, you mean?

tests/test_demo.py
```diff
@@ -0,0 +1,36 @@
...
+        self.r = weblog.get("/my-cool-variable")
+
+    def test_very_nice_system_test_has_cool_tag_on_span(self):
+        span_count = 0
+        for _, _, span in interfaces.library.get_spans(request=self.r):
+            # can be info, debug, error... This will be printed out in case of failure of the test.
+            logger.info(f" !Span inspection! {span}")
+            span_count += 1
+
+        assert span_count <= 1, "Oh no, this endpoint should only have one span."
```
cbeauchesne, line 32: Just for having an assertion in line with the comment

```suggestion
        assert span_count == 1, "Oh no, this endpoint should only have one span."
```

cbeauchesne:
@colin-higgins ,

Can we close this PR ?

PR #3231, opened 2024-10-14
OnBoarding: Bug marker

robertomonteromiguel:
## Motivation
We need a bug marker for onboarding, in order to skip test by virtual machine properties

## Workflow

## Reviewer checklist

### DIFF
--- a/utils/virtual_machine/utils.py
+++ b/utils/virtual_machine/utils.py
@@ -13,6 +13,8 @@ def parametrize_virtual_machines(bugs: list[dict] = None):
     *     @parametrize_virtual_machines(bugs=[{"vm_cpu":"amd64", "weblog_variant":"app1", "reason": "APMON-1576"}])
     - Marks as bug for all vm with cpu type "amd64" and all weblogs
     *     @parametrize_virtual_machines(bugs=[{"vm_cpu":"amd64","reason": "APMON-1576"}])
+    - Marks as bug for all vms that belong to "debian" and all weblogs
+    *     @parametrize_virtual_machines(bugs=[{"vm_branch":"debian","reason": "APMON-1576"}])
     Reason is mandatory for each bug and it MUST reference to a JIRA ticket.
     """
 
@@ -26,6 +28,7 @@ def parametrize_virtual_machines(bugs: list[dict] = None):
                 for bug in bugs:
                     if (
                         (not "vm_name" in bug or vm.name == bug["vm_name"])
+                        and (not "vm_branch" in bug or vm.os_branch == bug["vm_branch"])
                         and (not "vm_cpu" in bug or vm.os_cpu == bug["vm_cpu"])
                         and (not "weblog_variant" in bug or context.weblog_variant == bug["weblog_variant"])
                     ):
@@ -39,6 +42,8 @@ def parametrize_virtual_machines(bugs: list[dict] = None):
             if bug_found == False:
                 parameters.append(vm)
 
-        return pytest.mark.parametrize("virtual_machine", parameters)(func)
+        return pytest.mark.parametrize(
+            "virtual_machine", parameters, ids=getattr(context.scenario, "required_vm_names", [])
+        )(func)
 
     return decorator
--- a/utils/virtual_machine/utils.py
+++ b/utils/virtual_machine/utils.py
@@ -24,7 +24,7 @@ def parametrize_virtual_machines(bugs: list[dict] = None):
             if bugs and hasattr(bugs, "__len__") and len(bugs) > 0:
                 for bug in bugs:
                     if (
-                        ("vm_name" not in bug or vm.name == bug["vm_name"])
+                        (not "vm_name" in bug or vm.name == bug["vm_name"])
                         and (not "vm_cpu" in bug or vm.os_cpu == bug["vm_cpu"])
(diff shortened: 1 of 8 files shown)


utils/virtual_machine/utils.py
```diff
@@ -0,0 +1,40 @@
...
+        parameters = []
+
+        for vm in getattr(context.scenario, "required_vms", []):
+            bug_found = False
+            if bugs and hasattr(bugs, "__len__") and len(bugs) > 0:
+                for bug in bugs:
+                    if (
+                        ("vm_name" not in bug or vm.name == bug["vm_name"])
+                        and (not "vm_cpu" in bug or vm.os_cpu == bug["vm_cpu"])
+                        and (not "weblog_variant" in bug or context.weblog_variant == bug["weblog_variant"])
```
szegedi, line 29: maybe just call this `weblog`? Not sure if `weblog_variant` conveys additional meaning.

utils/virtual_machine/utils.py
```diff
@@ -0,0 +1,40 @@
...
+    def decorator(func):
+
+        parameters = []
+
+        for vm in getattr(context.scenario, "required_vms", []):
+            bug_found = False
+            if bugs and hasattr(bugs, "__len__") and len(bugs) > 0:
+                for bug in bugs:
+                    if (
+                        ("vm_name" not in bug or vm.name == bug["vm_name"])
```
szegedi, line 27:
```suggestion
                        (not "vm_name" in bug or vm.name == bug["vm_name"])
```
for consistency with the other two?

utils/virtual_machine/utils.py
```diff
@@ -0,0 +1,40 @@
...
+                        and (not "vm_cpu" in bug or vm.os_cpu == bug["vm_cpu"])
+                        and (not "weblog_variant" in bug or context.weblog_variant == bug["weblog_variant"])
+                    ):
```
szegedi, line 30: Maybe I'm daft, but isn't the predicate almost-but-not-exactly inverted here? I would expect that the condition for marking the test as known failure would be:
```suggestion
                        ("vm_name" in bug and vm.name == bug["vm_name"])
                        or ("vm_cpu" in bug and vm.os_cpu == bug["vm_cpu"])
                        or ("weblog_variant" in bug and context.weblog_variant == bug["weblog_variant"])
                    ):
```
I'm sure it's just me missing something here.

szegedi (changes requested):

utils/virtual_machine/utils.py
```diff
@@ -0,0 +1,40 @@
...
+                        parameters.append(pytest.param(vm, marks=pytest.mark.xfail(reason=f"bug: {bug['reason']}")))
```
cbeauchesne, line 31: Could you add a check on the reason, to ensure it's a JIRA ticket ? 

See https://github.com/DataDog/system-tests/blob/main/utils/_decorators.py#L12

The good way is to  probably to add a function `ensure_jira_ticket(reason:str)` that raise a `ValueError` and reuse it.

cbeauchesne (changes requested):

utils/virtual_machine/utils.py
robertomonteromiguel: I'm using the same name (weblog_variant) as other decorators:
` @flaky(weblog_variant="test-app-java-buildpack", reason="APMON-1595")`

utils/virtual_machine/utils.py
robertomonteromiguel: I think it's ok. But please, confirm me this after this explanation:

for example:
* mark as bug all machines with cpu==amd64 AND weblog_variant=="xyz"
`bugs=[{"vm_cpu": "amd64", "weblog_variant":"xyz", "reason": "APMON-1576"}]`

* mark as bug all machines with vm_name="vm1" AND cpu==amd64 AND weblog_variant=="xyz"
`bugs=[{"vm_name": "vm1", "vm_cpu": "amd64", "weblog_variant":"xyz", "reason": "APMON-1576"}]`
I mean, if I have a machine "vm1" with cpu "arm64" is going to execute the test

utils/virtual_machine/utils.py
cbeauchesne: I don't know if `weblog` is a better name (probably `variant` is a better one, it's more generic, and this abstraction may not concerns only `weblog`). 

But for this particular PR, I agree with Roberto : continue to use the one used elsewhere to not add any entropy. And if we feel it should change, adress this globally in another PR.

I pick my entire part of responsibility here, I've never settled for a name.

utils/virtual_machine/utils.py
robertomonteromiguel: done!

utils/virtual_machine/utils.py
szegedi: Makes sense – I didn't realize it is used as it is elsewhere. Thanks for the explanation!

utils/virtual_machine/utils.py
szegedi: Oh, got it – if there _is_ a bug marker, then the vm needs to match all specified fields to be a known failing one, and it of course implicitly matches any unspecified fields. Thanks for bearing with me!

szegedi (approved):

PR #3243, opened 2024-10-15
add smoke test for crash logs from python [INPLAT-36]

emmettbutler:
This pull request adds a "smoke test" verifying the Crash Tracking functionality. It takes the approach of requesting an endpoint known to trigger a segmentation fault in the weblog app, then querying test agent for logs generated by Crash Tracking for the particular Runtime ID being tested.

## Reviewer checklist


cbeauchesne (dismissed):

cbeauchesne (dismissed):
There are some lint errors. Also, a class misses feature decorator (see `docs/edit/features.md`). Please ping me on slack if you need any help.

tests/auto_inject/test_auto_inject_install.py
```diff
@@ -75,6 +74,15 @@ def test_profiling(self, virtual_machine):
         self._test_install(virtual_machine, profile=True)
 
 
+@features.crashtracking
```
brettlangdon, line 77: Do we also want `@features.ssi_crashtracking` ?

or maybe that instead of `@features.crashtracking`?

Trying to think how we'll tie this to the reporting we need to do, the `ssi_crashtracking` feature has both "ssi" and "crashtracking" tags.


also 👍🏻 for adding this but as `@missing_feature`, that is smart.

utils/interfaces/_test_agent.py
```diff
@@ -56,6 +56,17 @@ def get_telemetry_for_runtime(self, runtime_id):
...
+    def get_crashlog_for_runtime(self, runtime_id):
+        logger.debug(f"Try to find a crashlog related to runtime-id {runtime_id}")
+        assert runtime_id is not None, "Runtime ID not found"
+        telemetry_msgs = []
+        for data_received in self._data_telemetry_list:
+            print(data_received)
+            if data_received["request_type"] == "logs" and data_received["runtime_id"] == runtime_id:
+                telemetry_msgs.append(data_received)
+
+        return telemetry_msgs
```
brettlangdon, line 68:
```suggestion
        return [l for l in self.get_telemetry_logs() if l["runtime_id"] == runtime_id]
```

I think this is the same?


I didn't realize we had duplicated some helpers, the `get_crash_reports` helper below also checks if we have crash tags, otherwise we might just be grabbing some random telemetry logs that aren't crash reports.


Not sure exactly how much this matters, but .... 🤷🏻

brettlangdon (approved):
very minor comments/nits, otherwise lgtm

tests/auto_inject/test_auto_inject_install.py
```diff
@@ -75,6 +74,15 @@ def test_profiling(self, virtual_machine):
...
+@scenarios.container_auto_injection_install_script_crashtracking
+class TestContainerAutoInjectInstallScriptCrashTracking(base.AutoInjectBaseTest):
+    @parametrize_virtual_machines()
+    @missing_feature(reason="Crash logs not currently forwarded to org 749096")
```
robertomonteromiguel, line 81: But we are testing docker ssi, this test class if for the aws ssi. It doesn't make sense in this PR

utils/_context/_scenarios/docker_ssi.py
```diff
@@ -141,8 +141,8 @@ def fill_context(self, json_tested_components):
             logger.stdout(f"{key}: {self._tested_components[key]}")
 
     def post_setup(self):
-        logger.stdout("--- Waiting for all traces to be sent to test agent ---")
-        time.sleep(5)  # wait for the traces to be sent to the test agent
+        logger.stdout("--- Waiting for all traces and telemetry to be sent to test agent ---")
+        time.sleep(60)  # wait for the traces and telemetry to be sent to the test agent
```
robertomonteromiguel, line 145: I think waiting for 60 is too much. In most cases, this is not necessary
Could you try to make it configurable for this case?

utils/_context/_scenarios/__init__.py
```diff
@@ -630,6 +629,14 @@ def all_endtoend_scenarios(test_object):
         github_workflow="libinjection",
     )
 
+    container_auto_injection_install_script_crashtracking = InstallerAutoInjectionScenario(
```
robertomonteromiguel, line 632: This is for aws tests. It doesn't make sense in this PR

robertomonteromiguel (changes requested):

utils/_context/_scenarios/docker_ssi.py
brettlangdon: We have to wait long enough for telemetry data to flush, which could be up to 60 seconds :/

If we are able to shut down the app, we can probably force a flush.

utils/_context/_scenarios/docker_ssi.py
```diff
@@ -142,8 +142,11 @@ def fill_context(self, json_tested_components):
 
     def post_setup(self):
         logger.stdout("--- Waiting for all traces and telemetry to be sent to test agent ---")
-        time.sleep(60)  # wait for the traces and telemetry to be sent to the test agent
-        interfaces.test_agent.collect_data(f"{self.host_log_folder}/interfaces/test_agent")
+        data = None
+        attempts = 0
+        while attempts < 30 and not data:
+            data = interfaces.test_agent.collect_data(f"{self.host_log_folder}/interfaces/test_agent")
```
brettlangdon, line 149: this might not be enough, since we might get a trace, but not telemetry data?

emmettbutler: The tests are passing on CI so it seems like it's sufficient for the moment.

utils/_context/_scenarios/__init__.py
```diff
@@ -142,38 +142,38 @@ def all_endtoend_scenarios(test_object):
...
-        appsec_rules="/appsec_blocking_rule.json",
+        weblog_env={"DD_APPSEC_RULES": "/appsec_blocking_rule.json"},
         doc="AppSec tests for GraphQL integrations",
         github_workflow="graphql",
         scenario_groups=[ScenarioGroup.APPSEC],
     )
     appsec_rules_monitoring_with_errors = EndToEndScenario(
         "APPSEC_RULES_MONITORING_WITH_ERRORS",
-        appsec_rules="/appsec_custom_rules_with_errors.json",
+        weblog_env={"DD_APPSEC_RULES": "/appsec_custom_rules_with_errors.json"},
```
robertomonteromiguel, line 176: why are we updating this?¿

robertomonteromiguel (changes requested):

brettlangdon (approved):

robertomonteromiguel (approved):

PR #3282, opened 2024-10-22
[ruby] Add 1.x support in Ruby weblogs

vpellan:
## Motivation

I wanted to enable easy wins, but also wanted to add the version since it works in the manifest. Trying to run system-tests on 1.x branch was not working. 1.x branch is still active and should still be tested against system-tests

## Changes

Both auto instrumentation and healthcheck endpoints were using `Datadog` namespace, but 1.x is using `DDTrace`. I applied what has been done in Sinatra weblogs and parametric weblog for auto-instrumentation (require in begin/rescue block), and added a condition for getting the version in healthcheck endpoints (`defined?(Datadog::VERSION)`)

## Reviewer checklist

### DIFF
--- a/utils/build/docker/ruby/rack/config.ru
+++ b/utils/build/docker/ruby/rack/config.ru
@@ -7,7 +7,15 @@ require 'json'
 
 # tracer configuration of Rack integration
 
-require 'datadog/auto_instrument'
+begin
+  require 'ddtrace/auto_instrument'
+rescue LoadError
+end
+
+begin
+  require 'datadog/auto_instrument'
+rescue LoadError
+end
 require 'datadog/kit/appsec/events'
 
 Datadog.configure do |c|
--- a/utils/build/docker/ruby/rack/config.ru
+++ b/utils/build/docker/ruby/rack/config.ru
@@ -8,13 +8,9 @@ require 'json'
 # tracer configuration of Rack integration
 
 begin
-  require 'ddtrace/auto_instrument'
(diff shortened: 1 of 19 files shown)


Strech (approved):
👍🏼

cbeauchesne (approved):

utils/build/docker/ruby/rack/config.ru
```diff
@@ -7,7 +7,15 @@ require 'json'
...
-require 'datadog/auto_instrument'
+begin
+  require 'ddtrace/auto_instrument'
+rescue LoadError
+end
+
+begin
+  require 'datadog/auto_instrument'
+rescue LoadError
+end
```
y9v, line 18: I think we should unify those:

```ruby
begin
  require 'datadog/auto_instrument'
rescue LoadError
  require 'ddtrace/auto_instrument'
end
```

This was we know that at least one of them was loaded (or another `LoadError` will be raised)

y9v (changes requested):

utils/build/docker/ruby/rack/config.ru
vpellan: Thanks, I'll do that in Sinatra and parametric weblogs too then

vpellan: Fixed in 801bd8ab996e8465a2fe20ddcebaebc55a15595c

y9v (approved):

PR #3284, opened 2024-10-22
[python]attempt to enable telemetry tests for python [INPLAT-36]

emmettbutler:
Adds smoke tests for auto-injection telemetry from dd-trace-py. Currently skipped as `bug` due to https://github.com/DataDog/dd-trace-py/pull/11101

## Reviewer checklist


cbeauchesne (commented):
I think you need to define python weblogs in `utils/docker_ssi/docker_ssi_definitions.py`

cc @robertomonteromiguel

robertomonteromiguel:
Yes, I think we can set this PR as draft, because we need to add the python scripts, docker files....
@iamluc is looking that too

emmettbutler:
@robertomonteromiguel let me know if this looks right to you. I believe I've got it all working with the exception of the tests actually passing, since there's a known issue with Python wherein it doesn't send SSI telemetry (iirc).

robertomonteromiguel (changes requested):
There is still one error on CI

PR #3295, opened 2024-10-23
[ruby] Add -dev to healthcheck when not using gem from Rubygems

vpellan:
## Motivation

This will enable the activation of system-tests on ruby dev environment for the upcoming version (e.g.: current version is 2.4.0, healthcheck will return 2.4.0-dev and system-tests will increment the patch number, so tests can be activated by declaring 2.4.1-dev in the manifest)

## Changes

This adds -dev to the version given by healthchecks when not using datadog/ddtrace gem from Rubygems by verifying that the source of a Rubygems is a Bundler::Source::Rubygems object

## Reviewer checklist

### DIFF
--- a/utils/build/docker/ruby/graphql23/app/controllers/system_test_controller.rb
+++ b/utils/build/docker/ruby/graphql23/app/controllers/system_test_controller.rb
@@ -8,11 +8,14 @@ class SystemTestController < ApplicationController
   end
 
   def healthcheck
+    gemspec = Gem.loaded_specs['datadog'] || Gem.loaded_specs['ddtrace']
+    version = gemspec.version.to_s
+    version = "#{version}-dev" unless gemspec.source.is_a?(Bundler::Source::Rubygems)
     render json: { 
       status: 'ok',
       library: {
         language: 'ruby',
-        version: defined?(Datadog::VERSION) ? Datadog::VERSION::STRING : DDTrace::VERSION::STRING
+        version: version
       }
     }
   end
--- a/utils/build/docker/ruby/rack/config.ru
+++ b/utils/build/docker/ruby/rack/config.ru
@@ -61,11 +61,14 @@ end
 # /healthcheck
 class Healthcheck
   def self.run
+    gemspec = Gem.loaded_specs['datadog'] || Gem.loaded_specs['ddtrace']
+    version = gemspec.version.to_s
+    version = "#{version}-dev" unless gemspec.source.is_a?(Bundler::Source::Rubygems)
     response = {
       status: 'ok',
       library: {
         language: 'ruby',
-        version: defined?(Datadog::VERSION) ? Datadog::VERSION::STRING : DDTrace::VERSION::STRING
+        version: version
       }
     }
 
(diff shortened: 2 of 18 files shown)


utils/build/docker/ruby/graphql23/app/controllers/system_test_controller.rb
```diff
@@ -8,11 +8,22 @@ def root
   end
 
   def healthcheck
+    version = if defined?(Datadog::VERSION)
+                if Gem.loaded_specs['datadog'].source.is_a?(Bundler::Source::Rubygems)
```
y9v, line 12: I am afraid this will raise a `NoMethodError` if the gem is not loaded:
```ruby
> Gem.loaded_specs['datadog'].source
(irb):1:in `<main>': undefined method `source' for nil (NoMethodError)
```

y9v (changes requested):

utils/build/docker/ruby/graphql23/app/controllers/system_test_controller.rb
vpellan: But if Datadog::VERSION is defined, it means that the gem is loaded ?

vpellan: To give more context: We test the library so it should always be loaded. If Datadog::VERSION is not defined, that means that DDTrace::VERSION is defined (and we're on the 1.x branch) If none is defined then the library is not loaded and every tests will fail anyway

Strech: The logic here is the following:

1. `Datadog` is a namespace which exists in both cases
2. `Datadog::VERSION` exists only when `datadog` gem is loaded and only in new version
3. Otherwise it's `ddtrace` and we know for sure it's also loaded

hence I think we can proceed with those changes.

utils/build/docker/ruby/graphql23/app/controllers/system_test_controller.rb
```diff
@@ -8,11 +8,22 @@ def root
...
+                  Datadog::VERSION::STRING
+                else
+                  "#{Datadog::VERSION::STRING}-dev"
+                end
+              elsif Gem.loaded_specs['ddtrace'].source.is_a?(Bundler::Source::Rubygems)
+                DDTrace::VERSION::STRING
+              else
+                "#{DDTrace::VERSION::STRING}-dev"
+              end
```
Strech, line 21: I have only suggestion to rewrite it into this

```suggestion
    gemspec = defined?(Datadog::VERSION) ? Gem.loaded_specs['datadog'] : Gem.loaded_specs['ddtrace']

    version = gemspec.version.to_s
    version = "#{version}-dev" unless gemspec.source.is_a?(Bundler::Source::Rubygems)
```

Strech (approved):
I think the logic is correct. Well done 👏🏼 

I also think we might use solely gemspec to return the version and to reduce the code.

utils/build/docker/ruby/graphql23/app/controllers/system_test_controller.rb
vpellan: Thanks ! I applied your suggestion (and used an OR to get the gemspec) in f1a9c15ed552e0709fa6e6d02c2b8d2ab76c1627

PR #3305, opened 2024-10-24
Add Suspicious Request Blocking test without path_params

vpellan:
## Motivation

Rack does not send anything to the path_params address (Rails and Sinatra does) but we still want to test blocking on multiple request addresses

## Changes

On the advice of @christophe-papazian I wrote a similar test without path-params. All the weblogs that passes these tests with path_params should also pass them without, but it will also enable testing that feature on a pure Rack app 

## Reviewer checklist

### DIFF
--- a/tests/appsec/blocking_rule.json
+++ b/tests/appsec/blocking_rule.json
@@ -576,6 +576,76 @@
       ]
     },
     {
+      "id": "tst-037-013",
+      "name": "Test block on multiple request addresses (without path_params)",
+      "tags": {
+        "type": "lfi",
+        "crs_id": "000013",
+        "category": "attack_attempt"
+      },
+      "conditions": [
+        {
+          "parameters": {
+            "inputs": [
+              {
+                "address": "server.request.method"
+              }
+            ],
+            "regex": "GET"
+          },
+          "operator": "match_regex"
+        },
+        {
+          "parameters": {
+            "inputs": [
+              {
+                "address": "server.request.uri.raw"
+              }
+            ],
+            "regex": "wX1GdUiWdVdoklf0pYBi5kQApO9i77tN"
+          },
+          "operator": "match_regex"
+        },
+        {
+          "parameters": {
+            "inputs": [
+              {
+                "address": "server.request.query"
+              }
+            ],
+            "regex": "T3d1nKdkTWIG03q03ix9c9UlhbGigvwQ"
+          },
+          "operator": "match_regex"
+        },
+        {
+          "parameters": {
+            "inputs": [
+              {
+                "address": "server.request.headers.no_cookies"
+              }
+            ],
+            "regex": "siDzyETAdkvKahD3PxlvIqcE0fMIVywE"
+          },
+          "operator": "match_regex"
+        },
+        {
+          "parameters": {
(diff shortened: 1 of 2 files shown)


cbeauchesne (dismissed):
`./format.sh` is your friend

vpellan:
> `./format.sh` is your friend

Sorry, fixed in 0ebb371122140c5bbe78cccb9f5f3b11b44ddaa7

tests/appsec/test_blocking_addresses.py
```diff
@@ -549,6 +549,41 @@ def test_blocking_before(self):
...
+    def test_blocking_before_without_path_params(self):
+        """Test that blocked requests are blocked before being processed"""
+        # first request should not block and must set the tag in span accordingly
+        assert self.set_req1.status_code == 200
+        assert self.set_req1.text == "Value tagged"
+        interfaces.library.validate_spans(self.set_req1, _assert_custom_event_tag_presence("clean_value_3882"))
+        """Test that blocked requests are blocked before being processed"""
+        assert self.block_req2.status_code == 403
+        interfaces.library.assert_waf_attack(self.block_req2, rule="tst-037-013")
+        interfaces.library.validate_spans(self.block_req2, _assert_custom_event_tag_absence())
```
Strech, line 585:
```suggestion
        
        """Test that blocked requests are blocked before being processed"""
        assert self.block_req2.status_code == 403
        interfaces.library.assert_waf_attack(self.block_req2, rule="tst-037-013")
        interfaces.library.validate_spans(self.block_req2, _assert_custom_event_tag_absence())
```

tests/appsec/blocking_rule.json
```diff
@@ -575,6 +575,76 @@
...
+          "operator": "match_regex"
+        },
+        {
+          "parameters": {
+            "inputs": [
+              {
+                "address": "server.request.uri.raw"
+              }
+            ],
+            "regex": "wX1GdUiWdVdoklf0pYBi5kQApO9i77tN"
```
Strech, line 605: To be honest, I think cryptic values make it hard to read the tests itself. Instead I would like to see something, which will immediately give me a sign in the tests, that I see a value, which suppose to trigger something.

Ideas:

1. malicious-uri
2. security-violation-uri
3. is-attack-uri
4. ...

I know that naming is hard, but you will see how tests change after that

Strech (changes requested):

tests/appsec/test_blocking_addresses.py
vpellan: Fixed in 2bfd6af0f8ad4f75847d4debbc025a1ef312f67f

tests/appsec/blocking_rule.json
Strech: Alright, then we can add the "salt" to the good-meaning value, like

```
malicious-uri-wX1GdUiWdVdoklf0pYBi5kQApO9i77tN
```

Strech (approved):
👏🏼 Well done. Let's goooo!

robertomonteromiguel (approved):

cbeauchesne (approved):

PR #3300, opened 2024-10-23
parametric: simplify start span

mabdinur:
## Motivation

- Simplify the start span parametric endpoint
- Adds a `parametric_endpoint_parity` feature to track which parametric endpoints are supported in each application.
  - Adds test cases for the start_span and the new extract_header endpoints. Tests for all supported parametric endpoints will be added in a future PR.

## Changes

- Removes the following arguments from the `/trace/span/start` parametric endpoint:
  - http_headers
  - links
  - tags
- Moves HTTP header extraction from `/trace/span/start` endpoint to a new `/trace/span/extract_headers` 
- Updates all tests to use `/trace/span/add_link` to add a span link to a span and `/trace/span/extract_headers` to extract distributed tracing headers.

## Workflow

1. ⚠️ Create your PR as draft ⚠️
2. Work on you PR until the CI passes (if something not related to your task is failing, you can ignore it)
3. Mark it as ready for review
    * Test logic is modified? -> Get a review from RFC owner. We're working on refining the `codeowners` file quickly.
    * Framework is modified, or non obvious usage of it -> get a review from [R&P team](https://dd.enterprise.slack.com/archives/C025TJ4RZ8X)

:rocket: Once your PR is reviewed, you can merge it!

🛟 [#apm-shared-testing](https://dd.enterprise.slack.com/archives/C025TJ4RZ8X) 🛟

## Reviewer checklist

* [ ] If PR title starts with `[<language>]`, double-check that only `<language>` is impacted by the change
* [ ] No system-tests internal is modified. Otherwise, I have the approval from [R&P team](https://dd.enterprise.slack.com/archives/C025TJ4RZ8X)
* [ ] CI is green, or failing jobs are not related to this change (and you are 100% sure about this statement)
* [ ] A docker base image is modified?
    * [ ] the relevant `build-XXX-image` label is present
* [ ] A scenario is added (or removed)?
    * [ ] Get a review from [R&P team](https://dd.enterprise.slack.com/archives/C025TJ4RZ8X)

PR #329, opened 2022-04-28
enable missing python test

vdeturckheim:

PR #3345, opened 2024-10-30
parametric: Adds a feature to track the parity for parametric endpoints

mabdinur:
## Motivation

Introduces `parametric_endpoint_parity` feature.

The goal of the parametric_endpoint_parity feature is to:
- Track which parametric endpoints are implemented in each language
- Document and assert the expected input and outputs for each parametric endpoint
- Future proofing: Catch unexpected regressions/modifications in parametric endpoints
- Track which parametric apps are incomplete and require additional work
- Require all future parametric endpoints to have one test case in `parametric_endpoint_parity.py` 

## Changes

- Adds unit tests/documentation for supported parametric endpoints
- Updates the python parametric app to ensure it is fully up to spec. This will be the reference implementation for all endpoints.
- Marks some parametric endpoints for removal
- Marks some arguments in parametric endpoints for removal
- Includes a link to the `parametric_endpoint_parity` tab in the feature parity dashboard in contributing docs. Before introducing new tests developers should be aware of the health and availability of parametric endpoints. 

## Next steps

- Fix all failing tests in `test_parametric_endpoints.py`
- Remove: 
   - `_TestSpan.span_get_name`, `_TestSpan.span_get_resource`, `_TestSpan.span_get_meta`, `_TestSpan.span_get_metric` interfaces/endpoints. These operations are not supported in most languages. Tests should be refactored to use the test agent to retrieve span data. 
   - _TestOtelSpan.get_attribute`, `_TestOtelSpan.get_name`, `_TestOtelSpan.get_links` interfaces and parametric endpoints. These operations are not supported in the opentelemetry-api. Tests should be refactored to use the test agent to retrieve span data.
   - `APMLibrary.current_span` and `APMLibrary.otel_current_span` interfaces/endpoints. Retrieving the current active span is not possible in all parametric apps. Many parametric apps are multi-threaded do not store the "current span" as the active span. Spans should always be retrieved by span_id.
   - `APMLibrary.http_client_request` endpoint. This endpoint only exists in 1 or 2 apps. It does not test a span/tracer api. The `make_distinct_call` weblog endpoint should be used instead. 
   -  http_headers, origin, and links arguments from `trace/span/start`. Most libraries do not support initializing a datadog span with these values. We will provide a alternative mechanisms to apply these values to spans: https://github.com/DataDog/system-tests/pull/3300

## Workflow

## Reviewer checklist


utils/parametric/spec/otel_trace.py
```diff
@@ -1,17 +1,7 @@
 from typing import TypedDict
+from enum import Enum
```
mabdinur, line 2: In test_parametric_endpoints, I document usages of the `SpanKind` and `StatusCode` enums. I think this is a better interface than asking developers to use a set of constants.

utils/parametric/spec/otel_trace.py
```diff
@@ -62,3 +52,61 @@ def convert_to_proto(attributes: dict) -> pb.Attributes:
         if ret:
             list_attr[k] = ret
     return pb.Attributes(key_vals=list_attr)
+
+
+class SpanKind(Enum):
```
cbeauchesne, line 57: This enum should be in `utils/dd_constants.py` ? 

It would be a good idea to also add a link to the RFC.

Same comment for enum above

utils/parametric/spec/otel_trace.py
```diff
@@ -62,3 +52,61 @@ def convert_to_proto(attributes: dict) -> pb.Attributes:
...
+    #: path latency relationship between producer and consumer spans.
+    PRODUCER = 4
+
+    #: Indicates that the span describes a consumer receiving a message from a
+    #: broker. Unlike client and server, there is usually no direct critical
+    #: path latency relationship between producer and consumer spans.
+    CONSUMER = 5
+
+
+# TODO: remove these constants and use the enum values directly
```
cbeauchesne, line 87: Todo to be adressed :)

utils/_features.py
```diff
@@ -2426,3 +2426,13 @@ def serverless_span_pointers(test_object):
         """
         pytest.mark.features(feature_id=328)(test_object)
         return test_object
+
+    @staticmethod
+    def parametric_endpoint_parity(test_object):
+        """
+        Enforces standardized behaviors for configurations across the tracing libraries.
```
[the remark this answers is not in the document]
cbeauchesne: https://feature-parity.us1.prod.dog/#/?feature=329 does not describe this, it's "#329: Source Code vulnerability - Stack Trace".

Ping me if you need any help to create a feature in FPD.

cbeauchesne (changes requested):

utils/parametric/spec/otel_trace.py
mabdinur: This is this a big refactor. I moved these changes to their own PR: https://github.com/DataDog/system-tests/pull/3370/files

utils/_features.py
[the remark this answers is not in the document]
cbeauchesne: is it resolved ?

tests/parametric/test_parametric_endpoints.py
```diff
@@ -0,0 +1,677 @@
...
+        - name: str
+        - service: Optional[str]
+        - resource: Optional[str]
+        - parent_id: Optional[Union[str,int]]
+        - type: Optional[str]
+        - tags: Optional[List[Tuple[str, str]]]
+
+        Supported Return Values:
+        - span_id: Union[int, str]
+        - trace_id: Union[int, str]
```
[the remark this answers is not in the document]
mabdinur: It's a bit of a mess. For example in javascript serializing 64bit integers results in percision issues. So in the nodejs app (and tracer) trace_id and span_id are usually stored as strings. Also how trace_ids are represented is it's own can of worms. Some tracers/parametric_apps store trace_ids as hex, others return only the lower 64bits of the trace_id, some always return strings while others always return integers. It's a mess. Fixing this issue will require updating most parametric endpoints and thats beyond the scope of the PR. I have a medium priority JIRA card that documents this issue. For now I think we should just document the current state. We currently work around this issue via the find_span, find_trace, and find_span_in_trace helpers that hide these inconsistencies.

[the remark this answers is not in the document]
mabdinur: Yup that's the plan but making this change here would bloat this PR. I will address this issue in a future task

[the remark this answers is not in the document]
bm1549: @mabdinur do we have a card tracking this? No need to post here, just wanted to make sure it's there!

utils/build/docker/python/parametric/apm_test_client/server.py
```diff
@@ -37,6 +38,10 @@
 
 spans: Dict[int, Span] = {}
 otel_spans: Dict[int, OtelSpan] = {}
+# Store the active span for each tracer in an array to allow for easy global access
+# FastAPI resets the contextvar containing the active span after each request
+active_ddspan = [None]
```
ZStriker19, line 43: Why a list and not just a direct assignment? Also could we just use `spans` and make sure we always append to it and then know the last span is always the active one? Or do we need to allow for something more nuanced?

utils/parametric/_library_client.py
```diff
@@ -530,6 +540,13 @@ def set_error(self, typestr: str = "", message: str = "", stack: str = ""):
     def add_link(self, parent_id: int, attributes: dict = None, http_headers: List[Tuple[str, str]] = None):
         self._client.span_add_link(self.span_id, parent_id, attributes, http_headers)
 
+    def finish(self):
+        self._client.finish_span(self.span_id)
+
+    ### Do not use the methods below in parametric tests, they will be removed in a future PR ####
```
ZStriker19, line 546: Much better and simpler pattern for verifying data.

utils/parametric/_library_client.py
```diff
@@ -941,41 +959,50 @@ def otel_start_span(
...
 
-    def get_tracer_config(self) -> Dict[str, Optional[str]]:
-        return self._client.get_tracer_config()
+    def otel_current_span(self) -> Union[_TestOtelSpan, None]:
+        resp = self._client.otel_current_span()
+        if resp is None:
+            return None
+        return _TestOtelSpan(self._client, resp["span_id"], resp["trace_id"])
+
+    def is_alive(self) -> bool:
```
ZStriker19, line 1003: How will this be used in future PRs?

tests/parametric/test_parametric_endpoints.py
```diff
@@ -0,0 +1,732 @@
+# TODO(mabdinur): Move this file to tests/test_the_test/
+"""
+This module provides one unit test for each parametric endpoint.
+The results of these unit tests are reported to the feature parity dashboard.
+Parametric endpoints that are not tested in this file are not yet supported.
+Avoid using these endpoints in the parametric tests.
```
ZStriker19, line 6:
```suggestion
Avoid using those endpoints in the parametric tests.
```
Should we list those? Or should we just delete this line knowing that we intend on adding tests for them. I feel like if I found an "unsupported" endpoint that I needed... I'd probably just use it anyways... or maybe would write the test for it.

tests/parametric/test_parametric_endpoints.py
```diff
@@ -0,0 +1,732 @@
...
+        - tags: Optional[List[Tuple[str, str]]]
+
+        Supported Return Values:
+        - span_id: Union[int, str]
+        - trace_id: Union[int, str]
+        """
+        with test_library:
+            with test_library.start_span("parent") as s1:
+                pass
+
```
ZStriker19, line 43:
```suggestion
# to test proper parenting behavior
```

tests/parametric/test_parametric_endpoints.py
```diff
@@ -0,0 +1,732 @@
...
+        Validates that /traces/span/finish API finishes a span and sends it to the agent.
+
+        Supported Parameters:
+        - span_id: Union[int, str]
+        Supported Return Values:
+        """
+        with test_library:
+            # Avoids calling start_span_generator.__exit__() since this method manually finishes the span.
+            start_span_generator = test_library.start_span("span")
+            s1 = start_span_generator.__enter__()
```
ZStriker19, line 81: why use `__enter__` rather than with?

tests/parametric/test_parametric_endpoints.py
ZStriker19, line 6: Maybe add a link to the docs showing how to startup the python server and view the endpoints?  https://github.com/DataDog/system-tests/blob/main/docs/scenarios/parametric.md#http

utils/_features.py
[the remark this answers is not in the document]
mabdinur: No it's not. My initial plan was to merge this PR with a unique ID and then update the FPD. It seems like I got it backwards. 

 Can I get your help updating the feature dashboard?

utils/build/docker/python/parametric/apm_test_client/server.py
mabdinur: We can assign the `active_ddspan` to object but we will need to use the `global` keyword in every endpoint/function that modifies this field. If the global keyword is not used then the function will modify it's own copy of the global variable and this will introduce failures that are hard to debug. By using a list here we make setting/getting the active_span dummy/munir proof

utils/parametric/_library_client.py
mabdinur: It will be used in crash tracking tests or any test that should kill the parametric app.

tests/parametric/test_parametric_endpoints.py
mabdinur: test_library.start_span.__exit__() calls span finish, we only want to execute the logic before the span object is yielded

tests/parametric/test_parametric_endpoints.py
mabdinur: YES!! I will add this

tests/parametric/test_parametric_endpoints.py
```diff
@@ -0,0 +1,732 @@
...
+
+@scenarios.parametric
+@features.parametric_endpoint_parity
+class Test_Parametric_DDSpan_Start:
+    def test_start_span(self, test_agent, test_library):
+        """
+        Validates the /trace/span/start API creates a new span:
+
+        Supported Parameters:
+        - span_id: int
```
mabdinur, line 28:
```suggestion
```

tests/parametric/test_parametric_endpoints.py
mabdinur: Nah we shouldn't list, it's meant to be a simple warning. I plan to link this file in our onboarding docs. We should clearly state that all new endpoints should have a unit test here.

utils/_features.py
```diff
@@ -2426,3 +2426,13 @@ def serverless_span_pointers(test_object):
...
+
+        https://feature-parity.us1.prod.dog/#/?feature=329
+        """
+        pytest.mark.features(feature_id=329)(test_object)
```
mabdinur, line 2437: TODO: Update this ID

manifests/golang.yml
```diff
@@ -468,6 +468,18 @@ tests/:
     test_otel_sdk_interoperability.py: missing_feature
     test_otel_span_with_baggage.py:
       Test_Otel_Span_With_Baggage: missing_feature
+    test_parametric_endpoints.py:
+      Test_Parametric_DDSpan_Add_Link: missing_feature (add_link endpoint is not implemented)
+      Test_Parametric_DDSpan_Set_Resource: missing_feature (set_resource endpoint is not implemented)
```
mtoffl01, line 473: Span links and resource can only be configured when starting a span. These fields can't be modified on an existing span.
So, the features aren't supported... hence why the endpoints aren't implemented

manifests/golang.yml
```diff
@@ -468,6 +468,18 @@ tests/:
...
+      Test_Parametric_DDTrace_Baggage: missing_feature (baggage is not supported)
```
mtoffl01, line 474: Baggage _is_ supported (you can add a baggage item to a span with `span.SetBaggageItem(k, v)`, but yes the parametric doesn't have an endpoint that allows you to set baggage.
Also, I'm not sure what you mean by "baggage is not supported" -- like, baggage is supported, but not up to the otel spec. So what does it mean for baggage to be supported?

tests/parametric/test_parametric_endpoints.py
```diff
@@ -0,0 +1,733 @@
...
+        test_library.crash()
+        assert not test_library.is_alive()
+
+
+@scenarios.parametric
+@features.parametric_endpoint_parity
+class Test_Parametric_DDTrace_Current_Span:
+    def test_current_span(self, test_agent, test_library):
+        """
+        Validates that /trace/span/current_span runs the current Datadog span.
```
mtoffl01, line 316:
```suggestion
        Validates that /trace/span/current_span returns the active Datadog span.
``` 
?

manifests/golang.yml
```diff
@@ -468,6 +468,18 @@ tests/:
...
+      Test_Parametric_DDTrace_Crash: missing_feature (crash endpoint is not implemented)
+      Test_Parametric_DDTrace_Current_Span: missing_feature (current span endpoint is not defined)
```
mtoffl01, line 476: `test_library.current_span()` seems to return the current active span on the tracer, but dd-trace-go doesn't have this concept, spans are stored in local context objects.
Do we still want to say the endpoint isn't implemented, or do we want to say this feature isn't supported / compatible?

tests/parametric/test_parametric_endpoints.py
```diff
@@ -0,0 +1,733 @@
...
+        span = find_only_span(traces)
+        assert span["error"] == 1
+
+
+@scenarios.parametric
+@features.parametric_endpoint_parity
+class Test_Parametric_OtelSpan_Set_Name:
+    def test_otelspan_set_name(self, test_agent, test_library):
+        """
+        Validates that /trace/otel/set_name sets a name on a span.
```
mtoffl01, line 576:
```suggestion
        Validates that /trace/otel/set_name sets the resource name on a span.
```

tests/parametric/test_parametric_endpoints.py
```diff
@@ -0,0 +1,733 @@
...
+        - attributes: Optional[Dict[str, str]]
+        - links: Optional[List[Link]]
+
+
+        Supported Return Values:
+
+        - span_id: Union[int, str]
+        - trace_id: Union[int, str]
+
+        Validates the /trace/otel/end_span API finishes a span and sends it to the agent.
```
mtoffl01, line 491: So this test validates that the start_span and end_span endpoints are supported, right?
Maybe this comment could be clearer, like:
Validates that the /trace/otel/start_span API creates a new span and that the /trace/otel/end_span API finishes a span and sends it to the agent.
<information about supported params and return values>

manifests/golang.yml
```diff
@@ -468,6 +468,18 @@ tests/:
...
+      # The set_name endpoint should set the resouce name (not the span name)
+      Test_Parametric_OtelSpan_Set_Name: bug (APMAPI-778)
+      # String attributes are incorrectly stored/serialized in a list
+      Test_Parametric_OtelSpan_Start_Finish: bug (APMAPI-795)
```
mtoffl01, line 480: They are? If you have more info, can we link to an actual dev card?

mtoffl01 (changes requested):
I'd suggest moving the comments about failing tests in the manifests to be inline with the test name. E.g, rather than:
```
      # The set_name endpoint should set the resouce name (not the span name)
      Test_Parametric_OtelSpan_Set_Name: bug (APMAPI-778)
```

We do:

```
      Test_Parametric_OtelSpan_Set_Name: bug (APMAPI-778) # The set_name endpoint should set the resouce name (not the span name)
```

With the multiline format I had some confusion around which comment is attached to which test entry.

tests/parametric/test_parametric_endpoints.py
```diff
@@ -0,0 +1,733 @@
...
+@scenarios.parametric
+@features.parametric_endpoint_parity
+class Test_Parametric_DDSpan_Add_Link:
+    def test_add_link(self, test_agent, test_library):
+        """
+        Validates that /trace/span/add_link adds a link to a span.
+
+        Supported Parameters:
+        - span_id: Union[int, str]
+        - parent_id: Union[int, str]
```
zacharycmontoya, line 214:
```suggestion
```

tests/parametric/test_parametric_endpoints.py
```diff
@@ -0,0 +1,733 @@
...
+
+@scenarios.parametric
+@features.parametric_endpoint_parity
+class Test_Parametric_DDTrace_Baggage:
+    def test_set_baggage(self, test_agent, test_library):
+        """
+        Validates that /trace/span/set_baggage sets a baggage item on a span.
+
+        Supported Parameters:
+        - span_id: Union[int, str]
```
zacharycmontoya, line 389: I'm not sure that it makes sense for the API to require a span_id to set baggage, since you can have baggage even without a corresponding span 🤔

tests/parametric/test_parametric_endpoints.py
```diff
@@ -0,0 +1,733 @@
...
+@features.parametric_endpoint_parity
+class Test_Parametric_OtelSpan_Start_Finish:
+    def test_span_start_and_finish(self, test_agent, test_library):
+        """
+        Validates the /trace/otel/start_span API creates a new span:
+
+        Supported Parameters:
+        - name: str
+        - timestamp (μs): Optional[int]
+        - span_kind: Optional[Literal[0, 1, 2, 3, 4, 5]] # integers correspoond -> ["UNSPECIFIED", "INTERNAL", "SERVER", "CLIENT", "PRODUCER", "CONSUMER"]
```
zacharycmontoya, line 480: Instead of passing these as ints, can we instead pass them as strings? Also, UNSPECIFIED should be renamed to INTERNAL based on the OpenTelemetry API: https://opentelemetry.io/docs/specs/otel/trace/api/#spankind

tests/parametric/test_parametric_endpoints.py
```diff
@@ -0,0 +1,733 @@
...
+        assert links[0]["span_id"] == int(s2.span_id)
+        assert links[0]["attributes"]["link.key"] == "value"
+
+
+@scenarios.parametric
+@features.parametric_endpoint_parity
+class Test_Parametric_OtelSpan_Set_Attribute:
+    def test_otel_set_attribute(self, test_agent, test_library):
+        """
+        Validates that /traces/otel/end_span API finishes a span and sends it to the agent.
```
zacharycmontoya, line 531:
```suggestion
        Validates that /traces/otel/set_attributes API sets the corresponding attributes on a span.
```

tests/parametric/test_parametric_endpoints.py
```diff
@@ -0,0 +1,733 @@
...
+
+        Supported Parameters:
+        - span_id: Union[int, str]
+        - key: str
+        Supported Return Values:
+        """
+        with test_library:
+            with test_library.otel_start_span("otel_get_attribute") as s1:
```
zacharycmontoya, line 539:
```suggestion
            with test_library.otel_start_span("otel_set_attribute") as s1:
```

tests/parametric/test_parametric_endpoints.py
```diff
@@ -0,0 +1,733 @@
...
+            with test_library.otel_start_span("otel_record_exception") as s1:
+                s1.record_exception("MyException Parametric tests rock", {"error.key": "value"})
+                s1.end_span()
+
+        traces = test_agent.wait_for_num_traces(1)
+        span = find_only_span(traces)
+        assert "events" in span["meta"]
+        events = json.loads(span["meta"]["events"])
+        assert len(events) == 1
+        assert events[0]["name"].lower() in ["exception", "error"]
```
zacharycmontoya, line 642: Can you also assert that the attribute is set correctly?

utils/parametric/_library_client.py
zacharycmontoya: So I know that the .NET parametric app crashes correctly, which leads me to think that this may be flaky. It looks like the `self._client.wait` command calls `container.reload()` which seems wrong. Can you instead do a check on the container object instead of calling wait?

manifests/dotnet.yml
```diff
@@ -351,6 +351,25 @@ tests/:
...
+      Test_Parametric_DDSpan_Add_Link: missing_feature (add_link parametric endpoint is not implemented)
+      # Set error endpoint does not set the error.type tag on spans
+      Test_Parametric_DDSpan_Set_Error: bug (APMAPI-778)
+      Test_Parametric_DDSpan_Set_Resource: missing_feature (set_resource parametric endpoint is not implemented)
+      Test_Parametric_DDTrace_Baggage: missing_feature (baggage is not supported)
+      # Crash endpoint does not crash the application, it only raises an exception in a thread
+      Test_Parametric_DDTrace_Crash: bug (APMAPI-778)
+      Test_Parametric_DDTrace_Current_Span: missing_feature (current_span parametric endpoint is not implemented)
+      # Tests below are failing due to the start_span endpoint consistently returning 500s, the root cause is unknown
+      Test_Parametric_OtelSpan_Events:  bug (APMAPI-778)
```
zacharycmontoya, line 364: We need to set `DD_TRACE_OTEL_ENABLED=true` in order to run these test. Can you add that to the library env for the Otel tests? If instead we should enable `DD_TRACE_OTEL_ENABLED=true` by default for the .NET parametric test app, then it fails the test `tests/parametric/test_otel_env_vars.py::Test_Otel_Env_Vars::test_otel_sdk_disabled_set`

tests/parametric/test_parametric_endpoints.py
mabdinur: They only operate on spans. I am just providing documentation/tests for existing behavior. We can update this and refactor the existing tests in a future PR.

cc: @rachelyangdog

tests/parametric/test_parametric_endpoints.py
mabdinur: Actually my plan is to pass in an Enum ([PR](https://github.com/DataDog/system-tests/pull/3370/files#diff-f5f86992245595c4ea7be47e8ce6502becdc5166459538487b189b106ff00093)). This change involves updating multiple lines so I am planning to make it in a future PR. For now I am just documenting how the endpoints work.

manifests/golang.yml
mabdinur: Ya we can after we confirm whether this is a bug with the golang tracer or the parametric app

tests/parametric/test_parametric_endpoints.py
mabdinur: These arguments allow us to do the the following operation in a parametric endpoint:
SpanA.add_link(SpanB, attributes)` where `SpanA.span_id==span_id and SpanB.span_id==parent_id`.

We should honestly rename `parent_id` but this is the spec we currently have 😭 . We can fix it in a follow up PR.

manifests/golang.yml
mabdinur: Ahh that makes sense. Yes I meant doesn't support baggage according to the RFC that Lucas wrote. I will update this comment to be more descriptive.

tests/parametric/test_parametric_endpoints.py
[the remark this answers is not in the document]
mabdinur: Yup, I created it 2 weeks ago.

tests/parametric/test_parametric_endpoints.py
mabdinur: Yup, I am a dummy

manifests/golang.yml
mabdinur: Ya that's a good point. We should say this feature isn't supported. This way someone doesn't go in and try to add the endpoint in the future.

manifests/dotnet.yml
mabdinur: Yup we have the same issue in python. I would like to avoid setting this environment variable for every test test uses the opentelemetry api. It would be nice to find a way to just set in one place but still allow tests override this env. I can make this change.

tests/parametric/test_parametric_endpoints.py
mabdinur: Great call out! Adding this assertion surfaced an issue in the js parametric app: https://github.com/DataDog/system-tests/pull/3345/commits/5be26e7ca0c551ad782092693df0f15ec9931c2f.

The record exception endpoint ignores the attributes parameter: https://github.com/DataDog/system-tests/blob/278f73ae5111a4ec4f1313e66250d86c073714a2/utils/build/docker/nodejs/parametric/server.js#L323.

utils/parametric/_library_client.py
```diff
@@ -42,13 +42,13 @@ def __init__(self, url: str, timeout: int, container: Container):
...
-        self._wait(timeout)
+        self.wait(timeout, "/non-existent-endpoint-to-ping-until-the-server-starts")
 
-    def _wait(self, timeout):
+    def wait(self, timeout, url):
         delay = 0.01
         for _ in range(int(timeout / delay)):
             try:
-                resp = self._session.get(self._url("/non-existent-endpoint-to-ping-until-the-server-starts"))
+                resp = self._session.get(self._url(url))
```
mabdinur, line 51:
```suggestion
                resp = self._session.get(self._url("/non-existent-endpoint-to-ping-until-the-server-starts"))
```

utils/parametric/_library_client.py
```diff
@@ -63,7 +63,7 @@ def _wait(self, timeout):
         else:
             self._print_logs()
             message = f"Timeout of {timeout} seconds exceeded waiting for HTTP server to start. Please check logs."
-            _fail(message)
+            raise RuntimeError(message)
```
mabdinur, line 66:
```suggestion
            _fail(message)
```

manifests/golang.yml
mtoffl01: > In python we use a web framework that stores spans in the context of a request. So a span is only active/current in the context of the request it was initialized in.
Yes, Go is the same.

> Is this a flow we can add/mock in the go parametric app.
No, not with the current functionality.

How do you think we should label this test then? 🤔

utils/parametric/_library_client.py
mabdinur: Ah that makes sense. The python app fails to reload. I will clean this up

utils/parametric/_library_client.py
```diff
@@ -42,13 +42,13 @@ def __init__(self, url: str, timeout: int, container: Container):
         self.container = container
 
         # wait for server to start
-        self._wait(timeout)
+        self.wait(timeout, "/non-existent-endpoint-to-ping-until-the-server-starts")
 
-    def _wait(self, timeout):
+    def wait(self, timeout, url):
```
mabdinur, line 47:
```suggestion
    def wait(self, timeout):
```

utils/parametric/_library_client.py
```diff
@@ -527,41 +543,50 @@ def otel_start_span(
...
-        return self._client.get_tracer_config()
+    def otel_current_span(self) -> Union[_TestOtelSpan, None]:
+        resp = self._client.otel_current_span()
+        if resp is None:
+            return None
+        return _TestOtelSpan(self._client, resp["span_id"], resp["trace_id"])
+
+    def is_alive(self) -> bool:
+        try:
+            self._client.wait(0.03, "/non-existent-endpoint")
```
mabdinur, line 589:
```suggestion
            self._client.wait(0.03)
```

ZStriker19 (approved):

manifests/golang.yml
```diff
@@ -468,6 +468,16 @@ tests/:
     test_otel_sdk_interoperability.py: missing_feature
     test_otel_span_with_baggage.py:
       Test_Otel_Span_With_Baggage: missing_feature
+    test_parametric_endpoints.py:
+      Test_Parametric_DDSpan_Add_Link: missing_feature (add_link endpoint is not implemented)
```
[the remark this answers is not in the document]
mabdinur: Yup, v2 should be released soon 🤞 and then we should be able to support this endpoint. We can even add the endpoint now and set the test to`Test_Parametric_DDSpan_Add_Link: v2.0.0`.

utils/parametric/_library_client.py
```diff
@@ -42,17 +42,18 @@ def __init__(self, url: str, timeout: int, container: Container):
...
+    def wait(self, timeout, reload):
         delay = 0.01
         for _ in range(int(timeout / delay)):
             try:
                 resp = self._session.get(self._url("/non-existent-endpoint-to-ping-until-the-server-starts"))
                 if resp.status_code == 404:
                     break
             except Exception:
-                self.container.reload()
+                if reload:
```
cbeauchesne, line 55: For which reason you may not want to reload?

tests/parametric/test_parametric_endpoints.py
```diff
@@ -0,0 +1,739 @@
+# TODO(mabdinur): Move this file to tests/test_the_test/
```
cbeauchesne, line 1: TODO to be addressed :)

cbeauchesne: I'm actually wondering if really we want to have this in test the test folder ?

utils/parametric/_library_client.py
mabdinur: In APMLibrary.is_alive() we want to check whether the parametric app is running. This is useful when validating the behavior of the trace/crash endpoint. I can also see a world where we want to use this method in more crash tracking tests or as a quick health check.

tests/parametric/test_parametric_endpoints.py
mabdinur: I will remove this comment. This PR is already doing a lot. We can discuss the merits of this change in a future PR.

mabdinur, line 1:
```suggestion
```

utils/parametric/_library_client.py
cbeauchesne: But if the container is stopped, there is no chance that the app is running, so you'll wait for nothing no? Or do I miss something ?

cbeauchesne: I think `_wait` may not be a good fit for `is_alive`, as it's prupose is to be a healtcheck, so you'll had to handle some extra stuff. May a suggest a dedicated function that : 

1. check the container status
2. check that the endpoint respond 404
3. returns True if 1. and 2. are successful

mabdinur: Done: 095a58ab2aa6eac7b4017f1d66a51a162e04622d

cbeauchesne (approved):

PR #3353, opened 2024-10-31
Avoid false positive, for tracer without appsec/iast

jandro996:
## Motivation

We are having easy wins in FP for languages that doesn't have IAST implemented

To avoid this false positives, we need to check first that we detect attacks(appsec)/the insecure endpoint is vulnerable(iast) that will be done by calling a test that raises an ASM event.

## Changes

Add a call to a test that raises an asm event before testing the ones that are not working with asm events and only test the propagation

## Workflow

## Reviewer checklist


uurien (approved):

tests/appsec/test_asm_standalone.py
```diff
@@ -56,6 +59,40 @@ def _assert_tags_value(span, obj, expected_tags):
         except (KeyError, AssertionError) as e:
             return False
 
+    def setup_check_product_is_enabled(func):
+        def wrapper(self):
+            headers = {}
+            if self.tested_product is "appsec":
```
cbeauchesne, line 65: string comparison uses '==' : 

```suggestion
            if self.tested_product == "appsec":
```

tests/appsec/test_asm_standalone.py
```diff
@@ -56,6 +59,40 @@ def _assert_tags_value(span, obj, expected_tags):
...
+                headers = {
+                    "User-Agent": "Arachni/v1",  # attack if APPSEC enabled
+                }
+            self.check_r = weblog.get(self.requestdownstreamUrl, headers=headers,)
+            func(self)
+
+        return wrapper
+
+    def check_product_is_enabled(func):
```
cbeauchesne, line 74: I'm afraid that the usage of a decorator will be misleading for non python user. As this does not really decorate the test method, but just add a common assertion, don't you think that it will be more easier to understand if you name it `assert_product_is_enabled`, and call it directly in test methods with `assert_product_is_enabled(self.check_r, self.tested_product` ?

tests/appsec/test_asm_standalone.py
```diff
@@ -56,6 +59,40 @@ def _assert_tags_value(span, obj, expected_tags):
         except (KeyError, AssertionError) as e:
             return False
 
+    def setup_check_product_is_enabled(func):
```
cbeauchesne, line 62: Same comment as below, I think it'll be easier to understand for non python user if we use a simple method called directly in `setup` bodies.

cbeauchesne (changes requested):

tests/appsec/test_asm_standalone.py
jandro996: I've changed it, thanks for the suggestion!

cbeauchesne (approved):

PR #3373, opened 2024-11-05
Khanayan123/enable config consistency test php

khanayan123:
## Motivation

Enable config consistency tests for PHP

## Changes

Enables configuration consistency tests for PHP and updates existing tests to validate PHP functionality in alignment with these configurations.

### DIFF
--- a/tests/test_config_consistency.py
+++ b/tests/test_config_consistency.py
@@ -241,7 +241,7 @@ class Test_Config_ClientIPHeader_Precedence:
         ("x-real-ip", "8.7.6.5"),
         ("true-client-ip", "5.6.7.2"),
         ("x-client-ip", "5.6.7.3"),
-        ("x-forwarded", "for=5.6.7.4"), # Note: APPSEC is currently discussing the correct values for this header
+        ("x-forwarded", "for=5.6.7.4"),  # Note: APPSEC is currently discussing the correct values for this header
         ("forwarded-for", "5.6.7.5"),
         ("x-cluster-client-ip", "5.6.7.6"),
         ("fastly-client-ip", "5.6.7.7"),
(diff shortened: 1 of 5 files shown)


utils/build/docker/php/parametric/server.php
```diff
@@ -150,7 +150,8 @@ function remappedSpanKind($spanKind) {
         \DDTrace\set_distributed_tracing_context($context["trace_id"], $context["distributed_tracing_parent_id"] ?? 0, $origin);
     }
     $span->name = arg($req, 'name');
-    $span->service = arg($req, 'service');
+    $service = arg($req, 'service');
+    $span->service = (!is_null($service) && $service !== '') ? $service : $span->service;
```
bwoebi, line 154:
```suggestion
    $span->service = !is_null($service) && $service !== '' ? $service : $span->service;
```Weird parenthesis.

**Review by link04 — APPROVED** [review]
LGTM

File: tests/test_config_consistency.py [hunk 2]
```diff
@@ -346,15 +364,25 @@ def test_kafka_integration_enabled_false(self):
 class Test_Config_IntegrationEnabled_True:
     """ Verify behavior of integrations automatic spans """
 
-    def setup_kafka_integration_enabled_true(self):
+    def setup_integration_enabled_true(self):
```
mabdinur, line 367: nit: Can we have two tests here? One for the dbm endpoint and one for the kafka endpoint. 

We can mark the kafka endpoint as irrelevant for php and mark the new dbm test as irrelevant for not kafka. I think this would make this test easier to read. 

This isn't a blocker I am also fine with the current changes

khanayan123: I guess depends on the purpose of the test, if we're testing to see if integrations are enabled/disabled then i'm in favor of keeping as is. If we're testing to see if kafka or pdo specifically is enabled or disabled them i'm in favor of splitting it

PR #3398, opened 2024-11-07
Revert "[python] Skip bugs for 2.16.0 release"

P403n1x87:
Reverts DataDog/system-tests#3385


cbeauchesne (commented):
Actually, those test will fail forever for version 2.16, so we need to keep, #3385 can't be reverted. I think the PR you want to amend is #3354

PR #3467, opened 2024-11-14
DEBUG-2334 fix running documentation with scenario

p-datadog:
## Motivation
Current documentation mixes upper and lower case for scenario name in the instructions for running a single file with a scenario.

## Changes
Change the instructions to use +S option to pass the actual scenario name as specified in the test file (in lower case). Now the instructions can be followed as written.

## Workflow

## Reviewer checklist

### DIFF
--- a/docs/execute/run.md
+++ b/docs/execute/run.md
@@ -15,10 +15,10 @@
 If the test contains `@scenarios.SCENARIO_NAME` such as `@scenarios.integrations`, then the `./run.sh` needs to be adjusted to the following:
 
 ```bash
-./run.sh SCENARIO_NAME tests/path_to_test.py
+./run.sh +S SCENARIO_NAME tests/path_to_test.py
 
 # Example: for @scenarios.integrations in tests/integrations/test_sql.py
-./run.sh INTEGRATIONS tests/integrations/test_sql.py
+./run.sh +S integrations tests/integrations/test_sql.py
 ```
 
 ## Run only one class, or one method


docs/execute/run.md
```diff
@@ -15,10 +15,10 @@
 If the test contains `@scenarios.SCENARIO_NAME` such as `@scenarios.integrations`, then the `./run.sh` needs to be adjusted to the following:
 
 ```bash
-./run.sh SCENARIO_NAME tests/path_to_test.py
+./run.sh +S SCENARIO_NAME tests/path_to_test.py
 
 # Example: for @scenarios.integrations in tests/integrations/test_sql.py
-./run.sh INTEGRATIONS tests/integrations/test_sql.py
+./run.sh +S integrations tests/integrations/test_sql.py
```
cbeauchesne, line 21: Scenario names are in upper case in `bash` commands, and `lowecase` in python.

Furthermore, the first argument is interpreted as the senario if it's in uppercase (that the prupose to keep it uppercase), in order to avoid to write the `+S`.

p-datadog: I get that the usage of uppercase might make some operations more convenient for those developers that use system tests all the time. However, since it is not possible to know which case to use when (you need to somehow have this knowledge already), using two different cases hampers discoverability and creates problems when searching (since both cases need to be checked).

The instructions, as written, say that scenario name is the value in the decorator in the test code. They do not mention the required transformation to upper case in order to use the scenario name in (some? all?) shell commands.

This PR proposes using the same case everywhere. This is already supported by the code and is easy to understand for newcomers to the project - the target audience, one would assume, for the instructions in question.

Alternatively, we could add the text of your comment to the instructions, but 1) the instructions would get much longer, 2) they would also take more time to understand, 3) it would still be impossible to copy-paste scenario name from a test source file to the terminal and have the test be correctly executed.

PR #3474, opened 2024-11-15
baggage tests max items update

rachelyangdog:
## Motivation

Fixing baggage max items implementation based off the change in the RFC. "If baggage data exceeds one or both of these limits, APM SDKs should drop baggage name/value pairs until both conditions are met. For example, if baggage contains 70 name/value pairs, the SDK should only add 64 of them to the baggage header and drop the other 6. The W3C leaves the process of selecting which pairs to keep or drop up to implementers. The simplest algorithm in this example is to keep the first 64 pairs and drop the last 6."
## Changes
updated the baggage max items test accordingly

## Workflow

## Reviewer checklist


tests/parametric/test_headers_baggage.py
```diff
@@ -218,7 +218,16 @@ def test_baggageheader_maxitems_inject_D016(self, test_library):
...
+                if key == "baggage":
+                    baggage_header = value
+                    break
+
+            # baggage should be injected
+            assert baggage_header is not None
+            items = baggage_header.split(",")
+            assert len(items) == max_items
 
     def test_baggageheader_maxbytes_inject_D017(self, test_library):
```
ida613, line 232: we also need to fix this one

robertomonteromiguel (approved):

tests/parametric/test_headers_baggage.py
```diff
@@ -214,17 +214,32 @@ def test_baggageheader_maxitems_inject_D016(self, test_library):
...
+                span.set_baggage(key, value)
 
-        headers = test_library.inject_headers(span.span_id)
-        assert not any("baggage" in item for item in headers)
+            headers = test_library.inject_headers(span.span_id)
+            for header in headers:
+                if "baggage" in header:
+                    baggage_header = header
+            header_size = len(baggage_header[1].encode("utf-8"))
+            assert header_size <= max_bytes
```
lucaspimentel, line 245: Should we also assert that there are 2 items? Based on the test for max items, I think the code would be something like this:
```
items = baggage_header[1].split(",")
assert len(items) == 2
```

I think it would be only 2 items because `key1`, `key2,` and `key3` would all be about 2,735 bytes long, so `key3` would exceed the limit.
```
len(key) = 4
len("=") = 1
len(value) = 8192 // 3 = 2730
len(',') = 1

4 + 1 + 2730 + 1 = 2735
```

lucaspimentel: Interestingly, if we dropped `key3` first (or skip it) then `key4` would fit. But the W3C spec doesn't dictate in which order to drop tags. If we drop them in order starting from the last one:
first, drop `key4` -> still too big
then drop `key3` -> now it what's left does fit

cbeauchesne:
FYI, You can update your branch to fix failure in CI.

tests/parametric/test_headers_baggage.py
rachelyangdog: I think we should make it clear in the RFC that we should include items in order until we reach the limit and drop everything else after that so that the tracers are more aligned. Also, the system-tests can be more specific and there would be less ambiguity.

lucaspimentel: I'll update the RFC, thanks.

cbeauchesne (approved):
Failure is not related (#3520)

PR #3516, opened 2024-11-20
DEBUG-2334 Ruby DI system tests

p-datadog:
## Motivation

Initial implementation of Dynamic Instrumentation in Ruby

## Changes

- Adds DebuggerController to rails70 weblog implementing log_probe, mix_probe and pii methods
- Adds Ruby to the test code that substitutes actual controller names into placeholders
- Marks tests that initial Ruby DI implementation does not implement as @missing_feature
- Adds pii_line test, which is a modification of the pii test using a line probe since Ruby does not currently implement local variable capture for method probes (this requires using method entry/exit trace points). Alters the PII verification code to read snapshot from line or method location based on the test being executed

In order to run these tests, the changes in https://github.com/DataDog/dd-trace-rb/pull/4098 are required which are pending debugging of failing non-DI tests.

Since DI is still not available in any released version of Ruby library, none of the manifests are changed in this PR.

## Workflow

1. ⚠️ Create your PR as draft ⚠️
2. Work on you PR until the CI passes (if something not related to your task is failing, you can ignore it)
3. Mark it as ready for review
    * Test logic is modified? -> Get a review from RFC owner. We're working on refining the `codeowners` file quickly.
    * Framework is modified, or non obvious usage of it -> get a review from [R&P team](https://dd.enterprise.slack.com/archives/C025TJ4RZ8X)

:rocket: Once your PR is reviewed, you can merge it!

🛟 [#apm-shared-testing](https://dd.enterprise.slack.com/archives/C025TJ4RZ8X) 🛟

## Reviewer checklist

* [ ] If PR title starts with `[<language>]`, double-check that only `<language>` is impacted by the change
* [ ] No system-tests internal is modified. Otherwise, I have the approval from [R&P team](https://dd.enterprise.slack.com/archives/C025TJ4RZ8X)
* [ ] CI is green, or failing jobs are not related to this change (and you are 100% sure about this statement)
* [ ] A docker base image is modified?
    * [ ] the relevant `build-XXX-image` label is present
* [ ] A scenario is added (or removed)?
    * [ ] Get a review from [R&P team](https://dd.enterprise.slack.com/archives/C025TJ4RZ8X)

PR #3531, opened 2024-11-22
DEBUG-3180 enable Ruby dynamic instrumentation tests

p-datadog:
## Motivation

Initial implementation of dynamic instrumentation in Ruby.

## Changes
https://github.com/DataDog/system-tests/pull/3516 added the needed controller implementation for dynamic instrumentation system tests and adjusted the tests to work with the feature as implemented in Ruby. However, that PR did not turn on the tests for any tracer version.

Now that https://github.com/DataDog/dd-trace-rb/pull/4098 is merged, DI is usable in Ruby once the environment variable to enable it is set, and the feature will be shipped with the next tracer version (2.8.0).

This PR enables the dynamic instrumentation tests for Ruby tracer 2.8.0 or higher.

The companion PR to this one is https://github.com/DataDog/dd-trace-rb/pull/4146 which adds the scenarios to the list of scenarios to be executed in the Ruby tracer CI.

This PR also changes decorators for Ruby from missing_feature to irrelevant. missing_feature decorator runs the setup (and the tests and expects them to fail) which causes unsupported probes to be sent to the tracer via remote config. The tracer logs that the metric/span/span decoration probe specifications are not valid/understood (while running the log probe tests, which are supported) as an error and the system tests fail when they assert that application log does not contain any errors. To avoid logging the errors, we need to skip the setup thus mark the tests as irrelevant.

## Workflow

## Reviewer checklist

### DIFF
--- a/tests/debugger/utils.py
+++ b/tests/debugger/utils.py
@@ -70,6 +70,8 @@ class _Base_Debugger_Test:
 
     rc_state = None
     weblog_responses = []
+    
+    setup_failures = []
 
     def initialize_weblog_remote_config(self):
         if self.get_tracer()["language"] == "ruby":
@@ -78,7 +80,10 @@ class _Base_Debugger_Test:
             # Therefore, we need to issue a request to the application for
             # remote config to start.
             response = weblog.get("/debugger/init")
-            assert response.status_code == 200
+            if response.status_code != 200:
+                # This should fail the test immediately but the failure is
+                # reported after all of the setup and the test are attempted
+                self.setup_failures.append("Failed to get /debugger/init: expected status code: 200, actual status code: %d" % (response.status_code))
 
     ###### set #####
     def set_probes(self, probes):
@@ -404,6 +409,10 @@ class _Base_Debugger_Test:
             }
 
         return _Base_Debugger_Test.tracer
+
+    def assert_setup_ok(self):
+        if self.setup_failures:
+            assert "\n".join(self.setup_failures) is None
 
     ###### assert #####
     def assert_rc_state_not_error(self):
--- a/tests/debugger/utils.py
+++ b/tests/debugger/utils.py
@@ -70,7 +70,7 @@ class _Base_Debugger_Test:
 
     rc_state = None
     weblog_responses = []
-    
+
     setup_failures = []
 
     def initialize_weblog_remote_config(self):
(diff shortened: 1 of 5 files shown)


p-datadog:
Ruby CI has been repaired. PHP CI is failing on all PRs.

cbeauchesne:
Follow-up after a chat with @p-datadog : 

Test on ruby are not passing because there is a test that check that no error is reported : some tests are marked as xfailed because the feature is not implemented. But as they are executed in the background, they cause some errors that make the no-error assertion fails.

If the test must be adapted, then you can move forward by using : 

```python
@missing_feature(context.library == "ruby", force_skip=True). # test must be adapted to not fail when an error is reported
```

**Or**, if error should really not be reported, then it means that the implementation must be adapted.

@shurivich could you help us to know if @p-datadog implementation is fine and the test should be adapted, or if really no error should be reported, even if some unsupported part are sent threw RC ?

tests/debugger/utils.py
```diff
@@ -71,6 +71,15 @@ class _Base_Debugger_Test:
...
     weblog_responses = []
 
+    def initialize_weblog_remote_config(self):
+        if self.get_tracer()["language"] == "ruby":
+            # Ruby tracer initializes remote configuration client from
+            # middleware that is only invoked during request processing.
+            # Therefore, we need to issue a request to the application for
+            # remote config to start.
+            response = weblog.get("/debugger/init")
+            assert response.status_code == 200
```
cbeauchesne, line 81: Sorry, I also miss that point. `setup` function must never fail. If you want to add any assertion, it must be done inside test functions.

```
def setup_stuff:
    self.init_response = weblog.get("/debugger/init")

def test_stuff:
    assert self.init_response.status_code == 200
```

cbeauchesne (changes requested):

tests/debugger/utils.py
p-datadog: Why must `setup` never fail?

cbeauchesne: "Test functions" are instrumented by the framework to offer a friendly output on failure (logs, some variables values on `assert`. 

`setup` functions are executed during the lifespan of the tested infra, but they are not "test functions". If they fails, you'll have a basic stack trace painful to understand

shurivich:
> @shurivich could you help us to know if @p-datadog implementation is fine and the test should be adapted, or if really no error should be reported, even if some unsupported part are sent threw RC ?

We had a bug in our setup.  We fixed it, and switched to the `missing_feature` here d526f8e097105e1a3f86d4a6fb494b2d5f3444e3

marcotc (approved):
Approval for the Ruby changes.

cbeauchesne (approved):

PR #3581, opened 2024-11-28
[Debugger] Ensure all probes have a version property

watson:
## Motivation

All probes sent via Remote Config should have a root `version` property. For some reason, one was not present in our mocked probe configs located in `tests/debugger/probes`. I'm not sure why this hasn't resulted in system test errors previously since the diagnostics schema expects that a version is reported. However, without this, at least the Node.js tracer, would set `version: undefined` in the object before it was serialized to JSON, which would completely remove the `version` property from the JSON output.

Maybe they other tracers fall back to `0` if there's no `version` property in the probe config?

## Changes

Hardcode all mocked probe configs to contain `version: 0`.

## Workflow

## Reviewer checklist


robertomonteromiguel (approved):
From the system-tests core point of view it is correct. But you would need the approval of other teams.

p-datadog:
This PR seems reasonable since it only adds a field to the input (which I believe Ruby implementation will ignore), but I will test locally before adding my checkmark.

shurivich (commented):
We have a method of enriching all probes with required information. 
https://github.com/DataDog/system-tests/blob/02c78efc9fe629d70812a996d87ff07ab256cec4/tests/debugger/utils.py#L76

I suggest to add the code:
               `probe["version"] = 0`

after the line 

https://github.com/DataDog/system-tests/blob/02c78efc9fe629d70812a996d87ff07ab256cec4/tests/debugger/utils.py#L92

instead of modifying all json probe files.

watson:
What's the reason we don't want to update the json at the source?

shurivich:
> What's the reason we don't want to update the json at the source?
Single responsibility - updates are handled in one place, ensuring clarity. 
Less changes.

But maybe it's just a matter of code taste, so it's up to you to decide.
Wanted to make sure that you're aware of this option, too.

watson:
@shurivich I think it makes a lot of sense to have a way to dynamically update the JSON to suite the language being instrumented and other non-fixed properties - exactly like you've done in `tests/debugger/utils.py` - otherwise we'd end up with a lot of duplication 👍 

However, the `version` property is currently static - just like the `id` property - and doesn't depend on the tracer (this isn't the version of the tracer, but the "probe version" and is tightly connected to the `id`).

So if you don't mind, I'll just leave it as is - at least for now. I think that makes it easier to understand the schema when just viewing the JSON file. Anyways, we can always refactor it if we want to make to the probe version dynamic ☺️

cbeauchesne:
Sorry, the script that computes which scenario to run failed, so this PR hasn't be tested 😢 

I've reverted it, and fixed the script.

To correct this PR, you must very probably update the hash, as the content you've sent has changed, I can help you to do so.

PR #3647, opened 2024-12-09
remove duplicate code in decorators

mabdinur:
## Workflow

## Reviewer checklist

### DIFF
--- a/tests/parametric/test_span_events.py
+++ b/tests/parametric/test_span_events.py
@@ -9,7 +9,7 @@ from utils.parametric.spec.trace import find_span, find_trace
 @rfc("https://docs.google.com/document/d/1cVod_VI7Yruq8U9dfMRFJd7npDu-uBpste2IB04GyaQ")
 @scenarios.parametric
 @features.span_events
-@missing_feature(reason="Agent does not advertise native span events serialization support yet")
+@missing_feature(True, reason="Agent does not advertise native span events serialization support yet")
 class Test_Span_Events:
     def _test_span_with_event(self, _library_env, test_agent, test_library, retrieve_events):
         """Test adding a span event, with attributes, to an active span.
--- a/utils/_decorators.py
+++ b/utils/_decorators.py
@@ -1,6 +1,7 @@
 import inspect
 import os
 import re
+from functools import partial
 
 import pytest
 import semantic_version as semver
@@ -49,25 +50,12 @@ def _ensure_jira_ticket_as_reason(item, reason: str):
         pytest.exit(f"Please set a jira ticket for {nodeid}, instead of reason: {reason}", 1)
 
 
-def _get_skipped_item(item, skip_reason):
+def _get_expected_failure_item(item, skip_reason, noxfail):
     if inspect.isfunction(item) or inspect.isclass(item):
         if not hasattr(item, "pytestmark"):
             setattr(item, "pytestmark", [])
 
-        item.pytestmark.append(pytest.mark.skip(reason=skip_reason))
-
-    else:
-        raise ValueError(f"Unexpected skipped object: {item}")
-
-    return item
-
-
-def _get_expected_failure_item(item, skip_reason, force_skip: bool = False):
-    if inspect.isfunction(item) or inspect.isclass(item):
-        if not hasattr(item, "pytestmark"):
-            setattr(item, "pytestmark", [])
-
-        if force_skip:
+        if noxfail:
             item.pytestmark.append(pytest.mark.skip(reason=skip_reason))
         else:
             item.pytestmark.append(pytest.mark.xfail(reason=skip_reason))
@@ -78,7 +66,7 @@ def _get_expected_failure_item(item, skip_reason, force_skip: bool = False):
 
 
 def _should_skip(condition=None, library=None, weblog_variant=None):
-    if condition is not None and not condition:
+    if condition is False:
         return False
 
     if weblog_variant is not None and weblog_variant != context.weblog_variant:
@@ -106,43 +94,30 @@ def _should_skip(condition=None, library=None, weblog_variant=None):
     return True
 
 
-def missing_feature(condition: bool = None, library=None, weblog_variant=None, reason=None, force_skip: bool = False):
+def decorator(skip, noxfail, condition, decorator_type, reason, callback, function_or_class):
+    if inspect.isclass(function_or_class):
+        assert condition is not None, _MANIFEST_ERROR_MESSAGE
+
+    if decorator_type in ("bug", "flaky"):
+        _ensure_jira_ticket_as_reason(function_or_class, reason)
+
+    full_reason = decorator_type if reason is None else f"{decorator_type} ({reason})"
+    if not skip:
+        return function_or_class
+    return callback(function_or_class, full_reason, noxfail)
+
+
+def missing_feature(condition=None, library=None, weblog_variant=None, reason=None, force_skip: bool = False):
     """decorator, allow to mark a test function/class as missing"""
-
     skip = _should_skip(library=library, weblog_variant=weblog_variant, condition=condition)
-
-    def decorator(function_or_class):
-
-        if inspect.isclass(function_or_class):
-            assert condition is not None or (library is None and weblog_variant is None), _MANIFEST_ERROR_MESSAGE
-
-        if not skip:
-            return function_or_class
-
-        full_reason = "missing_feature" if reason is None else f"missing_feature ({reason})"
-
-        return _get_expected_failure_item(function_or_class, full_reason, force_skip=force_skip)
-
-    return decorator
+    return partial(decorator, skip, force_skip, condition, "missing_feature", reason, _get_expected_failure_item)
 
 
 def irrelevant(condition=None, library=None, weblog_variant=None, reason=None):
     """decorator, allow to mark a test function/class as not relevant"""
-
     skip = _should_skip(library=library, weblog_variant=weblog_variant, condition=condition)
-
-    def decorator(function_or_class):
-
-        if inspect.isclass(function_or_class):
-            assert condition is not None, _MANIFEST_ERROR_MESSAGE
-
-        if not skip:
-            return function_or_class
-
-        full_reason = "irrelevant" if reason is None else f"irrelevant ({reason})"
-        return _get_skipped_item(function_or_class, full_reason)
-
-    return decorator
+    noxfail = True
+    return partial(decorator, skip, noxfail, condition, "irrelevant", reason, _get_expected_failure_item)
 
 
 def bug(condition=None, library=None, weblog_variant=None, reason=None, force_skip: bool = False):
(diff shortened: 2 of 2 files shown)


cbeauchesne:
rebasing to fix issue on python

cbeauchesne (changes requested):
CI is failing

mabdinur:
> CI is failing

The java parametric test failure was due to a flaky test. After one rerun that scenario is now green.  

The php failures seem unrelated to this PR.  The same tests are also failing on the main branch: https://github.com/DataDog/system-tests/actions/runs/12257217102/job/34195001337.

cbeauchesne:
For PHP : #3658

tests/parametric/test_span_events.py
```diff
@@ -9,7 +9,7 @@
 @rfc("https://docs.google.com/document/d/1cVod_VI7Yruq8U9dfMRFJd7npDu-uBpste2IB04GyaQ")
 @scenarios.parametric
 @features.span_events
-@missing_feature(reason="Agent does not advertise native span events serialization support yet")
+@missing_feature(True, reason="Agent does not advertise native span events serialization support yet")
```
cbeauchesne, line 12: Is this change necessary? by not specifying a condition, it should be flagged.

utils/_decorators.py
```diff
@@ -106,109 +94,52 @@ def _should_skip(condition=None, library=None, weblog_variant=None):
...
-        if inspect.isclass(function_or_class):
-            assert condition is not None or (library is None and weblog_variant is None), _MANIFEST_ERROR_MESSAGE
-
-        if not skip:
-            return function_or_class
-
-        full_reason = "missing_feature" if reason is None else f"missing_feature ({reason})"
-
-        return _get_expected_failure_item(function_or_class, full_reason, force_skip=force_skip)
+def decorator(expected_to_fail, force_skip, condition, decorator_type, reason, function_or_class):
```
cbeauchesne, line 97: Prevent this method to be exported by error : 

```suggestion
def _decorator(expected_to_fail, force_skip, condition, decorator_type, reason, function_or_class):
```

utils/_decorators.py
```diff
@@ -49,20 +50,7 @@ def _ensure_jira_ticket_as_reason(item, reason: str):
...
-        item.pytestmark.append(pytest.mark.skip(reason=skip_reason))
-
-    else:
-        raise ValueError(f"Unexpected skipped object: {item}")
-
-    return item
-
-
-def _get_expected_failure_item(item, skip_reason, force_skip: bool = False):
+def _get_expected_failure_item(item, skip_reason, force_skip):
```
cbeauchesne, line 53: I feel that name is not the best one, this function covers two use case (skipped and expected_failure), so its name should not make us think it cover only one. And the argument `force_skip` adds a little bit to the confusion

What about 

```python
def _add_pytest_marker(item, reason:str, pytest_marker):
   ...

# Or even 

def _get_decorated_item(item, reason:str, pytest_marker):
    ...

# usage for expected failure:
_add_pytest_marker(item, "blabla", pytest.mark.skip if force_skip else pytest.mark.xfail)

# usage for irrelevant and flaky
_add_pytest_marker(item, "blabla", pytest.mark.skip)
```

It would make the intent inside the caller very clear, WDYT ?

utils/_decorators.py
cbeauchesne: Could you also make `decorator_type`  an string enum, rather a free string ?

utils/_decorators.py
```diff
@@ -106,109 +94,52 @@ def _should_skip(condition=None, library=None, weblog_variant=None):
...
-
-        if not skip:
-            return function_or_class
-
-        full_reason = "missing_feature" if reason is None else f"missing_feature ({reason})"
-
-        return _get_expected_failure_item(function_or_class, full_reason, force_skip=force_skip)
+def decorator(expected_to_fail, force_skip, condition, decorator_type, reason, function_or_class):
+    if inspect.isclass(function_or_class):
+        assert condition is not None, _MANIFEST_ERROR_MESSAGE
```
cbeauchesne, line 99: Allow to entirely skip a class for all libs : 

```suggestion
        assert condition is not None or (library is None and weblog_variant is None), _MANIFEST_ERROR_MESSAGE
```

utils/_decorators.py
```diff
@@ -106,109 +94,52 @@ def _should_skip(condition=None, library=None, weblog_variant=None):
...
+    return _get_expected_failure_item(function_or_class, full_reason, force_skip)
 
-        if not skip:
-            return function_or_class
 
-        full_reason = "incomplete_test_app" if reason is None else f"incomplete_test_app ({reason})"
+def missing_feature(condition=None, library=None, weblog_variant=None, reason=None, force_skip: bool = False):
+    """decorator, allow to mark a test function/class as missing"""
+    expected_to_fail = _expected_to_fail(library=library, weblog_variant=weblog_variant, condition=condition)
+    return partial(decorator, expected_to_fail, force_skip, condition, "missing_feature", reason)
```
cbeauchesne, line 113: Using named argument will be easier to maintain 

```suggestion
    return partial(decorator, expected_to_fail=expected_to_fail, force_skip=force_skip, condition=condition, decorator_type="missing_feature", reason=reason)
```

utils/_decorators.py
```diff
@@ -106,109 +94,52 @@ def _should_skip(condition=None, library=None, weblog_variant=None):
...
 
-        return _get_expected_failure_item(function_or_class, full_reason, force_skip=force_skip)
 
-    return decorator
+def incomplete_test_app(condition=None, library=None, weblog_variant=None, reason=None):
+    """Decorator, allow to mark a test function/class as not compatible with the tested application"""
+    expected_to_fail = _expected_to_fail(library=library, weblog_variant=weblog_variant, condition=condition)
+    force_skip = False
+    return partial(decorator, expected_to_fail, force_skip, condition, "incomplete_test_app", reason)
```
cbeauchesne, line 120: Same comment as above : 

```suggestion
    return partial(decorator, expected_to_fail=expected_to_fail, force_skip=False, condition=condition, decorator_type="incomplete_test_app", reason=reason)
```

utils/_decorators.py
```diff
@@ -106,109 +94,52 @@ def _should_skip(condition=None, library=None, weblog_variant=None):
...
+        return function_or_class
+    return _get_expected_failure_item(function_or_class, full_reason, force_skip)
 
-        if not skip:
-            return function_or_class
 
-        full_reason = "incomplete_test_app" if reason is None else f"incomplete_test_app ({reason})"
+def missing_feature(condition=None, library=None, weblog_variant=None, reason=None, force_skip: bool = False):
+    """decorator, allow to mark a test function/class as missing"""
+    expected_to_fail = _expected_to_fail(library=library, weblog_variant=weblog_variant, condition=condition)
```
cbeauchesne, line 112: If I'm not wrong, we could also move the `_expected_to_fail` call inside `_decorator` ?

cbeauchesne (changes requested):

utils/_decorators.py
mabdinur: `def _add_pytest_marker(item, reason:str, pytest_marker):` looks good to me

tests/parametric/test_span_events.py
mabdinur: This change was require to use `assert condition is not None, _MANIFEST_ERROR_MESSAGE` with the missing_feature annotation. This is the only class annotation that sets condition=None.

mabdinur: In a follow up PR I am planning to prevent folks from annotation classes with the bug/flaky/irrelevant/... decorators. These decorations should only be applied to methods. In all other scenarios the manifest file should be used.

utils/_decorators.py
mabdinur: We can only use positional arguments in partial(....)

tests/parametric/test_span_events.py
cbeauchesne: > These decorations should only be applied to methods. 

Unfortunately, no. Manifests can handle only a subset of what is possible with decorators. This subset covers 99% of use cases, but for the last bit, we need to keep this capacity. Exemples (not exhaustive) : 

```python
@bug(context.library == "python@1.2.3") # previous and next version are ok
@bug(context.library >= "python@1.2.3" and context.agent_version > "4.5.6") # integration is broken
class Test_NoLuck:
   ...
```

mabdinur, line 12:
```suggestion
@missing_feature(reason="Agent does not advertise native span events serialization support yet")
```

utils/_decorators.py
```diff
@@ -14,6 +16,14 @@ def configure(config: pytest.Config):
     pass  # nothing to do right now
 
 
+class DecoratorType(enum.Enum):
```
cbeauchesne, line 19: Prevent this to be exported by error, and use `StrEnum` (less prone to typo error when values are `str`) : 

```suggestion
class _DecoratorType(enum.StrEnum):
```

We're getting close !

utils/_decorators.py
mabdinur: This fails:

```
from functools import partial 

def method(arg1, arg2, arg3):
    print(f"arg1 is [{arg1}]")
    print(f"arg2 is [{arg2}]")
    print(f"arg3 is [{arg3}]")

x = partial(method, arg2="I'm arg2", arg1="I'm arg1")

x("I'm arg3") # but x(arg3="I'm arg3") would work
```

cbeauchesne: I't because you gave `arg1` in the partial, and also in the call of `x`, so python does not know what is the value for `arg1`. But if there is no ambiguity, it's fine : 

``` python
from functools import partial 

def method(arg1, arg2, arg3):
    print(f"arg1 is [{arg1}]")
    print(f"arg2 is [{arg2}]")
    print(f"arg2 is [{arg3}]")

x = partial(method, arg3="I'm arg3", arg2="I'm arg2")

x("I'm arg1")
```

``` bash
> python tt.py 
arg1 is [I'm arg1]
arg2 is [I'm arg2]
arg2 is [I'm arg3]
```

Th key is that arguments you do not give to `partial`, and won't be passed as named argument must be at the begining of the argument list. It's not working here because the argument not provided to `partial` is `function_or_class`, and it's the last in the list. If you put it in first, it works fine : 

```python
def _decorator(function_or_class, marker, decorator_type, condition, library, weblog_variant, reason):
```

If you don't mind, I'd like to have this, I had very bad experience with positional arguments and `partial`, as it's very easy to mess up.

To be sure I'm not saying crap, I've tested locally. So if you're ok with my suggestion, I can push the commit, just let me know 😄

mabdinur: The pytest decorator is called with named arguments so `function_or_class` must be the first argument in the partial (if we plan to call pass the other arguments as keyword arguments). That makes sense. I don't have a preference either way (positional vs keyword), I'll make the change.

cbeauchesne (approved):
Failures are not related

Thanks for tackling this !

PR #3745, opened 2024-12-25
Khanayan123/add consistent config system tests

khanayan123:
## Motivation

Adds new system tests to assert config consistency

## Changes
Adds system tests for DD_TRACE_SAMPLING_RULES, DD_LOG_INEJCTION, DD_RUNTIME_METRICS_ENABLED, & DD_TRACE_128_BIT_TRACEID_LOGGING_ENABLED, DD_TRACE_AGENT_URL, as per https://docs.google.com/document/d/1kI-gTAKghfcwI7YzKhqRv2ExUstcHqADIWA4-TZ387o/edit?usp=sharing

DD_TRACE_128_BIT_TRACEID_LOGGING_ENABLED tests were tested against this PR:
https://github.com/DataDog/dd-trace-js/pull/5021


utils/build/docker/nodejs/express/app.js
```diff
@@ -237,6 +237,14 @@ app.get('/kafka/consume', (req, res) => {
     })
 })
 
+app.get('/log/library', (req, res) => {
```
mabdinur, line 240: Can we add a test for this new endpoint here: https://github.com/DataDog/system-tests/blob/main/tests/parametric/test_parametric_endpoints.py#L4. This will allow us to report to the feature parity dashboard on which endpoints are not implemented in different languages.

Also to make this endpoint more useful can we add the following required arguments:
 - log level (warn/error/info/debug/...), we can use an enum here and ensure the casing is always consistent.
 - message (right now it's optional)

manifests/cpp.yml
```diff
@@ -241,8 +241,14 @@ tests/:
     Test_Config_HttpServerErrorStatuses_FeatureFlagCustom: missing_feature
     Test_Config_IntegrationEnabled_False: missing_feature
     Test_Config_IntegrationEnabled_True: missing_feature
+    Test_Config_LogInjection_128Bit_TradeId_Default: missing_feature
```
mabdinur, line 244: Instead of missing feature can we use https://github.com/DataDog/system-tests/blob/25c79a3502d69e30d8ccf0461d15501ccbd0addf/docs/edit/skip-tests.md?plain=1#L5. Some of these fail because the the parametric apps do not implement the new endpoint.

cbeauchesne (commented):
could you update your branch? it'll fix the issue on ruby

tests/parametric/test_128_bit_traceids.py
```diff
@@ -193,7 +193,6 @@ def test_datadog_128_bit_generation_enabled(self, test_agent, test_library):
 
     @missing_feature(context.library == "golang", reason="not implemented")
     @missing_feature(context.library < "java@1.24.0", reason="Implemented in 1.24.0")
-    @missing_feature(context.library == "nodejs", reason="not implemented")
     @missing_feature(context.library == "ruby", reason="not implemented")
     @pytest.mark.parametrize("library_env", [{"DD_TRACE_PROPAGATION_STYLE": "Datadog"}])
     def test_datadog_128_bit_generation_enabled_by_default(self, test_agent, test_library):
```
simon-id, line 198: would be nice to have this be a version check instead of removing the whole line, just like java is doing right above.
(same thing in line 412)

manifests/java.yml
```diff
@@ -1648,8 +1648,16 @@ tests/:
     Test_Config_IntegrationEnabled_True:
       '*': irrelevant (kafka endpoints are not implemented)
       spring-boot: v1.42.0
+    Test_Config_IntegrationEnabled_False: irrelevant (not applicable to Go because of how they do auto instrumentation)
+    Test_Config_IntegrationEnabled_True: irrelevant (not applicable to Go because of how they do auto instrumentation)
```
simon-id, line 1652: this is the java manifest

manifests/nodejs.yml
```diff
@@ -849,13 +851,28 @@ tests/:
...
+      express4-typescript: irrelevant
+      nextjs: irrelevant
+    Test_Config_LogInjection_Default:
+      '*': *ref_3_0_0
+      express4-typescript: irrelevant
+      nextjs: irrelevant
+    Test_Config_LogInjection_Enabled:
+      '*': *ref_3_0_0
+      express4-typescript: irrelevant
+      nextjs: irrelevant
```
simon-id, line 871: are you sure those are irrelevant ? and not missing_feature instead ?

khanayan123: Those apps just don't implement the route that tests the feature, but the feature exists

tests/parametric/test_trace_sampling.py
```diff
@@ -147,6 +147,75 @@ def test_trace_sampled_by_trace_sampling_rule_glob_match(self, test_agent, test_
...
+                "DD_TRACE_SAMPLING_RULES": json.dumps(
+                    [
+                        {"name": "wEb.*", "sample_rate": 1},
+                    ]
+                ),
+            }
+        ],
+    )
+    def test_trace_sampling_rules_format_not_needed(self, test_agent, test_library):
+        """Glob should work by default without specifying the format"""
```
mabdinur, line 207: The assertions/spans in this test look identical to test_field_case_insensitivity. S
hould we add the `library_env` in this test to  test_field_case_insensitivity parameterized test? This way we have one test with 5 scenarios.

tests/test_config_consistency.py
```diff
@@ -415,3 +419,89 @@ def test_integration_enabled_true(self):
...
+class Test_Config_LogInjection_Enabled:
+    """Verify behavior of integrations automatic spans"""
+
+    def setup_log_injection_enabled(self):
+        self.message = "msg"
+        self.r = weblog.get("/log/library", params={"msg": self.message})
+
+    def test_log_injection_enabled(self):
+        assert self.r.status_code == 200
+        pattern = rf'"dd":\{{"trace_id":"[^"]+","span_id":"\d+","service":"[^"]+","version":"[^"]+","env":"[^"]+"\}},"msg":"{self.message}"'
```
mabdinur, line 436: Should we make this pattern more permissive? If there's whitespace between the entries then the pattern will fail to match. Also it requires log correlation fields to be wrapped in curly brackets and the message to printed after a comma. Also the values may not always be wrapped in quotes. 

This is an example of a log message generated by the python tracer: 

Python docs: https://docs.datadoghq.com/tracing/other_telemetry/connect_logs_and_traces/python/

```
import ddtrace
import logging

FORMAT = ('%(asctime)s %(levelname)s [%(name)s] [%(filename)s:%(lineno)d] '
          '[dd.service=%(dd.service)s dd.env=%(dd.env)s dd.version=%(dd.version)s dd.trace_id=%(dd.trace_id)s dd.span_id=%(dd.span_id)s] '
          '- %(message)s')
logging.basicConfig(format=FORMAT)

log = logging.getLogger(__name__)

with ddtrace.tracer.trace("test"):
    log.error("test")
```

Output:
`2025-01-06 17:21:00,421 ERROR [__main__] [reproduction.py:20] [dd.service=Downloads dd.env= dd.version= dd.trace_id=677c574c000000001679fe710d0f33ba dd.span_id=10973182204132464872] - test`

manifests/nodejs.yml
simon-id: irrelevant is for when something really makes no sense to test

tests/parametric/test_trace_sampling.py
khanayan123: I could just remove it from all the tests since I believe no language currently requires the glob environment variable

tests/test_config_consistency.py
khanayan123: I think the other way to solve this issue is to have weblog apps send json formatted logs

khanayan123: That would make parsing easier

tests/test_config_consistency.py
```diff
@@ -415,3 +421,113 @@ def test_integration_enabled_true(self):
...
+        assert re.match(r"\d+", trace_id), f"Invalid 64-bit trace_id: {trace_id}"
+
+
+@rfc("https://docs.google.com/document/d/1kI-gTAKghfcwI7YzKhqRv2ExUstcHqADIWA4-TZ387o/edit#heading=h.8v16cioi7qxp")
+@scenarios.runtime_metrics_enabled
+@features.tracing_configuration_consistency
+class Test_Config_RuntimeMetrics_Enabled:
+    # This test verifies runtime metrics by asserting the prescene of a metric in the dogstatsd endpoint
+    def test_config_runtimemetrics_enabled(self):
+        data = list(interfaces.library.get_data("/dogstatsd/v2/proxy"))[0]
```
zacharycmontoya, line 504: There might be a race condition here. At least in .NET we submit runtime metrics every 10s, so we'll need to add some sort of wait or configure the delay in a similar fashion for both the Enabled case and the Default case

manifests/ruby.yml
```diff
@@ -489,8 +489,14 @@ tests/:
     Test_Config_IntegrationEnabled_True:
       '*': irrelevant (endpoint not implemented)
       rails70: v2.0.0
+    Test_Config_LogInjection_128Bit_TradeId_Default: incomplete_test_app
+    Test_Config_LogInjection_128Bit_TradeId_Disabled: incomplete_test_app
+    Test_Config_LogInjection_Default: incomplete_test_app
+    Test_Config_LogInjection_Enabled: incomplete_test_app
     Test_Config_ObfuscationQueryStringRegexp_Configured: missing_feature
     Test_Config_ObfuscationQueryStringRegexp_Empty: missing_feature (environment variable is not supported)
+    Test_Config_RuntimeMetrics_Default: incomplete_test_app
```
mabdinur, line 498: Can we add a reason to these incomplete_test_app decorators

tests/test_config_consistency.py
```diff
@@ -415,3 +421,116 @@ def test_integration_enabled_true(self):
...
+    def setup_log_injection_enabled(self):
+        self.message = "msg"
+        self.r = weblog.get("/log/library", params={"msg": self.message})
+
+    def test_log_injection_enabled(self):
+        assert self.r.status_code == 200
+        pattern = r'"dd":\{[^}]*\}'
+        stdout.assert_presence(pattern)
+        dd = parse_log_injection_message(self.message)
+        print(55, dd)
```
mabdinur, line 441: Do we need this print?

tests/test_config_consistency.py
```diff
@@ -415,3 +421,116 @@ def test_integration_enabled_true(self):
...
+@features.tracing_configuration_consistency
+class Test_Config_LogInjection_Default:
+    def setup_log_injection_default(self):
+        self.message = "msg"
+        self.r = weblog.get("/log/library", params={"msg": self.message})
+
+    def test_log_injection_default(self):
+        assert self.r.status_code == 200
+        pattern = r'"dd":\{[^}]*\}'
+        stdout.assert_absence(pattern)
```
mabdinur, line 459: We should avoid negative tests. This test could pass due to an unexpected format or the feature is not implemented. 

cc: @cbeauchesne, @bm1549: This is an example of a negative test. Here we want to verify that log injection is disabled when the DD_LOG_INJECTION_ENABLED environment variable is not set. If a library does not support log injection at all then this test would still pass and show up as an easy win. 

Should this test exist? I think it still provides value even if it could be misinterpreted in the FPD.

mabdinur (approved):

tests/test_config_consistency.py
```diff
@@ -415,3 +421,116 @@ def test_integration_enabled_true(self):
...
+            for line in lines:
+                if runtime_metrics[context.library.library] in line:
+                    metric_found = True
+                    break
+            assert metric_found, f"The metric {runtime_metrics[context.library.library]} was not found in any line"
+            break
+
+
+@rfc("https://docs.google.com/document/d/1kI-gTAKghfcwI7YzKhqRv2ExUstcHqADIWA4-TZ387o/edit#heading=h.8v16cioi7qxp")
+@scenarios.tracing_config_nondefault
```
mtoffl01, line 516: `@scenarios.default`?

tests/parametric/test_128_bit_traceids.py
```diff
@@ -422,6 +421,26 @@ def test_w3c_128_bit_propagation_tid_in_chunk_root(self, test_agent, test_librar
...
+        assert len(trace) == 2
+        first_span = find_first_span_in_trace_payload(trace)
+        tid_chunk_root = first_span["meta"].get("_dd.p.tid")
+        assert tid_chunk_root is not None
+
+    @missing_feature(context.library == "ruby", reason="not implemented")
+    @missing_feature(context.library == "java", reason="not implemented")
+    @pytest.mark.parametrize(
+        "library_env",
+        [{"DD_TRACE_PROPAGATION_STYLE": "tracecontext", "DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED": "true"}],
```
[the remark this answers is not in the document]
khanayan123: The test in the file test_datadog_128_bit_generation_enabled_by_default already verifies that 128-bit generation is enabled by default. Other tests explicitly configure this setting to ensure that other aspects of the feature can be tested, even in scenarios where the language may not enable this functionality by default

mtoffl01 (approved):

tests/test_config_consistency.py
cbeauchesne: is there anything (telemetry, logs) that shows that the tracers has disabled log injection ?

tests/parametric/test_128_bit_traceids.py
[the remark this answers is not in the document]
mtoffl01: Hmm.. okay. Like we don't care whether DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED is true by default in this test, we just want to check what the behavior is when it is enabled?

I don't think we do that for other tests, but this isn't a blocker, it's fine. Maybe add a comment or something

tests/test_config_consistency.py
```diff
@@ -415,3 +421,123 @@ def test_integration_enabled_true(self):
...
+        dd = parse_log_injection_message(self.message)
+        trace_id = dd.get("trace_id")
+        assert re.match(r"^[0-9a-f]{32}$", trace_id), f"Invalid 128-bit trace_id: {trace_id}"
+
+
+@rfc("https://docs.google.com/document/d/1kI-gTAKghfcwI7YzKhqRv2ExUstcHqADIWA4-TZ387o/edit#heading=h.8v16cioi7qxp")
+@scenarios.tracing_config_nondefault_3
+@features.tracing_configuration_consistency
+class Test_Config_LogInjection_128Bit_TradeId_Disabled:
+    """Verify 128 bit traceid are disabled in log injection when DD_TRACE_128BIT_TRACE_ID=false"""
```
mtoffl01, line 486:
```suggestion
    """Verify 128 bit traceid are disabled in log injection when DD_TRACE_128_BIT_TRACEID_LOGGING_ENABLED=false"""
```

tests/test_config_consistency.py
```diff
@@ -415,3 +421,123 @@ def test_integration_enabled_true(self):
...
+        self.message = "msg"
+        self.r = weblog.get("/log/library", params={"msg": self.message})
+
+    def test_log_injection_128bit_traceid_disabled(self):
+        assert self.r.status_code == 200
+        pattern = r'"dd":\{[^}]*\}'
+        stdout.assert_presence(pattern)
+        dd = parse_log_injection_message(self.message)
+        trace_id = dd.get("trace_id")
+        assert re.match(r"\d+", trace_id), f"Invalid 64-bit trace_id: {trace_id}"
```
mtoffl01, line 498:
```suggestion
        assert re.match(r"^\d{1,20}$", trace_id), f"Invalid 64-bit trace_id: {trace_id}"
```
?

utils/_context/_scenarios/__init__.py
```diff
@@ -761,6 +764,12 @@ class _Scenarios:
     external_processing = ExternalProcessingScenario("EXTERNAL_PROCESSING")
     ipv6 = IPV6Scenario("IPV6")
 
+    runtime_metrics_enabled = EndToEndScenario(
```
cbeauchesne, line 767: Can you add this scenario in `.github/workflows/run-end-to-end.yml` ?

cbeauchesne (changes requested):

cbeauchesne:
seems good, just waiting for the CI to finish to approve the PR

cbeauchesne (approved):

PR #3812, opened 2025-01-14
Add test to check absence of client computed stats

estringana:
## Motivation

The added test is a regression test which caused a incident on PHP. When ASM Standalone was implemented, it was required to send the header `Datadog-Client-Computed-Stats` as `yes` when it was a ASM Standalone request. However the way it was implemented in PHP was to add the header `Datadog-Client-Computed-Stats` always. Then the value  of the header was `yes` when it was a ASM Standalone request and `no` when not. This was generating the agent to never compute stats as it was interpreting the value `no` as if it was `yes`.

## Workflow

## Reviewer checklist

### DIFF
--- a/tests/appsec/test_asm_standalone.py
+++ b/tests/appsec/test_asm_standalone.py
@@ -742,6 +742,30 @@ class SCAStandalone_Telemetry_Base:
 
 @rfc("https://docs.google.com/document/d/12NBx-nD-IoQEMiCRnJXneq4Be7cbtSc6pJLOFUWTpNE/edit")
 @features.appsec_standalone
+@scenarios.appsec_no_stats
+class Test_AppSecStandalone_NotEnabled:
+    """Test expected behaviour when standalone is not enabled."""
+
+    def setup_client_computed_stats_header_is_not_present(self):
+        trace_id = 1212121212121212122
+        self.r = weblog.get(
+            "/",
+            headers={
+                "x-datadog-trace-id": str(trace_id),
+            },
+        )
+
+    def test_client_computed_stats_header_is_not_present(self):
+        spans_checked = 0
+        for data, _, span in interfaces.library.get_spans(request=self.r):
+            assert span["trace_id"] == 1212121212121212122
+            assert "datadog-client-computed-stats" not in [x.lower() for x, y in data["request"]["headers"]]
+            spans_checked += 1
+        assert spans_checked == 1
+
+
+@rfc("https://docs.google.com/document/d/12NBx-nD-IoQEMiCRnJXneq4Be7cbtSc6pJLOFUWTpNE/edit")
+@features.appsec_standalone
 @scenarios.appsec_standalone
 class Test_AppSecStandalone_UpstreamPropagation(AppSecStandalone_UpstreamPropagation_Base):
     """APPSEC correctly propagates AppSec events in distributing tracing with DD_EXPERIMENTAL_APPSEC_STANDALONE_ENABLED=true."""
--- a/tests/appsec/test_asm_standalone.py
+++ b/tests/appsec/test_asm_standalone.py
@@ -748,10 +748,12 @@ class Test_AppSecStandalone_NotEnabled:
 
     def setup_client_computed_stats_header_is_not_present(self):
         trace_id = 1212121212121212122
+        parent_id = 34343434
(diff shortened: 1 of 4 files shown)


tests/appsec/test_asm_standalone.py
jandro996, line 1: I'm not sure if there is a current test in another scenario that checks this datadog-client-computed-stats header, but IMHO we need one more test case that checks datadog-client-computed-stats=true when standalone billing is disabled and metrics enabled, to ensure the previous is working as before the asm feature was implemented.

estringana: Not sure if I'm following you correctly. How would you make APM to send the header `datadog-client-computed-stats=true`? In PHP that header was not sent

jandro996: Maybe it's something that only happens in java, but the header is set with empty value, but to keep it simple we can continue with the PR as it is

utils/_context/_scenarios/__init__.py
```diff
@@ -63,6 +63,12 @@ class _Scenarios:
 
     profiling = ProfilingScenario("PROFILING")
 
+    asm_e2e = EndToEndScenario(
```
cbeauchesne, line 66: What about using the default scenario? the only APPSEC env var configured here is `DD_IAST_SECURITY_CONTROLS_CONFIGURATION`, all others values will be the same.

cbeauchesne (changes requested):

utils/_context/_scenarios/__init__.py
estringana: The default scenario has `DD_TRACE_COMPUTE_STATS=true` which is impacting the test I did.

cbeauchesne: ah :) Could you add this in the scenario's doc ?

estringana: Done! By the way, It seems as this scenario is not being executed here on github. Any idea why?

cbeauchesne: Because you need to add it in `.github/workflows/end-to-end.yml`. Speaking of which, could you change its name, maybe APPSEC_NO_STATS ?

cbeauchesne (changes requested):
Need to add new scenario in the CI

utils/_context/_scenarios/__init__.py
estringana: I did what you said. But it is still not executed on the pipeline :(

cbeauchesne: Indeed, I found the reason, see my request change. Sorry, I should have seen that on my first review 😊

estringana: All good mate! I applied your request change. Thanks for the help

cbeauchesne: it's now executed : https://github.com/DataDog/system-tests/actions/runs/12990672620/job/36226967654?pr=3812#step:14:58

😉

tests/appsec/test_asm_standalone.py
```diff
@@ -741,6 +741,30 @@ def test_app_dependencies_loaded(self):
...
+    """Test expected behaviour when standalone is not enabled."""
+
+    def setup_client_computed_stats_header_is_not_present(self):
+        trace_id = 1212121212121212122
+        self.r = weblog.get(
+            "/",
+            headers={
+                "x-datadog-trace-id": str(trace_id),
+            },
+        )
```
iunanua, line 757: I believe you have to provide the `x-datadog-parent-id` header along with `x-datadog-trace-id`.
If not `x-datadog-parent-id` is ignored (at least this is how [nodejs tracer works](https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/opentracing/propagation/text_map.js#L502))

```suggestion
        trace_id = 1212121212121212122
        parent_id = 34343434
        self.r = weblog.get(
            "/",
            headers={
                "x-datadog-trace-id": str(trace_id),
                "x-datadog-parent-id": str(parent_id),
            },
        )
```

estringana: Thanks!

cbeauchesne (approved):
Failures are not related

PR #3899, opened 2025-01-27
bugfix(ssi): fix profiling result checking

betterengineering:
## Motivation

The original code would provide a non-zero exit code, which would cause the loop to retry and eventually time out:
```
            if isinstance(data, list) and len(data) > 0:
                return r.status_code
            return -1
```

I missed this behavior given it wasn't explicit. This change explicitly checks the result when querying for a profile and raises a typed exception if it does not exist.

## Changes

This commit resolves an issue where my refactor no longer checked the result when querying for a profile. This changes resolves that issue.

## Workflow

## Reviewer checklist

### DIFF
--- a/utils/onboarding/backend_interface.py
+++ b/utils/onboarding/backend_interface.py
@@ -9,10 +9,14 @@ from utils.tools import logger
 API_HOST = "https://dd.datadoghq.com"
 
 
+class ProfileNotFoundError(Exception):
+    pass
+
+
 def wait_backend_trace_id(trace_id, profile: bool = False, validator=None):
     runtime_id = _query_for_trace_id(trace_id, validator=validator)
-    if profile:
-        _query_for_profile(runtime_id)
+    if profile and not _query_for_profile(runtime_id):
+        raise ProfileNotFoundError("Profile events not found for runtime_id: {runtime_id}")
 
 
 def _headers():


utils/onboarding/backend_interface.py
```diff
@@ -9,10 +9,14 @@
...
 
+class ProfileNotFoundError(Exception):
+    pass
+
+
 def wait_backend_trace_id(trace_id, profile: bool = False, validator=None):
     runtime_id = _query_for_trace_id(trace_id, validator=validator)
-    if profile:
-        _query_for_profile(runtime_id)
+    if profile and not _query_for_profile(runtime_id):
```
szegedi, line 18: This will almost always fail now, as at the time of the first query the profile is almost certain to not be in the backend yet.
`_query_for_profile` needs to keep retrying if it gets a 200 with empty array until the timeout expires. The reason for this is that there can be quite a bit of time before the profile materializes after the trace, for multiple reasons:
* The profiler in the tested app will only send the profile after 10-60 seconds (normal period to send a profile is every 60 seconds, the test framework sets an env var on tested webapps that overrides it to 10 seconds but not all languages' profilers honor it.)
* Even after the library sent the profile to the agent (and it to the backend), the profile might need some time to propagate through queues before it shows up in query results against the event platform.

I believe the best course of action might be moving the logic into `_make_request` by either passing a function that performs payload validation on 200 response, or pass a boolean flag for "profile" and have a separate `if` for this case in the 200 handler and only return when `isinstance(data, list) and len(data) > 0` is satisfied.

szegedi (changes requested):

PR #3929, opened 2025-01-29
baggage weblog tests

rachelyangdog:
## Motivation

Adding new weblog tests to support tracers that implement baggage separate from spans. 
Tests injection and extraction logic of baggage ensuring that limits (such as max bytes and max items) and encoding rules are respected across all tracers. 
## Workflow

## Reviewer checklist


utils/_context/_scenarios/__init__.py
```diff
@@ -80,6 +80,18 @@ class _Scenarios:
         doc="Test W3C trace style",
     )
 
+    datadog_baggage_propagation = EndToEndScenario(
```
zacharycmontoya, line 83: Can we re-use the default scenario (no `DD_TRACE_PROPAGATION_STYLE` configuration) instead of adding a new scenario? We should re-use scenarios as much as possible because each scenario adds the overhead of running the agent container, weblog containers, and various test proxies

tests/test_baggage.py
```diff
@@ -0,0 +1,161 @@
...
+        assert self.r.status_code == 200
+        data = json.loads(self.r.text)
+        headers = data["request_headers"]
+        if isinstance(headers, dict):
+            assert "baggage" not in headers
+        elif isinstance(headers, list):
+            assert all(header.get("key") != "baggage" for header in headers)
+
+
+@scenarios.only_baggage_propagation
```
zacharycmontoya, line 88: Do we need a separate scenario to only test the baggage propagator? I suspect that only having the `baggage` header in the incoming request (like you have here) would be sufficient, unless there are some edge cases that you had in mind

tests/test_baggage.py
```diff
@@ -0,0 +1,161 @@
...
+        # Ensure we respect the max items limit
+        assert len(items) <= self.max_items
+
+
+@scenarios.default
+@features.datadog_baggage_headers
+class Test_Baggage_Headers_Max_Bytes:
+    def setup_main(self):
+        self.max_bytes = 8192
+        baggage_items = {
```
zacharycmontoya, line 138: nit: I think this test case can be revised so it's even simpler and easier for authors. If you only test two keys and set the value lengths to `self.max_bytes // 2` it makes it much easier to understand that:
- 8192/2 will be even (no weird modulo's)
- It's more apparent that the length of the keys, equal signs, and delimiter will make the total exceed the max_bytes

zacharycmontoya (approved):
LGTM with some small suggestions!

tests/test_baggage.py
```diff
@@ -0,0 +1,161 @@
+import json
+from utils import weblog, interfaces, features, scenarios
+
+
+def extract_baggage_value(request_headers):
```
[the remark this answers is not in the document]
rachelyangdog: I agree! I think separating it in a separate PR is preferred so that a reviewer from each language can approve the change. I will create a follow-up PR and link it to this PR. 😁

manifests/dotnet.yml
```diff
@@ -552,6 +552,13 @@ tests/:
   stats/:
     test_miscs.py:
       Test_Miscs: missing_feature
+  test_baggage.py:
+    Test_Baggage_Headers_Basic: v3.6.0
+    Test_Baggage_Headers_Malformed: v3.6.0
+    Test_Baggage_Headers_Malformed2: v3.6.0
```
mhlidd, line 558:
```suggestion
    Test_Baggage_Headers_Malformed_2: v3.6.0
```

Can we make this change for readability, similarly to how we do it for Scenarios? (e.g. `TRACING_CONFIG_NONDEFAULT_2`

rachelyangdog: I can make this change in a new PR! I just wanna merge this PR ASAP and the CI is taking extraaa long to run. I'll make sure to introduce this change in a separate PR.

tests/test_baggage.py
```diff
@@ -0,0 +1,164 @@
...
+    def test_main(self):
+        interfaces.library.assert_trace_exists(self.r)
+        assert self.r.status_code == 200
+        data = json.loads(self.r.text)
+        baggage_header_value = extract_baggage_value(data["request_headers"])
+        assert baggage_header_value is not None
+        header_str = baggage_header_value[0] if isinstance(baggage_header_value, list) else baggage_header_value
+        items = header_str.split(",")
+        # Ensure we respect the max items limit
+        assert len(items) <= self.max_items
```
bm1549, line 130: This should be == not <=, right? It'd still pass ofc, but it'd be more strict with ==

rachelyangdog: updated!

PR #4044, opened 2025-02-10
Protobuf system test for schema tracking

vandonr:
## Motivation

Follow up on the development of this new feature, only available in Java and dotnet for now.
All languages are supposed to produce the same schema ID when extracting "from" the same proto file, so this is a perfect use case for a system test.

depends on https://github.com/DataDog/dd-trace-dotnet/pull/6647
and https://github.com/DataDog/dd-trace-java/pull/8403

## Changes

- added a test for this use case that inspects the tags that get generated. A little bit convoluted on the span selection because we do not really control where the protobuf tags will be added
- added 2 endpoints to the java and dotnet weblogs, one doing a protobuf serialization and the other doing the deserialization.

don't be too afraid by the size of the diff, it's mostly because of the generated files: 4,118 lines for C# and 12,618 lines for java 😅 

## Workflow

## Reviewer checklist

* [x] If PR title starts with `[<language>]`, double-check that only `<language>` is impacted by the change


tests/test_protobuf.py
```diff
@@ -0,0 +1,53 @@
...
+        self.serialization_response = weblog.get("/protobuf/serialize")
+        self.deserialization_response = weblog.get(f"/protobuf/deserialize?msg={self.serialization_response.text}")
+
+    @missing_feature(
+        context.library in ["ruby", "cpp", "golang", "nodejs", "php", "python"],
+        reason="no schema tracking for protobuf yet",
+    )
+    @missing_feature(context.library < "dotnet@3.12.0", reason="wrong schema.name written")
+    @missing_feature(context.library < "java@1.47.0", reason="bug on extraction of nested schemas")
+    def test_protobuf(self):
```
robertomonteromiguel, line 19: we should define all these missing_feature in the manifests

robertomonteromiguel (changes requested):

tests/test_protobuf.py
vandonr: oh you mean the library versions ? Or also the languages where it's not available yet ?

robertomonteromiguel: both cases. I think we can safely use this python class without such annotations.

smola (commented):
Ideally the new endpoints should be documented at https://github.com/DataDog/system-tests/blob/main/docs/weblog/README.md

vandonr:
github CI seems to be stuck ? But otherwise we should be good to go

robertomonteromiguel (approved):

PR #4067, opened 2025-02-11
[Tracing] Enable all runtime metrics weblog tests for existing tracer behaviors

zacharycmontoya:
## Motivation

We want to assert all of the known behaviors and expectations of the Runtime Metrics features across our libraries. This gives us an up-to-date test spec for each library.

## Changes

- Enables the runtime metrics tests for Go, Ruby, and Python
- For Ruby, this also required adding the `dogstatsd-ruby` gem to all weblog application variants
- For Python, this also required parsing distributions through the backend `/api/beta/sketches` endpoint and exposing it through the agent interface, via a new `get_sketches` function

## Workflow

## Reviewer checklist


tests/test_config_consistency.py
```diff
@@ -569,12 +569,21 @@ def setup_main(self):
     def test_main(self):
         assert self.req.status_code == 200
 
-        runtime_metrics = [
+        runtime_metrics_gauges = [
```
bm1549, line 572: [nit] Should we change the `time.sleep(10)` above to instead be a poller to reduce the expected runtime for these tests?

tests/test_config_consistency.py
```diff
@@ -605,12 +614,21 @@ def setup_main(self):
...
```
bm1549, line 617: Duplicate code here seems like it should be a helper function like `wait_for_runtime_metrics`

tests/test_config_consistency.py
```diff
@@ -569,12 +569,21 @@ def setup_main(self):
...
+            for _, metric in interfaces.agent.get_sketches()
+            if metric["metric"].startswith("runtime.") or metric["metric"].startswith("jvm.")
+        ]
+
+        assert len(runtime_metrics_gauges) > 0 or len(runtime_metrics_sketches) > 0
+
+        runtime_metrics = runtime_metrics_gauges if len(runtime_metrics_gauges) > 0 else runtime_metrics_sketches
 
         for metric in runtime_metrics:
             tags = {tag.split(":")[0]: tag.split(":")[1] for tag in metric["tags"]}
```
bm1549, line 589: Should we be verifying that `tags.get("lang")` matches the language we expect to see for the running test?

bm1549 (approved):
LGTM with a few comments. No need for re-review after addressing

tests/test_config_consistency.py
zacharycmontoya: Good call, I agree that we should assert what value each library emits for the `lang`. I'll update this to assert the per-language value

tests/test_config_consistency.py
cbeauchesne: it actually should be handled by the library interface wait time, that is done at the very end of the `setup` phase.

emmettbutler (approved):

cbeauchesne (approved):

tests/test_config_consistency.py
zacharycmontoya: Done in ec09375714c3e59d4775219995f9c6a74a929e6e

tests/test_config_consistency.py
zacharycmontoya: Done in c2d226c0d12c1ca2ad95fd432f79e7c683c74961

tests/test_config_consistency.py
zacharycmontoya: I agree, this should be addressed separately but I'm not sure I want to spend time on it immediately since the 10s wait should be sufficient for now without too much downside

PR #4136, opened 2025-02-19
[Debugger] Add telemetry tests

shurivich:
## Motivation
Add telemetry tests

## Workflow

## Reviewer checklist


OmerRaviv:
A few broad strokes feedbacks:
1. We already have two pre-existing test files that test instrumentation telemetry data [tests/test-telemetry.py](https://github.com/DataDog/system-tests/blob/5a67fc359c991cf651e5e9eae3d1c883fd011bcf/tests/test_telemetry.py) and [tests/parametric/test_telemetry.py](https://github.com/DataDog/system-tests/blob/4618e704584d637268ec1a21e8781103387455f8/tests/parametric/test_telemetry.py#L4). I think it would be preferable to re-use one of those approaches, instead of inventing a third :)   
2. As you probably noticed we have an inconsistency in how we currently report the values - e.g. the Python tracer currently reports it as dd_exception_replay_enabled instead of just exception_replay_enabled. This can easily be fixed at intake by adding a normalize rule to the language-specific normalization rules file, in this case [python_config_rules.json](https://github.com/DataDog/dd-go/blob/prod/trace/apps/tracer-telemetry-intake/telemetry-payload/static/python_config_rules.json). After submitting a PR for that, we would need to [run the update script as described here](https://github.com/DataDog/system-tests/blob/main/docs/edit/runbook.md#test_config_telemetry_completeness) to copy the normalization json files over from `dd-go`, and extract out and re-use the existing logic from `test_telemetry.py` to read those json files.
3. I would argue that using approvals / gold files is not really the right fit for these tests and introduce unnecessary complexity, making the tests harder to read. The reason I say this is because after normalizing the names, the output of each and every tracer should be exactly the same.
4.  When we report the configuration values,  we report them in this [format](https://github.com/DataDog/instrumentation-telemetry-api-docs/blob/main/GeneratedDocumentation/ApiDocs/v2/SchemaDocumentation/Schemas/conf_key_value.md). Note that we need to be careful to report the origin (as described in that last link) correctly for tracking and troubleshooting purposes: it will usually be `env_var` because that's what most customers are using, but we should correctly report `code` instead when the user enabled the feature programmatically (which is common in python and ruby), and now that we are adding in-product enablement, we should report `remote-config` when enabled through a click in the UI. As we discussed in the agent sync meeting, we should also ensure that we report `default` when nothing was explicitly set.

PR #4152, opened 2025-02-20
change session fingerprint test to use identify endpoint

simon-id:
## Motivation

No need for one more endpoint...

## Changes

Change the session test to use an existing weblog endpoint


docs/weblog/README.md
```diff
@@ -875,7 +875,7 @@ c377db41-b664-4e30-af57-5df2e803bec7
 Examples:
 - `GET`: `/session/new`
 
-### \[GET\] /session/user
+### **UNUSED** \[GET\] /session/user
```
manuel-alvarez-alvarez, line 878: Maybe we can just remove it as a separate PR

manuel-alvarez-alvarez (approved):

docs/weblog/README.md
simon-id: I think we should remove it once all other libs have removed it :+1:

simon-id: otherwise imagine a dev debugging something and not finding docs for the endpoint he's working on, it's weird

PR #4182, opened 2025-02-24
Modify Config Consistency log injection tests

mtoffl01:
## Motivation
The existing log injection tests were written with Nodejs as the blueprint and therefore expected logs to be in this format: `dd: { trace_id: <value>, span_id: <value> }`. However, Node is the outlier and other tracers produce logs with this format instead: `dd.trace_id: <value>, dd.span_id: <value>`. 

This PR updates test helper functions to use the more popular format, while still accommodating Node. It also references a jira card containing the TODO to align Node with the other libs: https://datadoghq.atlassian.net/browse/APMAPI-1199

## Changes
- Fixes a typo in the test names from "Trade" to "Trace"
- Modifies test helper function logic to support both log message formats

## Workflow

## Reviewer checklist


BridgeAR (commented):
Instead of special handling go, should we not just change the way this is published in the go tracer?

mtoffl01:
@BridgeAR Thanks to your callout, I actually modified the test logic to be more "accurate" -- the original test was written with Nodejs as the blueprint, but with the expectation that it will be enabled for all libraries. I was just the first one to enable it for my library (Golang), so I left the default logic for Node and made an "alternative" option for Go. However, I've since changed the logic so that node's pattern is the "alternative," because most libraries format their logs like Go (I believe).

Your same question still applies though -- "Instead of special handling <for any library>, should we not just change the way this is published in the <affected> tracer?"
My answer is: Maybe, but:
1. The result of both formats is the same: trace id and span id can be queried from the logs via `dd.trace_id` and `dd.span_id` in the logs backend in order to locate the correlated trace (more context here: https://docs.datadoghq.com/tracing/troubleshooting/correlated-logs-not-showing-up-in-the-trace-id-panel?tab=jsonlogs#trace-id-option)
2. I'm not sure what kind of pushback a request to change the format in Node would have, and I'm not sure the value of the change is big enough to warrant the effort (due to # 1 above) 

cc @khanayan123 if you want to weigh in?

BridgeAR:
@mtoffl01 thank you for the update!

If only Node.js is deviating in its format, I would not see a reason against aligning it (my naive expectation is that changing the output does not require much work).

If multiple parsers deviate, let's support both.

mtoffl01:
Moving back to draft as we discuss next steps for Node

tests/test_config_consistency.py
```diff
@@ -662,13 +661,34 @@ def get_runtime_metrics(agent):
...
-# Parse the JSON-formatted log message from stdout and return the 'dd' object
+# Parses the JSON-formatted log message from stdout and returns it
 def parse_log_injection_message(log_message):
     for data in stdout.get_data():
         try:
             message = json.loads(data.get("message"))
         except json.JSONDecodeError:
             continue
+        # For Nodejs
         if message.get("dd") and message.get(log_injection_fields[context.library.library]["message"]) == log_message:
```
mabdinur, line 672: Can we simplify this to:
```suggestion
        if context.library == "nodejs":
```

tests/test_config_consistency.py
```diff
@@ -662,13 +661,34 @@ def get_runtime_metrics(agent):
...
+    return None
+
+
+def parse_log_trace_id(message):
+    if message.get("dd.trace_id"):
+        return message.get("dd.trace_id")
+    # For Nodejs
+    if message.get("trace_id"):
+        return message.get("trace_id")
+    return None
```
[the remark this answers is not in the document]
mtoffl01: Oh okay that's some python magic

tests/test_config_consistency.py
```diff
@@ -662,13 +661,28 @@ def get_runtime_metrics(agent):
...
     for data in stdout.get_data():
         try:
             message = json.loads(data.get("message"))
         except json.JSONDecodeError:
             continue
-        if message.get("dd") and message.get(log_injection_fields[context.library.library]["message"]) == log_message:
+        # APMAPI-1199
+        if (
+            context.library == "nodejs"
+            and message.get(log_injection_fields[context.library.library]["message"]) == log_message
```
mabdinur, line 674: This condition is hard to understand. Can we leave comment to document why we need to to check log_injection_fields for the nodejs weblog app. It would be nice if we could avoid this but we shouldn't invest too much time into it if this is challenging.

cbeauchesne (commented):
AGTM for framework usage. I let @mabdinur do the review for the test code.

tests/test_config_consistency.py
```diff
@@ -662,13 +665,28 @@ def get_runtime_metrics(agent):
     return runtime_metrics_gauges, runtime_metrics_sketches
 
 
-# Parse the JSON-formatted log message from stdout and return the 'dd' object
+# Parses the JSON-formatted log message from stdout and returns it
+# To pass tests that use this function, ensure your library has an entry in log_injection_fields
 def parse_log_injection_message(log_message):
```
mabdinur, line 670:
```suggestion
def parse_log_injection_message(log_message):
    # Parses the JSON-formatted log message from stdout and returns it
    # To pass tests that use this function, ensure your library has an entry in log_injection_fields
```
nit: in python we usually put the comment in the test 😅

mabdinur (approved):

PR #4334, opened 2025-03-18
retry npm install when it fails

rochdev:
## Motivation

This is a common cause of flakiness, often with a random 500 error from npm or from an ECONNRESET. Rerunning it reduces the flakiness pretty much to 0 (or as close as we can realistically expect without rerunning many times).

## Changes

Retry `npm install` when it fails.

## Workflow

## Reviewer checklist

### DIFF
--- a/utils/_context/_scenarios/parametric.py
+++ b/utils/_context/_scenarios/parametric.py
@@ -402,7 +402,7 @@ COPY {nodejs_reldir}/*.js /usr/app/
 COPY {nodejs_reldir}/*.sh /usr/app/
 COPY {nodejs_reldir}/npm/* /usr/app/
 
-RUN npm install
+RUN npm install || npm install
 
 COPY {nodejs_reldir}/../install_ddtrace.sh binaries* /binaries/
 RUN /binaries/install_ddtrace.sh
--- a/utils/build/docker/nodejs/express4-typescript.Dockerfile
+++ b/utils/build/docker/nodejs/express4-typescript.Dockerfile
@@ -31,7 +31,7 @@ CMD ./app.sh
 
 COPY utils/build/docker/nodejs/install_ddtrace.sh binaries* /binaries/
 
-RUN npm install
+RUN npm install || npm install
 RUN /binaries/install_ddtrace.sh
 RUN npm run build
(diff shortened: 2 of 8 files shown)


cbeauchesne:
Would it be more clean to use `npm` config options : https://docs.npmjs.com/cli/v8/using-npm/config#fetch-retries

chatgpt told it can be achieved by `npm config set fetch-retries 2`

rochdev:
> chatgpt told it can be achieved by npm config set fetch-retries 2

Interestingly that is the default and it doesn't seem to work (or some other issue might be at play). We've been using a manual retry for a while in dd-trace and it seems to pretty much have resolved the issue entirely (short of a full-blown npm outage).

cbeauchesne:
> it doesn't seem to work (or some other issue might be at play).

Do you mean that it does **not** try a second time, or it may try, but it does not solve the issue? 

Because if it's the default, and if it actually tries twice, then the equivalent of `npm install || npm install` would be `npm config set fetch-retries 4`. 

I just checked, and from what I see in logs, it respects the config given (even +1) : 

```
> npm config set registry https://httpstat.us/500
> npm config set fetch-retries 10
> npm install axios --verbose

... logs

npm http fetch GET 500 https://httpstat.us/500/axios 315458ms attempt #11 (cache skip)
```

See the `#11` 

Well, whatever, if you're more confortable with the double call, I'm fine with that.

cbeauchesne (approved):

rochdev:
> Do you mean that it does not try a second time, or it may try, but it does not solve the issue?

I think it does try a second time, but it may do so with a strategy that doesn't quite work for us. I agree than the new code is technically equivalent to 4 retries, but since it's done differently, it's not quite equivalent. For example, in dd-trace-js we use yarn which doesn't retry out of the box, yet retrying manually (so effectively 2 retries total) fixed the issue entirely. This is also true of `npm view`. Basically, because our use case uses so many different flavours and versions of package managers, it seems to be a safer bet to retry manually even if that's not pretty, instead of relying on potentially incorrect or variable implementations.

rochdev:
We had another occurrence of the issue, going to merge this for now to see if it makes a difference. We can re-evaluate this if it ends up not fixing the issue.

PR #4401, opened 2025-03-27
[golang] fix(process-discovery): Go tracer reports `go` instead of `golang`

dmehala:
## Motivation
`test_library.lang` evaluates to `golang` for Go, which does not align with the language reported by the Go tracer. This fix resolves the issue by matching against `go` instead of `golang`.

## Reviewer checklist

* [x] If PR title starts with `[<language>]`, double-check that only `<language>` is impacted by the change

### DIFF
--- a/tests/parametric/test_process_discovery.py
+++ b/tests/parametric/test_process_discovery.py
@@ -4,7 +4,8 @@ import pytest
 import json
 import msgpack
 from jsonschema import validate as validation_jsonschema
-from utils import features, scenarios
+from utils import features, scenarios, context
+from utils._context.component_version import Version
 
 
 def find_dd_memfds(test_library, pid: int) -> list[str]:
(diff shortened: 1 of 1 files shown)


tests/parametric/test_process_discovery.py
```diff
@@ -63,7 +65,9 @@ def test_metadata_content(self, test_library, library_env):
             # assert tracer_metadata["hostname"]
             # TODO(@dmehala): how to get the version?
             # assert tracer_metadata["tracer_version"] ==
```
dmehala, line 67: Hi @cbeauchesne , do you know how could I retrieve the version of the tracer here?

dmehala:
On more thing @cbeauchesne , how can I enable the test on the `HEAD` of dd-trace-go?

cbeauchesne:
Jobs starting with `[System Tests (golang, dev)](https://github.com/DataDog/system-tests/actions/runs/14102713473/job/39503512686?pr=4401#logs)` is what's you're looking for : https://github.com/DataDog/system-tests/actions/runs/14102713473/job/39503512686?pr=4401

tests/parametric/test_process_discovery.py
cbeauchesne: `context.library.version`. Though, it may contains some difference, versions of dd-trace repo are very heterogeneous,  I had to do some trick to support all of them, meaning that the value you'll see in `context.library.version`  can differ from the one reported.

cbeauchesne: But at least, checking major/minor/patch should be good enough. Hacks are mostly on pre-release / built parts

PR #4410, opened 2025-03-27
config_consistency: clean up test reporting

mabdinur:
## Motivation

Ensure features report the correct status to the FPD

## Changes

- Enable XPassing tests
- Fix broken sampling test
- Mark irrelevant test cases as irrelevant instead of missing_feature/broken app

## Workflow

## Reviewer checklist

### DIFF
--- a/manifests/java.yml
+++ b/manifests/java.yml
@@ -1841,16 +1841,16 @@ tests/:
       '*': irrelevant (kafka endpoints are not implemented)
       spring-boot: v1.42.0
     Test_Config_LogInjection_128Bit_TraceId_Disabled:
-      '*': incomplete_test_app (weblog endpoint not implemented)
+      '*': irrelevant (weblog endpoint not implemented)
       'spring-boot-3-native': v1.48.0
     Test_Config_LogInjection_128Bit_TraceId_Enabled:
-      '*': incomplete_test_app (weblog endpoint not implemented)
+      '*': irrelevant (weblog endpoint not implemented)
       'spring-boot-3-native': v1.48.0
     Test_Config_LogInjection_Default:
-      '*': incomplete_test_app (weblog endpoint not implemented)
+      '*': irrelevant (weblog endpoint not implemented)
       'spring-boot-3-native': v1.48.0
     Test_Config_LogInjection_Enabled:
-      '*': incomplete_test_app (weblog endpoint not implemented)
+      '*': irrelevant (weblog endpoint not implemented)
       'spring-boot-3-native': v1.48.0
     Test_Config_ObfuscationQueryStringRegexp_Configured: v1.39.0
     Test_Config_ObfuscationQueryStringRegexp_Default: v1.39.0
--- a/manifests/nodejs.yml
+++ b/manifests/nodejs.yml
@@ -1011,33 +1011,33 @@ tests/:
     Test_Config_HttpServerErrorStatuses_FeatureFlagCustom: missing_feature
     Test_Config_IntegrationEnabled_False:
       '*': *ref_5_25_0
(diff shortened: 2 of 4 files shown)


manifests/java.yml
```diff
@@ -1841,16 +1841,16 @@ tests/:
       '*': irrelevant (kafka endpoints are not implemented)
       spring-boot: v1.42.0
     Test_Config_LogInjection_128Bit_TraceId_Disabled:
-      '*': incomplete_test_app (weblog endpoint not implemented)
+      '*': irrelevant (weblog endpoint not implemented)
```
mhlidd, line 1844: For the Java ones, these weblog apps don't support the endpoint `log/library`. Shouldn't it stay under `incomplete_test_app` then?

mabdinur: The issue here is that tests marked with `incomplete_test_app` show up as failures in the FPD (feature health views).

If we plan to update all weblog apps to have the `/log/library` endpoint then we should keep this as incomplete_test_app. If not, we should use `irrelevant` so that the FPD ignores the failing test cases.

mhlidd: Hmm that's fine then. Wondering what the motivation for having `incomplete_test_app` as failure is.

mhlidd (approved):
LGTM for java

manifests/java.yml
mabdinur: In the short term I just want the FPD to be green/yellow for config consistency. Long term we need a better way to surface incomplete_test_app issues in the FPD

mtoffl01: This interim change could make it hard to differentiate between true `irrelevant`s and `incomplete_test_app`s disguised as `irrelevant`s

mtoffl01: At the very least, I'd suggest leaving a consistent description next to `irrelevant`s that are actually `incomplete_test_app`s so that we can rediscover them in the future. Like standardize on "weblog endpoint not implemented" or "endpoint not implemented." Or, link a jira card that explains this.

mabdinur: True. The bigger issue is that weblog variants have diverged significantly, with many endpoints implemented in only one or two apps. A better approach would be to implement new `library/log` endpoint in every app, ensuring that irrelevant retains its intended meaning. I think we should also enforce this policy for new endpoints, where it makes sense (cc: @cbeauchesne). The real problem is the large number of endpoints and weblog apps we need to maintain, making maintenance a pain. This PR attempts to work around the issue, but I think we should focus on solving it directly. I'll move this PR to a draft state.


@colin-higgins It would be nice to have special treatment for test outcomes with the new `incomplete_test_apps` annotation (but not exactly sure what the experience should be)

cbeauchesne: > The issue here is that tests marked with incomplete_test_app show up as failures in the FPD (feature health views).

It is exactly what we want : if a feature can't be tested (whatever the reason) then nothing can't ensure it's health, so it can't be considered as healthy.

You can see with Colin how to have better view in FPD, but it;s a visualisation issue. The test declaration must follow the key rules, irrelevant is for feature that make no sense, or will never be reasonnably implemented.

In particular, if you use `irrelevant`, you'll defeat all the purpose of the test : 

* the test will never show up in FPD, making one believe the feature is ok, where it may not be ok
* if for some reason, someone correct it (either the feature, or the test app), the tes won';t be executed, and won't show up as xpass.

So, long story short, please keep  `incomplete_test_app`

mabdinur:
We should update the UI to better handle `incomplete_test_app`. We should not mark failing/missing endpoints with irrelevant.

PR #4455, opened 2025-04-02
Skip test_config_telemetry_completeness for .NET

bouwkast:
## Motivation

`dd-trace-dotnet` does not block on system test failures.
`dd-trace-dotnet` has had this same check in its own tests and blocks on that.

This removes a point of friction of having to merge multiple PRs prior to merging one in .NET.

When the file is automatically updated we can re-enable the test.

## Changes

Marking the test as `irrelevant` as `dd-trace-dotnet` already has it.

## Workflow

## Reviewer checklist


tests/test_telemetry.py
```diff
@@ -614,6 +614,10 @@ def test_app_started_product_info(self):
                     "appsec" in products
                 ), "Product information is not accurately reported by telemetry on app-started event"
 
+    @irrelevant(
+        library="dotnet",
+        reason="dd-trace-dotnet already has this, will re-enable when this automatically updates the dd-go files.",
```
cbeauchesne, line 619: If I'm not wrong,  the test in .NET is not exactly this one : the present test check that keys used are present in dd-go repo, while The test on dd-trace-dotnet repo does not fail if the key is missing on dd-go.

If i'm right, could you just keep the message `re-enable when this automatically updates the dd-go files.` ? I would not like to let someone think that this test could be also skipped for other repo for a bad reason.

cbeauchesne (changes requested):

tests/test_telemetry.py
bouwkast: Like so?

```suggestion
        reason="Re-enable when this automatically updates the dd-go files.",
```

bouwkast: But yes, the test in `dd-trace-dotnet` is a _copy_ of `dd-go`'s file unlike this one here

lucaspimentel: > is a copy of `dd-go`'s file

Is it really? I tried copying the file recently and it has lots of diffs. I think we just add individual setting in the copy in `dd-trace-dotnet` without pull the copy from `dd-go`.

bouwkast: A copy made at some other point in time in the past and rarely synced😄

cbeauchesne (approved):

cbeauchesne:
Rebased to fix issues

PR #4470, opened 2025-04-03
Introduce a way to add extra argument to library cmd on parametric scenario

cbeauchesne:
## Motivation

Allow to add arbitary argument to library container command in parametric scenario

## Changes

Add `library_extra_command_arguments` parameter. 

## Usage : 

```python
import pytest
from utils import scenarios


class Test_stuff:
    @scenarios.parametric
    @pytest.mark.parametrize(
        "library_extra_command_arguments",
        [
            ["--arg1"]
            ["--arg2=12"]
        ],
    )
    def test_stuff(self, test_agent, test_library):
        pass
```

## Workflow

## Reviewer checklist

### DIFF
--- a/utils/build/docker/java/parametric/run.sh
+++ b/utils/build/docker/java/parametric/run.sh
@@ -33,6 +33,7 @@ DISABLE_INTEGRATIONS=(-Ddd.integration.servlet-request-body.enabled=false \
 OPTIMIZATION_OPTIONS=(-XX:TieredStopAtLevel=1)
 
 # Start client application
+# shellcheck disable=SC2086
 java -Xmx128M -javaagent:"${DD_JAVA_AGENT}" \
   $ENABLE_OTEL_TRACING_API \
   "${ENABLE_CRASH_TRACKING[@]}" \
--- a/utils/build/docker/java/parametric/run.sh
+++ b/utils/build/docker/java/parametric/run.sh
@@ -39,4 +39,5 @@ java -Xmx128M -javaagent:"${DD_JAVA_AGENT}" \
   "${DISABLED_FEATURES[@]}" \
   "${DISABLE_INTEGRATIONS[@]}" \
   "${OPTIMIZATION_OPTIONS[@]}" \
+  ${SYSTEM_TESTS_EXTRA_COMMAND_ARGUMENTS:-} \
   -jar target/dd-trace-java-client-1.0.0.jar
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -195,6 +195,16 @@
             "python": "${workspaceFolder}/venv/bin/python"
         },
         {
+            "name": "Run PARAMETRIC scenario",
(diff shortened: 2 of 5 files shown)


mtoffl01 (approved):
(Not a blocker) Thanks for adding the `SYSTEM_TESTS_EXTRA_COMMAND_ARGUMENTS` to the java container script! But is there a good way to mark that the other libraries will have to do this as well, without us having to do it for all libs right now?

cbeauchesne:
I forgot to push a commit ...

utils/build/docker/java/parametric/run.sh
```diff
@@ -33,10 +33,12 @@ DISABLE_INTEGRATIONS=(-Ddd.integration.servlet-request-body.enabled=false \
...
 # Start client application
+# shellcheck disable=SC2086
 java -Xmx128M -javaagent:"${DD_JAVA_AGENT}" \
   $ENABLE_OTEL_TRACING_API \
   "${ENABLE_CRASH_TRACKING[@]}" \
   "${DISABLED_FEATURES[@]}" \
   "${DISABLE_INTEGRATIONS[@]}" \
   "${OPTIMIZATION_OPTIONS[@]}" \
+  ${SYSTEM_TESTS_EXTRA_COMMAND_ARGUMENTS:-} \
   -jar target/dd-trace-java-client-1.0.0.jar
```
mtoffl01, line 44:
```suggestion
  -jar target/dd-trace-java-client-1.0.0.jar \
  ${SYSTEM_TESTS_EXTRA_COMMAND_ARGUMENTS:-}
```

mtoffl01: Apparently only official JVM options (`-Xmx` etc)can come before the `-jar`, all other options e.g. application options must come after.

mtoffl01 (changes requested):

utils/build/docker/java/parametric/run.sh
mtoffl01: Ah. And I realized that the feature I'd like to test relies on JVM args, not application args, so we need to keep the command as you have it (ignore my suggestion above). As a user of the `library_extra_command_arguments` for the java test library, I cannot use `--arg1`, I need to use a JVM option. Perhaps this should be documented, although I don't know of a good generic way to do so, as it may be different for different languages.

cbeauchesne: you can definitly do something like : 

```python
from utils import context

# python condese

extra_args = ["--default-option"]

if context.library.name == "java":
    extra_args = ["--java-options=1"]
else:
    extra_args = ["--default-option=true"]

class Test_stuff:
    @scenarios.parametric
    @pytest.mark.parametrize(
        "library_extra_command_arguments",
        [
            extra_args,
        ],
    )
    def test_stuff(self, test_agent, test_library):
        pass
```

PR #4525, opened 2025-04-14
Small improvment on CI

cbeauchesne:
## Motivation

* faster CI time
* easier to relaunch when the distant artifact has been updated

### The job "load dev artifact" has been removed, it's now done in "compute workflow parameter"

* remove one job -> faster CI
* each lib will start once its dev is loaded
    *  `prod` don't have to wait that `dev` artifact is loaded
    *  `ruby`/`dev` can start, even if `python`/`dev` artifact is not yet loaded
* if an issue is fixed in a distant repo, `dev` can be restarted by restarting the corresponding  `compute workflow parameter`, without needing to restart everything.

### `system-tests.yml` new parameters, only meant to be used in system-tests CI (and nightlies)

* `_system_tests_dev_mode` : `true` is the workflow must load agent and library `dev` artifact
* `_system_tests_library_target_branch` : `str`, if the library dev artifact must use an arbitrary branch, instead of `main`/`master`. 

## Workflow

## Reviewer checklist


utils/scripts/compute-workflow-parameters.py
```diff
@@ -23,25 +23,36 @@ class CiData:
...
+        gitlab_environment: str,
     ):
         # this data struture is a dict where:
         #  the key is the workflow identifier
         #  the value is also a dict, where the key/value pair is the parameter name/value.
         self.data: dict[str, dict] = {}
         self.language = library
-        self.environment = ci_environment
+        self.gitlab_environment = gitlab_environment
+        self.ci_environment = (
```
robertomonteromiguel, line 42: Why not use the variable “ci_environment” instead of "gitlab_environment".  Finally I see that they will have the same value “dev” or “prod”.

utils/scripts/compute-workflow-parameters.py
```diff
@@ -218,15 +232,25 @@ def _get_workflow_map(scenario_names: list[str], scenario_group_names: list[str]
...
     # Misc
-    parser.add_argument("--ci-environment", type=str, help="Used internally in system-tests CI", default="custom")
+    parser.add_argument(
+        "--explicit-artifact-name", type=str, help="If an artifact name is explicitly provided", default=""
+    )
+    parser.add_argument(
+        "--system-tests-dev-mode", type=str, help="If an artifact name is explicitly provided", default="false"
+    )
+    parser.add_argument(
+        "--gitlab-environment", type=str, help="If an artifact name is explicitly provided", default="prod"
```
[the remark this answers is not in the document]
cbeauchesne: yep :)

utils/scripts/compute-workflow-parameters.py
cbeauchesne: yes, but IIUC, the way they are used is different : 

On github, the artifact is loaded before this process is called, so the value is the result of this logic : 

* if `system_tests_dev_mode == True` => `dev` (means `system-tests.yml` took care of pulling dev artifacts)
* else if no artifact provided => `prod` 
* else `custom` (arbitrary artifact)

Said differently, `system_tests_dev_mode`/`explicit_artifact_name` are inputs, `ci_environment` is an output.

On gitlab, `gitlab_environment` is both an input and an output, so it follows a different logic. Though, I also agree it's a little bit clunky. 

You think it would be preferable to keep it as `ci_environment`, and use the github logic if it's not specified ? : 

```python
if ci_environment is not None:
    self.ci_environment = ci_environment # explicitly specified
    # check that  `system_tests_dev_mode`/`explicit_artifact_name` are not set
else:
    self.ci_environment  = ... # github logic
```

robertomonteromiguel (approved):

PR #4564, opened 2025-04-25
test(debugger): excluding PII identifiers should work

tylfin:
## Motivation

This pull request enhances the PII redaction tests by introducing support for excluded identifiers. The changes ensure that specific identifiers, such as `cookie` and `sessionid`, are excluded from redaction and are validated accordingly. The most important updates include modifications to the test logic, the addition of new test scenarios, and updates to the configuration for excluded identifiers.

## Changes

* **Support for excluded identifiers in tests**:
  - Introduced the `EXCLUDED_IDENTIFIERS` list in `tests/debugger/test_debugger_pii.py` to specify identifiers that should not be redacted, such as `cookie` and `sessionid`.
  - Updated the `_assert` and `_validate_pii_keyword_redaction` methods to handle excluded identifiers, ensuring they are present in snapshots and not improperly redacted. [[1]](diffhunk://#diff-2cc81d31929c49f7bdc8fa6992d99f5d975abaea0843ee29909de7bf042d1728L132-R168) [[2]](diffhunk://#diff-2cc81d31929c49f7bdc8fa6992d99f5d975abaea0843ee29909de7bf042d1728R180-R194) [[3]](diffhunk://#diff-2cc81d31929c49f7bdc8fa6992d99f5d975abaea0843ee29909de7bf042d1728R204-R210)
* **New test scenarios for excluded identifiers**:
  - Added a new test case, `test_pii_redaction_excluded_identifiers`, to validate the behavior of excluded identifiers. This includes handling missing features and known bugs across different libraries.
* **Environment variable for excluded identifiers**:
  - Added `DD_DYNAMIC_INSTRUMENTATION_REDACTION_EXCLUDED_IDENTIFIERS` to the `DEBUGGER_PII_REDACTION` scenario specifying `cookie`, `_2fa`, and `sessionid`.

## Workflow

## Reviewer checklist


watson (commented):
If I understand the changes correctly, it looks like it's not checking the keys you've moved to `EXCLUDED_IDENTIFIERS` is being properly redacted in case there's no exclusions. That's technically a downgrade in the coverage of our tests. I would recommend that you concat the two arrays in the regular test to keep testing those keys are being properly redacted. Otherwise them not being redacted when excluded might just be because the tracer doesn't include them in the redaction list in the first place

tylfin:
@watson Yeah you're right, unfortunately combining the arrays in code won't work here because the exclusion is controlled at the environment-variable level. @shurivich is there a way to dynamically set an environment variable on a tracer in the setup-phase? I didn't think so...

tests/debugger/test_debugger_pii.py
```diff
@@ -215,11 +250,21 @@ def setup_pii_redaction_method_full(self):
     )
     @missing_feature(context.library == "nodejs", reason="Not yet implemented", force_skip=True)
     def test_pii_redaction_method_full(self):
-        self._assert(REDACTED_KEYS, REDACTED_TYPES)
+        self._assert(REDACTED_KEYS, REDACTED_TYPES, excluded_identifiers=None)
```
watson, line 253: @tylfin what I mean is this:

```suggestion
        self._assert(REDACTED_KEYS + EXCLUDED_IDENTIFIERS, REDACTED_TYPES, excluded_identifiers=None)
```

cbeauchesne:
> is there a way to dynamically set an environment variable on a tracer in the setup-phase?

You can't, for end-to-end scenario, when setup method are executed, the infra is already up & running.

Could you precise a little bit more what you'd like to achieve ?

tylfin:
@cbeauchesne We want to have the first two tests here with:

`DD_DYNAMIC_INSTRUMENTATION_REDACTION_EXCLUDED_IDENTIFIERS": ""`

and the last one with:

`DD_DYNAMIC_INSTRUMENTATION_REDACTION_EXCLUDED_IDENTIFIERS": "someValue1,someValue2,..."`

I could create another scenario but it's a bit heavy weight for one test

tylfin:
@watson It's a bit non-ideal, but I think we should move forward on this. Ideally the tracers are including the full PII list in unittests and this will be extra assurance for parity-sake

tests/debugger/test_debugger_pii.py
```diff
@@ -34,9 +34,9 @@
     "cipher",
     "clientid",
     "clientsecret",
+    "cookie",
     "connectionstring",
     "connectsid",
-    "cookie",
```
watson, line 39: I don't assume this was intentional?

utils/_context/_scenarios/__init__.py
```diff
@@ -625,10 +625,14 @@ class _Scenarios:
     tracing_config_nondefault_4 = EndToEndScenario(
         "TRACING_CONFIG_NONDEFAULT_4",
         weblog_env={
+            "DD_DYNAMIC_INSTRUMENTATION_ENABLED": "1",
```
watson, line 628: Just to keep it consistent with the other booleans below:

```suggestion
            "DD_DYNAMIC_INSTRUMENTATION_ENABLED": "true",
```

tests/debugger/test_debugger_pii.py
```diff
@@ -129,30 +134,39 @@ def _setup(self, *, line_probe=False):
...
+                fields = snapshot["captures"]["return"]["locals"]["pii"]["fields"]
+
+            # Check if fields that should be redacted are properly redacted
             for field_name in should_redact_field_names:
-                if line_probe:
-                    fields = snapshot["captures"]["lines"]["64"]["locals"]["pii"]["fields"]
-                else:
-                    fields = snapshot["captures"]["return"]["locals"]["pii"]["fields"]
+                if field_name in excluded_identifiers:
+                    continue
```
watson, line 169: You currently already remove the excluded identifiers from the `should_redact_field_names` list in the `test_pii_redaction_excluded_identifiers` before calling `_assert`, so there either shouldn't be a reason to have this check, or you should remove the line creating the `redacted_keys` array in `test_pii_redaction_excluded_identifiers`.

I think actually the best cause of action is to refactor the `_assert` method and remove the `redacted_keys` argument. Instead we should just inline the reference to `REDACTED_KEYS` where we need it. This argument is meant to represent the keys that the tracer defaults to redact and this is kind of a constant that shouldn't change between tests. So I think inlining it is perfectly fine in this case. Then you can live these two lines here to `continue` in case `field_name` is in the `excluded_identifiers` array.

tests/debugger/test_debugger_pii.py
```diff
@@ -165,6 +179,20 @@ def _validate_pii_keyword_redaction(self, should_redact_field_names, line_probe)
...
+                    check_field_name = field_name
+
+                if check_field_name in fields:
+                    excluded_found.append(field_name)
+
+                    # Excluded identifiers should have values
+                    if "value" not in fields[check_field_name]:
+                        improperly_redacted.append(field_name)
+
         error_message = ""
```
watson, line 196: Nit: The final error message might start with `. ` if `not_redacted` is empty. In these cases, I'd rather make `error_message` into an array and append to it inside each of the if-statement below and finally join it with `. ` and print it if it contains elements. But feel free to ignore as this isn't important

```suggestion
        error_messages = []
```

tests/debugger/test_debugger_pii.py
```diff
@@ -129,31 +134,37 @@ def _setup(self, *, line_probe=False):
         self.wait_for_all_probes(statuses=["EMITTING"])
 
     ############ assert ############
-    def _assert(self, redacted_keys, redacted_types, *, line_probe=False):
+    def _assert(self, redacted_keys, redacted_types, *, line_probe=False, excluded_identifiers=None):
```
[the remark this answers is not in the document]
tylfin: How would you differentiate between tests that expect the excluded identifiers to be redacted (e.g. debugger scenario) and the test that expects the excluded identifiers to be excluded and available as values?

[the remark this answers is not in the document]
watson: When iterating over `REDACTED_KEYS` in `_validate_pii_keyword_redaction`, I'd simply check if the key is present in the `excluded_identifiers` array and if so, not append it to `not_redacted`. So something like this (handwritten):

```diff
- if "value" in fields[check_field_name]:
+ if "value" in fields[check_field_name] and not field_name in excluded_identifiers:
      not_redacted.append(field_name)
+ elif field_name in excluded_identifiers and not "value" in fields[check_field_name]:
+     improperly_redacted.append(field_name)
```

Then I think you could drop the entire `for field_name in excluded_identifiers` loop as well, right?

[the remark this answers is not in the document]
tylfin: Hmm, okay tell me if you think that's more clear. It pulls some of the logic into _assert which might be confusing, but I did eliminate a few temp vars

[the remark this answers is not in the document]
watson: Update: We had a discussion on Slack and updated accordingly. Everything looks good now - just a small linting error 👍

watson (approved):

PR #4606, opened 2025-05-07
[APPSEC-57504] Enable session tracking and fingerprinting

Strech:
## Motivation

This is the last bits for rails session tracking and general fingerprinting

## Changes

An important change is that since runtime activation is not working in Ruby I have to change it to Remote Config and AppSec enabled.

### DIFF
--- a/manifests/ruby.yml
+++ b/manifests/ruby.yml
@@ -363,11 +363,11 @@ tests/:
     test_asm_standalone.py:
       Test_AppSecStandalone_NotEnabled: v2.13.0
       Test_AppSecStandalone_UpstreamPropagation: irrelevant
-      Test_AppSecStandalone_UpstreamPropagation_V2: v2.12.3-dev
+      Test_AppSecStandalone_UpstreamPropagation_V2: v2.13.0
       Test_IastStandalone_UpstreamPropagation: missing_feature
       Test_IastStandalone_UpstreamPropagation_V2: missing_feature
       Test_SCAStandalone_Telemetry: irrelevant
-      Test_SCAStandalone_Telemetry_V2: v2.12.3-dev
+      Test_SCAStandalone_Telemetry_V2: v2.13.0
       Test_UserEventsStandalone: missing_feature
     test_automated_login_events.py:
       Test_Login_Events: irrelevant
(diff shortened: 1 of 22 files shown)


uurien (approved):

manifests/ruby.yml
```diff
@@ -483,14 +494,36 @@ tests/:
...
+        rack: irrelevant (We do not instrument authentication frameworks that work with sinatra or rack)
+        sinatra14: irrelevant (We do not instrument authentication frameworks that work with sinatra or rack)
+        sinatra20: irrelevant (We do not instrument authentication frameworks that work with sinatra or rack)
+        sinatra21: irrelevant (We do not instrument authentication frameworks that work with sinatra or rack)
+        sinatra22: irrelevant (We do not instrument authentication frameworks that work with sinatra or rack)
+        sinatra30: irrelevant (We do not instrument authentication frameworks that work with sinatra or rack)
+        sinatra31: irrelevant (We do not instrument authentication frameworks that work with sinatra or rack)
+        sinatra32: irrelevant (We do not instrument authentication frameworks that work with sinatra or rack)
+        sinatra40: irrelevant (We do not instrument authentication frameworks that work with sinatra or rack)
+        uds-sinatra: irrelevant (We do not instrument authentication frameworks that work with sinatra or rack)
```
uurien, line 513: If we do not instrument, I understand that we are not supporting it yet, so it should be `missing_feature`, because eventually it could be supported in the future.

Irrelevant is for tests that doesn't make sense, for example, some deleted feature, things that aren't implemented because they are not compatible with the framework.

Strech: Ok, I will mark it as missing feature than

simon-id (approved):
LGTM, but it's odd i swear this was changed before already, maybe in a similar test

robertomonteromiguel (changes requested):
are the CI failures related with this PR?

Strech:
@robertomonteromiguel They will pass as soon as I rebase the code

robertomonteromiguel (changes requested):
still failing

Strech:
@robertomonteromiguel Thanks, but I think it's unnecessary to request changes every time this branch is red. We have a green build policy and I can't merge this PR without passing all tests.

Just some unnecessary friction created 😞

Strech:
Instead, I've created another PR to avoid breaking current `dd-trace-rb` builds. It's exactly the same with slight difference to correctly handle master-branch features (unreleased yet).

Could you @robertomonteromiguel take a look and give your 👍🏼 https://github.com/DataDog/system-tests/pull/4635 @y9v

PR #4862, opened 2025-06-27
DYNIS-22 check scm metadata in ddtags in DI

p-datadog:
## Workflow

## Reviewer checklist

### DIFF
--- a/tests/debugger/test_debugger_probe_snapshot.py
+++ b/tests/debugger/test_debugger_probe_snapshot.py
@@ -123,6 +123,66 @@ class Test_Debugger_Method_Probe_Snaphots(BaseDebuggerProbeSnaphotTest):
         self._validate_snapshots()
 
 
+@features.debugger_method_probe
+@scenarios.debugger_probes_snapshot_with_scm
+@missing_feature(context.library == "php", reason="Not yet implemented", force_skip=True)
+@missing_feature(context.library == "ruby", reason="Not yet implemented", force_skip=True)
+@missing_feature(context.library == "nodejs", reason="Not yet implemented", force_skip=True)
+class Test_Debugger_Method_Probe_Snaphots_With_SCM(BaseDebuggerProbeSnaphotTest):
+    """Tests for method-level probe snapshots"""
+
+    ### log probe ###
+    def setup_log_method_snapshot(self):
+        self._setup("probe_snapshot_log_method", "/debugger/log", "log", lines=None)
+
+    @missing_feature(context.library == "nodejs", reason="Not yet implemented")
+    def test_log_method_snapshot(self):
+        self._assert()
+        self._validate_snapshots()
+
+    ### span probe ###
+    def setup_span_method_snapshot(self):
+        self._setup("probe_snapshot_span_method", "/debugger/span", "span", lines=None)
+
+    def test_span_method_snapshot(self):
+        self._assert()
+        self._validate_spans()
+
+    ### span decoration probe ###
+    def setup_span_decoration_method_snapshot(self):
+        self._setup(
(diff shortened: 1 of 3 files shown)


tests/debugger/test_debugger_probe_snapshot.py
```diff
@@ -123,6 +123,66 @@ def test_mix_snapshot(self):
         self._validate_snapshots()
 
 
+@features.debugger_method_probe
+@scenarios.debugger_probes_snapshot_with_scm
+@missing_feature(context.library == "php", reason="Not yet implemented", force_skip=True)
+@missing_feature(context.library == "ruby", reason="Not yet implemented", force_skip=True)
+@missing_feature(context.library == "nodejs", reason="Not yet implemented", force_skip=True)
```
cbeauchesne, line 130: IIUC, you use `force_skip` because the test is very long to execute? 

If yes, do you mind adding a small comment explaining this ? It'll help limiting  the copy-paste effect.

 Same for other `force_skip` usage.

utils/_context/_scenarios/__init__.py
```diff
@@ -688,6 +688,16 @@ class _Scenarios:
         doc="Test scenario for checking if method probe statuses can be successfully 'RECEIVED' and 'INSTALLED'",
     )
 
+    debugger_probes_status_with_scm = DebuggerScenario(
```
cbeauchesne, line 691: Could you check if another scenario could host your tests, maybe one of the `DEBUGGER_` scenarios.

There are two good reason for doing so : 

* performance : scenarios are the bottleneck of performance (which may not be true in your situation, as some setup functions are terribly slow in debugger test)
* testing : we are more likely to catch interactions bugs in functional testing when we have complex scenarios

cbeauchesne (changes requested):

tests/debugger/test_debugger_probe_snapshot.py
p-datadog: Don't know, I copied whatever was specified for the non-SCM snapshot tests.

utils/_context/_scenarios/__init__.py
p-datadog: I could make the "regular" snapshot tests have the git metadata, however in this case we would lose test coverage when git metadata is not specified.

p-datadog: Yes, I can move the tests added in this PR to the existing snapshot scenario, at which point there will no longer be test coverage for sending snapshots without git tags. Is this what you are requesting?

tests/debugger/test_debugger_probe_snapshot.py
cbeauchesne: My request is about tests introduced in this PR, not others.

If you prefer not making any decision here, can you comment this point in the code, otherwise, future reader may think it's a purposeful decision, where it's not.

cbeauchesne: Tested on zoom with Oleg, those new tests are taking a while (4mn) , so `force_skip` is definitly worthy here.

utils/_context/_scenarios/__init__.py
p-datadog: There is no existing scenario.

cbeauchesne (approved):
From framework usage, all good.

cbeauchesne:
This [failure](https://github.com/DataDog/system-tests/actions/runs/16323880267/job/46108914632?pr=4862#step:80:109) may be caused because you also need to declare the class as `missing_feature` for `cpp_nginx`

y9v (approved):

PR #4863, opened 2025-06-27
Run bundle with only Gemfile & Gemfile.lock to get most gem dependenc…

p-datadog:
…ies into base docker image layers

## Motivation

Speed up image building: currently any change in dd-trace-rb test applications causes gem dependencies to be reinstalled, with this PR dependencies should only be reinstalled if gemfiles change

## Changes

Add a step to dockerfile to copy gemfile + gemfile.lock then remove references to dd-trace-rb (which are added later by the script that configures tracer) then run `bundle`. This is in addition to the existing `bundle` run which is done after gemfile is finalized by the tracer configuration script.

## Workflow

## Reviewer checklist

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -65,9 +65,6 @@ compute_pipeline:
       if [ -n "$SYSTEM_TESTS_SCENARIOS" ] || [ -n "$SYSTEM_TESTS_SCENARIOS_GROUPS" ]; then
         cp *.yml "$CI_PROJECT_DIR"
       fi
-  rules:
-    - if: '$SCHEDULED_JOB == ""'
-    - if: '$SCHEDULED_JOB == null'
   artifacts:
     paths:
       - nodejs_ssi_gitlab_pipeline.yml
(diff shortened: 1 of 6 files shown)


marcotc (approved):

cbeauchesne (approved):

Strech (changes requested):
Gems could be `ddtrace` and `datadog`, also we don't need development and test dependencies

p-datadog:
@Strech Removal of any dependencies is actually not desirable. I delete ddtrace because in some rails apps that resolves to a truly ancient version (not even 1.x but something like 1.0.x) which doesn't work. There are no rails apps that specify datadog presently, but if they did, and that resolved to the current version of datadog, it would be preferable to install that dependency also into the earlier docker layer such that the second `bundle install` wouldn't end up installing anything.

Strech:
> If datadog does not change version number (e.g. when testing against released version of datadog rather than the local development tree), the entire set of dependencies is stored in the pre-install layer

Yes and it's good, should we do it for all weblogs that way then?

> Since some version of datadog gets installed, it may install some or all of datadog prerequisites in the pre-install layer

This is questionable because different version require different dependencies versions, i.e there is no guarantee on having it not install them again (it depends) 🤷🏼

Strech (approved):
My comments now only non-blocking 🚀

PR #5014, opened 2025-08-01
Add test for incorrect trace+span sampling with propagated sampling decision

LotharSee:
## Motivation

I started activating some remote rules on Datadog account for Go services, and it resulted in the incorrect flagging of `ingestion_reason:remote_rule` for some traces which should be `ingestion_reason:rum`.
After digging more, I realized trace rule were applied when they shouldn't.

The added test here isn't exhaustive but catches exactly the issue I'm facing on production. I'm not sure it is the right file to put it (since it mixes spans sampling and trace sampling) but hopefully, this isn't mattering too much.

## Changes

Add test `test_single_rule_with_head_and_rule_trace_sampling_keep_019` and `test_single_rule_with_head_and_rule_trace_sampling_drop_020` to ensure trace sampling and span sampling applies properly when there is already a parent sampling decision.

Note: CI is currently red in Python because of `TestServiceActivationEnvVarMetric.test_service_activation_metric`, I'll rebase whenever it gets fixed on `main`.

### DIFF
--- a/tests/parametric/test_span_sampling.py
+++ b/tests/parametric/test_span_sampling.py
@@ -862,6 +862,7 @@ class Test_Span_Sampling:
     @bug(context.library == "php", reason="APMAPI-1545")
     @bug(context.library == "ruby", reason="APMAPI-1545")
     @bug(context.library == "python", reason="APMAPI-1545")
+    @bug(context.library == "cpp", reason="APMAPI-1545")
     @pytest.mark.parametrize(
         "library_env",
         [
@@ -935,6 +936,7 @@ class Test_Span_Sampling:
     @bug(context.library == "golang", reason="APMAPI-1545")
     @bug(context.library == "php", reason="APMAPI-1545")
     @bug(context.library == "python", reason="APMAPI-1545")
+    @bug(context.library == "cpp", reason="APMAPI-1545")
     @pytest.mark.parametrize(
         "library_env",
         [


tests/parametric/test_span_sampling.py
```diff
@@ -753,3 +753,78 @@ def test_entire_trace_dropped_when_dropping_policy_is_active018(self, test_agent
         traces = test_agent.wait_for_num_traces(0, clear=True)
 
         assert len(traces) == 0
+
+
+    @bug(context.library == "cpp", reason="APMAPI-1545")
```
LotharSee, line 758: Created a ticket for this issue, hopefully at the right place with the right fields
https://datadoghq.atlassian.net/browse/APMAPI-1545

tests/parametric/test_span_sampling.py
```diff
@@ -753,3 +753,78 @@ def test_entire_trace_dropped_when_dropping_policy_is_active018(self, test_agent
...
+        "library_env",
+        [
+            {
+                "DD_TRACE_PROPAGATION_STYLE": "datadog",
+                "DD_SPAN_SAMPLING_RULES": json.dumps([{"name": "web.request", "sample_rate": 1.0}]),
+                "DD_TRACE_SAMPLING_RULES": json.dumps([{"name": "web.request", "sample_rate": 1.0}]),
+            }
+        ],
+    )
+    def test_single_rule_with_head_and_rule_trace_sampling_019(self, test_agent, test_library):
```
LotharSee, line 774: All tests seem to finish with a number, not sure why, so I did so too :noideawhatiamdoing:

genesor (approved):

genesor (approved):

PR #5015, opened 2025-08-01
Add API Security endpoint discovery tests

jandro996:
## Motivation

https://docs.google.com/document/d/1txwuurIiSUWjYX7Xa0let7e49XKW2uhm1djgqjl_gL0/edit?tab=t.0 

## Changes

Test that all payloads match the defined schema
Test the mandatory fields for each endpoint
One test per optional endpoint field
Test is_first flag

## Workflow

## Reviewer checklist

### DIFF
--- a/tests/appsec/api_security/test_endpoint_discovery.py
+++ b/tests/appsec/api_security/test_endpoint_discovery.py
@@ -1,0 +1,26 @@
+from utils import interfaces, rfc, scenarios, weblog, features
+
+from utils.telemetry import validate_app_endpoints_schema
+
+
+@rfc("https://docs.google.com/document/d/1txwuurIiSUWjYX7Xa0let7e49XKW2uhm1djgqjl_gL0/edit?tab=t.0")
+@scenarios.appsec_api_security
+@features.api_security_endpoint_discovery
+class Test_Endpoint_Discovery:
+    def setup_endpoint_discovery(self):
+        """Setup for endpoint discovery tests."""
+        self.request = weblog.get("/")
+
+    def test_endpoint_discovery(self):
+        """Test for endpoint discovery in API security."""
+        validate_app_endpoints_schema()
+
+        for data in interfaces.library.get_telemetry_data():
+            content = data["request"]["content"]
+            if content.get("request_type") != "app-endpoints":
+                continue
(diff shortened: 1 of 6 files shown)


tests/appsec/api_security/test_endpoint_discovery.py
```diff
@@ -0,0 +1,245 @@
...
+@features.api_security_endpoint_discovery
+class Test_Endpoint_Discovery:
+    _main_setup_done: bool = False
+
+    def main_setup(self):
+        """Setup (once) for endpoint discovery tests."""
+        if Test_Endpoint_Discovery._main_setup_done:
+            return
+        for _ in range(10):
+            weblog.get("/")
```
IlyasShabi, line 17: why we need to do some calls?

tests/appsec/api_security/test_endpoint_discovery.py
```diff
@@ -0,0 +1,245 @@
...
+            endpoints.extend(payload.get("endpoints", []))
+
+        assert endpoints, "No endpoints discovered"
+        return endpoints
+
+    def setup_endpoint_discovery(self):
+        """Setup for endpoint discovery test."""
+        self.main_setup()
+
+    def test_endpoint_discovery(self):
```
IlyasShabi, line 49: I think it's better to check endpoints returned by telemetry to test some edge cases like GET/HEAD and *

christophe-papazian: I'm not sure to understand what you mean. We are already checking endpoints returned by telemetry, but as the tests are defined for all weblogs, we can't make assumptions on which method are allowed here.
Maybe we could add an endpoint to the weblog specification that only accept GET to check for this particular case? (This could cause problem to frameworks that accept all methods and can't restrict them, if there are any)

IlyasShabi: IMO we should test some edge cases like wildcard routes `*` mentioned in the RFC for endpoints that accept all methods if it's available for all languages.

> Maybe we could add an endpoint to the weblog specification that only accept GET to check for this particular case?

Yes we can do this also

IlyasShabi (approved):

PR #5294, opened 2025-09-18
[java] Activating xpass tests for java

nccatoni:
## Motivation

Some tests are currently passing but not yet activated. This leaves related features unprotected against regressions.

## Changes

Update manifest file to enable xpass tests.
Note: the updates are generated automatically this may lead to some yaml formatting modifications without any test activation.

## Workflow

## Reviewer checklist

### DIFF
--- a/manifests/java.yml
+++ b/manifests/java.yml
@@ -81,7 +81,7 @@ tests/:
           akka-http: bug (APPSEC-56888)
           play: v1.51.0-SNAPSHOT
           spring-boot-3-native: missing_feature (GraalVM. Tracing support only)
-          spring-boot-payara: bug (APPSEC-57921)
+          spring-boot-payara: v1.54.0-SNAPSHOT+c2aec92b86
           vertx3: v1.51.0-SNAPSHOT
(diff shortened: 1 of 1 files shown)


jandro996:
> **❔ question:** ‏Is there a reason to use a specific build? Did all those feature start passing from 1.54 dev? Or from earlier version? Is that common to use dev version rather than final / stable release as feature can changed / be reverted before we do the release?
> 
> Sorry, that's a lot of question and I'm not that familiar with the way it's supposed to be 😓

From what I know, it’s generally accepted to tag system-tests with snapshot versions, we’ve done it before and it really helps reduce the number of PRs needed to activate support in tracers.

However, I’m not fully sure about how specific we should be with the snapshot build number. It seems risky to rely on an exact snapshot unless strictly necessary.

Also, I think we should tag the version at which the test actually passes, not just the one suggested by the easy win, unless they’re the same, of course

jandro996 (approved):
There are some conflicts but after solve but I'm ok with the changes with you explanation
Thanks!

PR #5298, opened 2025-09-18
[nodejs] Activating xpass tests for nodejs

nccatoni:
## Motivation

Some tests are currently passing but not yet activated. This leaves related features unprotected against regressions.

## Changes

Update manifest file to enable xpass tests.
Note: the updates are generated automatically this may lead to some yaml formatting modifications without any test activation.

## Workflow

## Reviewer checklist

### DIFF
--- a/manifests/nodejs.yml
+++ b/manifests/nodejs.yml
@@ -933,7 +933,7 @@ tests/:
       Test_ConfigurationVariables_New_Obfuscation: *ref_5_59_0
     test_customconf.py:
       Test_ConfRuleSet: v2.0.0
-      Test_CorruptedRules_Telemetry: *ref_5_44_0
+      Test_CorruptedRules_Telemetry: bug (APPSEC-59077)
       Test_MissingRules: v2.0.0
       Test_NoLimitOnWafRules: v2.4.0
     test_event_tracking.py:
@@ -1048,7 +1048,7 @@ tests/:
       Test_AsmDdMultiConfiguration: *ref_5_59_0
       Test_BlockingActionChangesWithRemoteConfig: *ref_4_1_0
       Test_Empty_Config: *ref_3_19_0
-      Test_Invalid_Config: *ref_5_66_0
+      Test_Invalid_Config: bug (APPSEC-59077)
       Test_Multiple_Actions: *ref_4_1_0
       Test_Unknown_Action: *ref_4_1_0
       Test_UpdateRuleFileWithRemoteConfig: *ref_3_19_0


manifests/nodejs.yml
```diff
@@ -42,9 +42,11 @@ refs:
   - &ref_5_25_0 '>=5.25.0 || ^4.49.0'
   - &ref_5_26_0 '>=5.26.0 || ^4.50.0'
   - &ref_5_27_0 '>=5.27.0 || ^4.51.0'
-  - &ref_5_29_0 '>=5.29.0 || ^4.53.0'  # express 5 support
+  # express 5 support
+  - &ref_5_29_0 '>=5.29.0 || ^4.53.0'
```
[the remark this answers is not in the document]
nccatoni: Unfortunately, from what I found, there is only one widely used YAML parser in Python that preserves comments. And even though it is supposed to also work with references, I found that it sometimes moves end-of-line comments to the next line when using references.

manifests/nodejs.yml
```diff
@@ -1304,11 +1316,11 @@ tests/:
       Test_Parametric_DDSpan_Set_Resource: incomplete_test_app (set_resource endpoint is not implemented)
       Test_Parametric_DDSpan_Start: bug (APMAPI-778)  # The resource name of the child span is overidden by the parent span.
       Test_Parametric_DDTrace_Baggage: incomplete_test_app (baggage endpoints are not implemented)
-      Test_Parametric_DDTrace_Crash: incomplete_test_app (crash endpoint is not implemented)
+      Test_Parametric_DDTrace_Crash: v5.67.0
```
watson, line 1319: In the Node.js tracer we use a different way to refer to the versions rather than the actual version number, so this technically should be `*ref_5_67_0` instead of `v5.67.0`. But I'm sure that will complicate the automated script (also since `ref_5_67_0` currently doesn't exists above, so it should be added in the same PR).

There's no different between `ref_5_67_0` and `v5.67.0` currently because we currently only have one active release line, but once we get two again (which will happen very soon), this will become an issue.

simon-id: I would say it might be a silver lining actually. Consider this: the bot cannot create the alias easily, thus it's setting a raw vX.X.X version. But this version is also actually not guaranteed to be correct, since it's only taking the latest version, not the actual version where the feature was released. This bot is a failsafe, not something we should rely on. The teams should activate system tests with the verified released numbers for all branches etc. But since the bot cannot do that, we will be able to quickly look at the manifest, and immediately spot places where the bot did a "best-effort" job, that we should either double check and correct, by just looking for those vX.X.X instead of our usual refs. Does that make sense? it's a bit difficult to explain

PR #5334, opened 2025-09-24
Fix easywins for Java in test_baggage.py 

cecile75:
## Motivation
Have a better visibility of the actual status of the tests (easy wins)

## Changes
Acknowledge the tests that pass on the last Java tracer version in test_baggage.py for the different java weblog variants.

## Workflow

### DIFF
--- a/manifests/java.yml
+++ b/manifests/java.yml
@@ -2142,46 +2142,32 @@ tests/:
       Test_NotReleased: missing_feature
   test_baggage.py:
     Test_Baggage_Headers_Basic:
-      '*': incomplete_test_app (endpoint does not support fetching headers injected to downstream call)
+      '*': v1.53.0
+      akka-http: incomplete_test_app (endpoint does not support fetching headers injected to downstream call)
+      jersey-grizzly2: incomplete_test_app (endpoint does not support fetching headers injected to downstream call)
+      play: incomplete_test_app (endpoint does not support fetching headers injected to downstream call)
+      ratpack: incomplete_test_app (endpoint does not support fetching headers injected to downstream call)
+      resteasy-netty3: incomplete_test_app (endpoint does not support fetching headers injected to downstream call)
       spring-boot: v1.50.0
-      spring-boot-jetty: v1.53.0
-      spring-boot-openliberty: v1.53.0
-      spring-boot-payara: v1.53.0
-      spring-boot-undertow: v1.53.0
-      spring-boot-wildfly: v1.53.0
-      uds-spring-boot: v1.53.0
-      vertx3: v1.53.0
-      vertx4: v1.53.0
+      spring-boot-3-native: incomplete_test_app (endpoint does not support fetching headers injected to downstream call)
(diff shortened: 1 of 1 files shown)


manifests/java.yml
```diff
@@ -2144,21 +2144,74 @@ tests/:
...
+      spring-boot-undertow: v1.53.0
+      spring-boot-wildfly: v1.53.0
+      uds-spring-boot: v1.53.0
+      vertx3: v1.53.0
+      vertx4: v1.53.0
     Test_Baggage_Headers_Malformed:
       '*': incomplete_test_app (endpoint does not support fetching headers injected to downstream call)
+      akka-http: v1.53.0
+      play: v1.53.0
+      ratpack: v1.53.0
```
mhlidd, line 2159: Out of curiosity, why would `akka-http`, `play`, or `ratpack` work for the malformed baggage test but not the basic one? Given that baggage is working end-to-end in `spring-boot`, I would expect it to work for all cases in other libraries (given the weblog app is configured to accept injected headers).

manifests/java.yml
```diff
@@ -2144,21 +2144,74 @@ tests/:
     Test_Baggage_Headers_Basic:
       '*': incomplete_test_app (endpoint does not support fetching headers injected to downstream call)
```
mhlidd, line 2145: Does it make more sense to mark the `*` generic case as passing with `v1.53.0` and mark the few weblog variants that are failing as `incomplete_test_app` instead?

cbeauchesne (approved):

manifests/java.yml
cecile75: Yes, that would be better

manifests/java.yml
cecile75: You're right, the malformed baggage tests pass because the tests check that there is no baggage headers but in the basic test, there should be a baggage header and there is not. Basically there are never any baggage headers and in this current state, having the malformed baggage tests pass, doesn't really mean anything..

manifests/java.yml
```diff
@@ -2142,23 +2142,58 @@ tests/:
...
   test_baggage.py:
     Test_Baggage_Headers_Basic:
-      '*': incomplete_test_app (endpoint does not support fetching headers injected to downstream call)
+      '*': v1.53.0
+      akka-http: incomplete_test_app (endpoint does not support fetching headers injected to downstream call)
+      jersey-grizzly2: incomplete_test_app (endpoint does not support fetching headers injected to downstream call)
+      play: incomplete_test_app (endpoint does not support fetching headers injected to downstream call)
+      ratpack: incomplete_test_app (endpoint does not support fetching headers injected to downstream call)
+      resteasy-netty3: incomplete_test_app (endpoint does not support fetching headers injected to downstream call)
       spring-boot: v1.50.0
```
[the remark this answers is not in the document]
cecile75: We will lost the "passing since v1.50.0", that's why I keep it apart

PR #5366, opened 2025-09-29
[HandsOffConfig] Activate default tests and one telemetry test

anna-git:
## Motivation

Hands-off config has been merged and is now fully supported in .NET. 
Activation of 2 steps here:
- handsoff config main default test
- telemetry test (but not config-id) is reported

## Changes
Adapt weblog, adding some new configuration properties among which is profiler enabled


mtoffl01 (approved):
Modifications to manifests and tests/parametric LGTM.

PR #6421, opened 2026-03-03
[PHP] Enable stripe tests on PHP

estringana:
###  Enable Stripe automated payment event tests for PHP

  Adds the Stripe PHP SDK to the PHP weblog containers and implements the three endpoints required by tests/appsec/test_automated_payment_events.py, enabling the test suite for PHP.

### Changes

  Weblog endpoints (new PHP files):
  - stripe_create_checkout_session.php — POST /stripe/create_checkout_session
  - stripe_create_payment_intent.php — POST /stripe/create_payment_intent
  - stripe_webhook.php — POST /stripe/webhook

  All three route through a fake Stripe API key pointed at internal_server:8089 (the mocked backend).

###   Build infrastructure:
  - composer.json / composer.gte8.2.json — adds stripe/stripe-php:^10.0 as a dependency
  - apache-mod/build.sh — installs the Stripe SDK before ddtrace, because ddtrace injection into ZTS builds causes composer to segfault at image-build time
  - install_ddtrace.sh — re-runs composer require stripe/stripe-php after ddtrace installation (for non-apache builds), installing unzip as a prerequisite
  - php-fpm/build.sh — adds unzip to the base package install (required by composer)
  - rewrite-rules.conf — adds Apache rewrite rules for the three /stripe/* routes

### DIFF
--- a/utils/build/docker/php/common/install_ddtrace.sh
+++ b/utils/build/docker/php/common/install_ddtrace.sh
@@ -170,12 +170,16 @@ if [[ $IS_APACHE -eq 1 ]]; then
 fi
 
 # Install stripe SDK if not already present (base images may predate this dependency).
-# Skip when open-telemetry is present: those PHP >=8.1 base images have extra packages
-# installed via `composer require` that are not in the committed composer files, and
-# running `composer require` again would incorrectly remove them.
+# Use the correct composer file for the PHP version so we don't orphan open-telemetry
+# packages that were added to PHP >=8.2 base images via composer.gte8.2.json.
 if command -v composer &>/dev/null \
     && [ -f /var/www/html/composer.json ] \
-    && ! [ -d /var/www/html/vendor/stripe ] \
-    && ! [ -d /var/www/html/vendor/open-telemetry ]; then
-  cd /var/www/html && composer require stripe/stripe-php "^10.0" --no-interaction --ignore-platform-req=ext-mbstring || true
+    && ! [ -d /var/www/html/vendor/stripe ]; then
+  PHP_MAJOR=$(php -r 'echo PHP_MAJOR_VERSION;')
+  PHP_MINOR=$(php -r 'echo PHP_MINOR_VERSION;')
+  STRIPE_COMPOSER_FILE=composer.json
+  if [ "$PHP_MAJOR" -gt 8 ] || { [ "$PHP_MAJOR" -eq 8 ] && [ "$PHP_MINOR" -ge 2 ]; }; then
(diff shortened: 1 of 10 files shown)


utils/build/docker/php/common/install_ddtrace.sh
```diff
@@ -4,6 +4,16 @@ set -eux
 
 IS_APACHE=${1:-0}
 
+cd /var/www/html
```
estringana, line 11: I had to add this here as temporary. Right now adding the package zip is done on the base image. The same happens with composer dependencies, as stripe. Once the images are updated, then this should be removed

estringana:
@cbeauchesne can you please review and give me advise on how to go around that temporary `composer require` becuase of dependencies being resolve on base image?

utils/build/docker/php/common/install_ddtrace.sh
```diff
@@ -4,6 +4,24 @@ set -eux
...
+export COMPOSER=composer.json
+if [ "$(printf '%s\n' "$PHP_VERSION" "8.2" | sort -V | head -n1)" = "8.2" ]; then
+	export COMPOSER=composer.gte8.2.json
+fi
+if [ -f "$COMPOSER" ] && grep -Fq 'stripe/stripe-php' "$COMPOSER"; then
+  apt-get update -qq && apt-get install -y --no-install-recommends unzip
+  if [[ "$(printf '%s\n' "$PHP_VERSION" "8.1" | sort -V | head -n1)" = "8.1" ]]; then
```
cataphract, line 18: Like the bot said, this assumes without checking that PHP_VERSION is defined, with confusing errors o/wise. Better to error if PHP_VERSION is not defined, or default it to sth like `php -r "echo PHP_MAJOR_VERSION.'.'.PHP_MINOR_VERSION;`

cataphract (approved):

nccatoni (approved):

utils/build/docker/php/common/install_ddtrace.sh
estringana: Sorted

utils/build/docker/php/common/install_ddtrace.sh
```diff
@@ -4,6 +4,24 @@ set -eux
 
 IS_APACHE=${1:-0}
 
+cd /var/www/html
+export COMPOSER=composer.json
+if [ "$(printf '%s\n' "$PHP_VERSION" "8.2" | sort -V | head -n1)" = "8.2" ]; then
```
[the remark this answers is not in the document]
estringana: 👍

PR #6648, opened 2026-03-31
Add css obfuscation test [APMSP-2764]

Eldolfin:
## TODO
- [x] add a missing obfuscation_version test
- [x] add a obfuscation_version = 0 test
- [x] enable tests in manifest once enabled in dd-trace-py

## Motivation
- Test that libraries obfuscate spans before sending them to `/stats`

## Changes
- new tests

## Workflow

## Reviewer checklist

### DIFF
--- a/tests/stats/test_stats.py
+++ b/tests/stats/test_stats.py
@@ -277,6 +277,96 @@ class Test_Client_Stats_Future_Obfuscation_Version:
             )
 
 
+@features.client_side_stats_supported
+@scenarios.trace_stats_computation_missing_obfuscation_version
+class Test_Client_Stats_Missing_Obfuscation_Version:
+    """Test that the SDK skips client-side obfuscation when the agent does not advertise obfuscation_version"""
+
+    def setup_no_obfuscation(self):
+        """Setup for missing obfuscation version test - generates SQL spans"""
+        test_user_ids = ["1", "2", "admin", "test"]
+        for user_id in test_user_ids:
+            weblog.get(f"/rasp/sqli?user_id={user_id}")
+
+    def test_no_obfuscation(self):
+        """Test that the SDK does not obfuscate stats and does not send the obfuscation header
+        when the agent does not advertise obfuscation_version in /info.
+
+        Validates:
+        - Datadog-Obfuscation-Version header is NOT present on any stats payload
+        - SQL resource names are NOT obfuscated (raw literals still present)
+        """
+        sql_stats = []
+        obfuscation_header_found = False
+
+        for data in interfaces.library.get_data("/v0.6/stats"):
+            headers = {h[0].lower(): h[1] for h in data["request"]["headers"]}
+            if "datadog-obfuscation-version" in headers:
+                obfuscation_header_found = True
+
+            payload = data["request"]["content"]
+            for bucket in payload.get("Stats", []):
+                for stat in bucket.get("Stats", []):
+                    if stat.get("Type") == "sql" and stat["Resource"].startswith("SELECT"):
+                        sql_stats.append(stat)
+
+        assert not obfuscation_header_found, (
+            "Datadog-Obfuscation-Version header should NOT be present when agent does not advertise obfuscation_version"
+        )
+
+        assert len(sql_stats) == 4, "Expected 4 distinct SQL stats entries because obfuscation was not applied client-side"
+        for stat in sql_stats:
+            assert "?" not in stat["Resource"], (
+                f"SQL resource should NOT be obfuscated when agent does not advertise obfuscation_version, "
+                f"but got: '{stat['Resource']}'"
+            )
+
+
+@features.client_side_stats_supported
+@scenarios.trace_stats_computation_obfuscation_version_zero
+class Test_Client_Stats_Obfuscation_Version_Zero:
+    """Test that the SDK skips client-side obfuscation when the agent advertises obfuscation_version=0"""
+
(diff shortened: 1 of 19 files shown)


tests/stats/test_stats.py
```diff
@@ -133,6 +133,147 @@ def test_grpc_status_code(self):
         )
 
 
+@features.client_side_stats_supported  # FIXME: create a new feature ?
```
VianneyRuhlmann: I don't think this has to be a new feature since obfuscation is part of the CSS spec and technically a "required" feature

tests/stats/test_stats.py
```diff
@@ -133,6 +133,147 @@ def test_grpc_status_code(self):
...
+
+            payload = data["request"]["content"]
+            for bucket in payload.get("Stats", []):
+                for stat in bucket.get("Stats", []):
+                    if stat.get("Type") == "sql" and stat["Resource"].startswith("SELECT"):
+                        sql_stats.append(stat)
+
+        assert obfuscation_header_found, "Datadog-Obfuscation-Version header not found on any stats payload"
+
+        assert len(sql_stats) > 0, "Expected at least one SQL stats entry"
```
VianneyRuhlmann: Out of curiosity, is there something that can cause len(sql_stats) to not be 4 (1,2,admin,test), or is it just to make it easier to add more values to test_user_ids ?

utils/_context/_scenarios/__init__.py
```diff
@@ -119,6 +119,43 @@ class _Scenarios:
...
+        # Same as trace_stats_computation but with the agent being configured with obfuscation disabled, to test that
+        # the SDK correctly reads the obfuscation config from agent's /info and respects it.
+        weblog_env={
+            "DD_TRACE_STATS_COMPUTATION_ENABLED": "true",  # default env var for CSS
+            "DD_TRACE_COMPUTE_STATS": "true",
+            "DD_TRACE_FEATURES": "discovery",
+            "DD_TRACE_TRACER_METRICS_ENABLED": "true",  # java
+        },
+        agent_env={
+            "DD_APM_SQL_OBFUSCATION_MODE": "normalize_only",
```
VianneyRuhlmann, line 153: I don't see anything about disabling the obfuscation in the [spec](https://datadoghq.atlassian.net/wiki/spaces/APM/pages/5758458333/v1.2.0#Obfuscation-Decision-Logic) is it done by setting the obfuscation version to 0 (if not maybe we need a scenario for the missing/version 0 case)

tests/stats/test_stats.py
Eldolfin: For some reason stats appear to be sent twice by dd-trace-py so `len(sql_stats) == 8` but I feel like other tracers might send them only once which would also be correct for this test. I could however assert that each bucket has 4 stats entry if you want

utils/_context/_scenarios/__init__.py
Eldolfin: It's not a setting from the CSS spec but to avoid breaking changes in the obfuscation behavior when switching from agent obfuscation to client obfuscation I decided that following as much of the config as possible would be preferable.

https://github.com/DataDog/datadog-agent/blob/main/pkg/config/config_template.yaml#L1633

tests/stats/test_stats.py
Eldolfin: Ok I just tried it again and what I just said was something I misremembered from when I started working on this. it is indeed == 4 so I just change it to that instead

utils/_context/_scenarios/__init__.py
VianneyRuhlmann: Sorry I wasn't really clear what I meant was:
1) Agent obfuscation config doesn't seem to be covered in the current spec, maybe it should be updated (CC @ichinaski)
2) Does this scenario implies `Agent version 0 or missing` if not maybe we want to run a scenario for this case.

VianneyRuhlmann (approved):
LGTM, the idea for a version 0 scenario can be addressed in a follow-up PR if needed

utils/_context/_scenarios/__init__.py
Eldolfin: 1. Yes, I think this is a shortcoming of the spec currently, because with CSS disabled the agent would have followed the obfuscation config.
2. No it's exactly like `TRACE_STATS_COMPUTATION` but with config such as obfuscation makes the 4 spans different so that they end up in different buckets so it's visible. I'll add a similar test with a missing version 👍🏽

tests/stats/test_stats.py
```diff
@@ -68,20 +69,14 @@ def setup_obfuscation(self):
...
+            if s["Type"] != "sql" or "?" not in s["Resource"]:
+                continue
             logger.debug(f"asserting on {s}")
             hits += s["Hits"]
             top_hits += s["TopLevelHits"]
-            assert s["Type"] == "sql", "expect 'sql' type"
-        assert stats_count <= 4, (
-            "expect <= 4 stats"
-        )  # Normally this is exactly 2 but in certain high load this can flake and result in additional payloads where hits are split across two payloads
         assert hits == top_hits >= 4, "expect at least 4 'OK' hits and top level hits across all payloads"
```
Eldolfin, line 81: should be restored

Eldolfin:
@codex review

ichinaski (approved):

nccatoni (approved):

PR #6729, opened 2026-04-11
debugger: fix wait_for_all_probes to accumulate state across requests

watson:
## Motivation

`wait_for_all_probes` was silently broken for tracers/agents that send probe diagnostics across multiple HTTP requests. The function recreated a local `found_ids` set on every callback invocation, meaning probe IDs discovered from earlier requests were lost. It could only return `True` if a single request happened to contain diagnostics for all probes — otherwise it always timed out and returned `False`.

This was never caught because the return value was discarded. When we tried to assert on it (to surface setup timeouts as clear test failures), the bug became visible: Ruby and Go tests failed consistently.

## Changes

1. **Fix `wait_for_all_probes` in `tests/debugger/utils.py`**: Move `found_ids` from a local variable (recreated each callback invocation) to instance state (`self._wait_found_ids`, initialized once before the wait loop). Probe IDs now accumulate correctly across multiple HTTP requests from the agent.

2. **Report setup timeouts in `tests/debugger/test_debugger_probe_status.py`**: Record a `setup_failures` entry when `wait_for_all_probes` times out, so it surfaces as a clear test failure via `assert_setup_ok()` rather than being silently ignored. Uses `setup_failures` instead of a bare `assert` to avoid crashing pytest during the collection phase.

## Workflow

## Reviewer checklist

### DIFF
--- a/tests/debugger/utils.py
+++ b/tests/debugger/utils.py
@@ -392,11 +392,13 @@ class BaseDebuggerTest:
     def wait_for_all_probes(self, statuses: list[ProbeStatus], timeout: int = 30) -> bool:
         logger.debug("Wating for all probes")
         self._wait_successful = False
-        self._wait_found_ids: set[str] = set()
-        interfaces.agent.wait_for(lambda data: self._wait_for_all_probes(data, statuses=statuses), timeout=timeout)
+        found_ids: set[str] = set()
+        interfaces.agent.wait_for(
+            lambda data: self._wait_for_all_probes(data, statuses=statuses, found_ids=found_ids), timeout=timeout
+        )
         return self._wait_successful
 
-    def _wait_for_all_probes(self, data: dict[str, Any], statuses: list[ProbeStatus]):
+    def _wait_for_all_probes(self, data: dict[str, Any], statuses: list[ProbeStatus], found_ids: set[str]):
         def _check_all_probes_status(probe_diagnostics: ProbeDiagnosticsCollection, statuses: list[ProbeStatus]):
             statuses = statuses + ["ERROR"]
             logger.debug(f"Waiting for these probes to be in {statuses}: {self.probe_ids}")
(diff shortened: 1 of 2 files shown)


watson:
* **#6729** <a href="https://app.graphite.com/github/pr/DataDog/system-tests/6729?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/system-tests/6729?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

cbeauchesne (approved):

tests/debugger/utils.py
```diff
@@ -392,12 +392,11 @@ def send_weblog_request(self, request_path: str, *, reset: bool = True) -> None:
     def wait_for_all_probes(self, statuses: list[ProbeStatus], timeout: int = 30) -> bool:
         logger.debug("Wating for all probes")
         self._wait_successful = False
+        self._wait_found_ids: set[str] = set()
```
ajwerner: this is fine, but I feel like using the private member on the class strikes me as a little off. What about adding `found_ids` as a parameter to `_wait_for_all_probes` and then:

```python
found_ids = set()
wait_for = lambda data: self._wait_for_all_probes(data, statuses=statuses, found_ids=found_ids)
interfaces.agent.wait_for(wait_for, timeout=timeout)
```
or use https://docs.python.org/3/library/functools.html#functools.partial
```python
from functools import partial

found_ids = set()
wait_for = partial(
    self._wait_for_all_probes,
    statuses=statuses,
    found_ids=found_ids,
)
interfaces.agent.wait_for(wait_for, timeout=timeout)
```

ajwerner (approved):
Fine with me, my suggestions are nit-picky and I don't write Python much these days so take it with a grain of salt

tests/debugger/utils.py
watson: Good point 👍 I've pushed a fix

p-datadog (approved):
:+1: LGTM

PR #6742, opened 2026-04-14
Enhance Docker installation script for Ubuntu: add support for Ubuntu…

juanjux:
## Motivation

Fix several CI failures preventing the release of the ddtrace-py versions, example:

https://gitlab.ddbuild.io/DataDog/apm-reliability/dd-trace-py/-/jobs/1592849083

On Ubuntu 23.04 arm64 the VM codename is lunar, but Docker’s official apt repo often does not provide docker-ce for that (EOL) release on arm64, so the generic install in auto-inject_prepare_docker.yml could leave the host without Docker while the script still appeared to succeed because it did not use set -e; downstream steps then failed (docker missing, no docker.service, installer and container weblog broken). The fix adds a dedicated os_branch: ubuntu23 block that runs before the generic Debian path and, for lunar/mantic, points the Docker apt source at the jammy suite so docker-ce installs reliably, and enables set -euo pipefail so a failed install fails the provision step instead of silently continuing.

## Workflow

## Reviewer checklist


robertomonteromiguel (changes requested):
There is still failures on the pipeline. But I can see it's only failing for the app test-app-java-buildpack. We could exclude this machine only for this case: [https://github.com/DataDog/system-tests/blob/affeeeec0dfb4f748da97784fcacae349ad99830/utils/scripts/ci_orchestrators/aws_ssi.json#L415](https://github.com/DataDog/system-tests/blob/affeeeec0dfb4f748da97784fcacae349ad99830/utils/scripts/ci_orchestrators/aws_ssi.json#L415)

juanjux:
> There is still failures on the pipeline. But I can see it's only failing for the app test-app-java-buildpack. We could exclude this machine only for this case:
> 
> https://github.com/DataDog/system-tests/blob/affeeeec0dfb4f748da97784fcacae349ad99830/utils/scripts/ci_orchestrators/aws_ssi.json#L415

I defer to your experience here, WDYT would be better?

PR #6834, opened 2026-05-01
OTLP trace metrics export

ida613:
## Motivation

Datadog tracing libraries can export client-computed span aggregate metrics in OTLP format as a histogram (`traces.span.sdk.metrics.duration`), complementing OTLP trace export. This is the basis for RED metrics (Requests / Errors / Duration) derived from the histogram:

- **Hits** = `count{*}`
- **Errors** = `count{error:true}`
- **TopLevelHits** = `count{datadog.span.top_level:true}`

This PR adds black-box parametric coverage for that feature: parametric apps run with various configurations, spans are generated, and the resulting OTLP metrics (and native v0.6 stats, where relevant) are captured and asserted on. It lands the test cases called out in the RFC ([SEMCON-1093](https://docs.google.com/document/d/1cDB_NVkyY_DXzb_EfQS6BsGu5Rx3gnf-tbpgJCYbIkc/edit#heading=h.djcnhvj75u76)).

## What the tests validate

`tests/parametric/test_otlp_trace_metrics.py` — 42 tests grouped by functional requirement, under the `PARAMETRIC` scenario (`@features.client_side_stats_supported`):

| Req | Area | Tests |
| --- | --- | --- |
| FR01 | Enablement & configuration (explicit/default on/off, gated by metrics and OTLP trace export) | 5 |
| FR02 | Metric identity (single named histogram, no native/SMC names) & mutual exclusion with native v0.6 stats (incl. `Datadog-Client-Computed-Stats` header) | 4 |
| FR03 | Metric shape — delta-temporality histogram, unit `s` | 2 |
| FR04 | Span selection — only top-level or measured spans produce data points | 2 |
| FR05 | Sampling independence — metrics computed before head-based sampling, from 100% of spans | 1 |
| FR06 | OpenTelemetry span attributes (http/rpc/status/kind) and resource attributes — service identity on the resource, `service.name` on a data point when a span's service differs from the default, hostname, telemetry SDK | 14 |
| FR07 | OTel-semantics mode (`DD_TRACE_OTEL_SEMANTICS_ENABLED`) — only OpenTelemetry attributes, no `datadog.*`/`dd.*` | 4 |
| FR08 | Datadog attributes (default mode) — operation name, span type, top-level, origin, process tags | 8 |
| FR09 | RED metric derivation — single-point consistency and error count | 2 |

End-state focus: each test exercises one behaviour and asserts on the final exported payload (not on transitional state). Assertions are shape-tolerant where a value is library-dependent — FR03 checks `unit`, delta temporality and `0 < sum < 60`; FR09 checks `count`, `min == max` and `sum(bucketCounts)` for a single span. Flush cadence is fixed at 10s and shortened in tests via the internal `_DD_TRACE_METRICS_OTEL_FLUSH_INTERVAL` (ms). Tests pin HTTP/JSON export, the transport common to all libraries.

Cross-tracer parity (documented, not asserted): `sum`/`min`/`max` are exact in dd-trace-js (sketch scalars) but sketch-approximated in libdatadog/dd-trace-py; `host.name` source differs (libdatadog honors `DD_HOSTNAME`, dd-trace-js uses `os.hostname()`), so the hostname test asserts presence only.

## Activation

The SDK implementations ship in **future library releases**, so these tests cannot pass against currently released libraries in CI. They are therefore skipped for **all** languages via a module-level `missing_feature` entry in every manifest. `python` and `nodejs` additionally keep their per-test skips (library-specific gaps such as nodejs native v0.6 stats); leaving those in place means re-enabling the suite once the feature ships is simply removing the single module-level line.

## Supporting changes

- `utils/build/docker/nodejs/parametric/server.js`, `utils/build/docker/nodejs/install_ddtrace.sh` — parametric test-client support for the OTLP trace-metrics path.
- `utils/docker_fixtures/_test_agent.py` — polling helpers (`wait_for_num_otlp_metrics`, `wait_for_num_v06_stats`) to accommodate OTLP/native export latency.

## What to review

- The manifest activation strategy (module-level skip + retained per-test skips) and that re-enabling is a one-line change per language.
- That the assertions are shape-tolerant for library-dependent values yet still fail loudly with the actual payload captured in the failure message.
- The OTel semantic-convention attribute mapping and the default vs OTel-semantics mode boundaries (FR06/FR07/FR08).

## Reviewer checklist

* [ ] Anything but \`tests/\` or \`manifests/\` is modified ? I have the approval from [R&P team](https://dd.enterprise.slack.com/archives/C025TJ4RZ8X)
    * [ ] the relevant \`build-XXX-image\` label is present

[SEMCON-1093]: https://datadoghq.atlassian.net/browse/SEMCON-1093?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


manifests/cpp.yml
```diff
@@ -195,6 +195,7 @@ manifest:
   tests/parametric/test_otel_span_with_baggage.py::Test_Otel_Span_With_Baggage: missing_feature
   tests/parametric/test_otel_tracer.py::Test_Otel_Tracer::test_otel_force_flush: irrelevant (library does not implement OpenTelemetry)
   tests/parametric/test_otel_tracer.py::Test_Otel_Tracer::test_otel_simple_trace: irrelevant (library does not implement OpenTelemetry)
+  tests/parametric/test_otlp_trace_metrics.py: missing_feature (OTLP trace metrics export is only implemented in Python)
```
mabdinur:
```suggestion
  tests/parametric/test_otlp_trace_metrics.py: missing_feature
```

manifests/dotnet.yml
```diff
@@ -962,6 +962,7 @@ manifest:
       component_version: <=2.41.0
   tests/parametric/test_otel_span_with_baggage.py::Test_Otel_Span_With_Baggage: missing_feature
   tests/parametric/test_otel_tracer.py::Test_Otel_Tracer: v2.8.0
+  tests/parametric/test_otlp_trace_metrics.py: missing_feature (OTLP trace metrics export is only implemented in Python)
```
mabdinur:
```suggestion
  tests/parametric/test_otlp_trace_metrics.py: missing_feature
```

manifests/golang.yml
```diff
@@ -1187,6 +1187,7 @@ manifest:
   ? tests/parametric/test_otel_span_methods.py::Test_Otel_Span_Methods::test_otel_span_strict_reserved_attributes_overrides_analytics_event
   : "irrelevant (\"Go tracer decided to always set _dd1.sr.eausr: 1 for truthy analytics.event inputs, else 0\")"
   tests/parametric/test_otel_span_with_baggage.py::Test_Otel_Span_With_Baggage: missing_feature
+  tests/parametric/test_otlp_trace_metrics.py: missing_feature (OTLP trace metrics export is only implemented in Python)
```
mabdinur:
```suggestion
  tests/parametric/test_otlp_trace_metrics.py: missing_feature
```

manifests/java.yml
```diff
@@ -3781,6 +3781,7 @@ manifest:
   tests/parametric/test_otel_tracer.py::Test_Otel_Tracer::test_otel_force_flush:
     - declaration: missing_feature (OTel resource naming implemented in 1.24.0)
       component_version: <=1.23.0
+  tests/parametric/test_otlp_trace_metrics.py: missing_feature (OTLP trace metrics export is only implemented in Python)
```
mabdinur:
```suggestion
  tests/parametric/test_otlp_trace_metrics.py: missing_feature
```

manifests/nodejs.yml
```diff
@@ -2101,6 +2101,7 @@ manifest:
   tests/parametric/test_otel_span_methods.py::Test_Otel_Span_Methods::test_otel_start_span: missing_feature (New operation name mapping not yet implemented)
   tests/parametric/test_otel_span_with_baggage.py::Test_Otel_Span_With_Baggage: *ref_5_32_0
   tests/parametric/test_otel_tracer.py::Test_Otel_Tracer::test_otel_force_flush: missing_feature (Not implemented)
+  tests/parametric/test_otlp_trace_metrics.py: missing_feature (OTLP trace metrics export is only implemented in Python)
```
mabdinur:
```suggestion
  tests/parametric/test_otlp_trace_metrics.py: missing_feature
```

manifests/php.yml
```diff
@@ -839,6 +839,7 @@ manifest:
       component_version: <=0.95.0
   tests/parametric/test_otel_span_with_baggage.py::Test_Otel_Span_With_Baggage: missing_feature
   tests/parametric/test_otel_tracer.py::Test_Otel_Tracer: v0.94.0
+  tests/parametric/test_otlp_trace_metrics.py: missing_feature (OTLP trace metrics export is only implemented in Python)
```
mabdinur:
```suggestion
  tests/parametric/test_otlp_trace_metrics.py: missing_feature
```

manifests/ruby.yml
```diff
@@ -1569,6 +1566,7 @@ manifest:
   ? tests/parametric/test_otel_span_methods.py::Test_Otel_Span_Methods::test_otel_span_strict_reserved_attributes_overrides_analytics_event
   : "irrelevant (\"Ruby tracer decided to always set _dd1.sr.eausr: 1 for truthy analytics.event inputs, else 0\")"
   tests/parametric/test_otel_span_with_baggage.py::Test_Otel_Span_With_Baggage: missing_feature
+  tests/parametric/test_otlp_trace_metrics.py: missing_feature (OTLP trace metrics export is only implemented in Python)
```
mabdinur:
```suggestion
  tests/parametric/test_otlp_trace_metrics.py: missing_feature
```

manifests/rust.yml
```diff
@@ -209,6 +209,7 @@ manifest:
   : missing_feature (Not implemented)
   tests/parametric/test_otel_span_with_baggage.py::Test_Otel_Span_With_Baggage: missing_feature
   tests/parametric/test_otel_tracer.py::Test_Otel_Tracer: v0.0.1
+  tests/parametric/test_otlp_trace_metrics.py: missing_feature (OTLP trace metrics export is only implemented in Python)
```
mabdinur:
```suggestion
  tests/parametric/test_otlp_trace_metrics.py: missing_feature
```

manifests/python.yml
```diff
@@ -1887,6 +1887,8 @@ manifest:
       component_version: <4.3.1
   tests/parametric/test_otel_span_with_baggage.py::Test_Otel_Span_With_Baggage: v2.18.0
   tests/parametric/test_otel_tracer.py::Test_Otel_Tracer: v2.8.0
+  tests/parametric/test_otlp_trace_metrics.py::Test_FR01_Enablement_Configuration: v4.11.0.dev
+  tests/parametric/test_otlp_trace_metrics.py: v4.11.0.dev
```
mabdinur:
```suggestion
  tests/parametric/test_otlp_trace_metrics.py: missing_feature
```

tests/parametric/test_otlp_trace_metrics.py
```diff
@@ -0,0 +1,717 @@
...
+SERVICE = "test-otlp-stats-svc"
+TRUTHY = ("yes", "true", "1")
+# OTLP AggregationTemporality enum: DELTA == 1.
+AGGREGATION_TEMPORALITY_DELTA = 1
+
+# Common env shared by every test. The native OTLP stats worker ticks (and closes stat buckets)
+# on the stats writer interval, and only releases buckets older than 2 intervals, so use a short
+# interval to keep client-computed stats exported within the test window.
+_BASE_ENVVARS = {
+    "OTEL_EXPORTER_OTLP_METRICS_PROTOCOL": "http/protobuf",
```
mabdinur:
```suggestion
    "OTEL_EXPORTER_OTLP_METRICS_PROTOCOL": "http/json",
```

tests/parametric/test_otlp_trace_metrics.py
```diff
@@ -0,0 +1,717 @@
...
+
+SPAN_DURATION_METRIC = "dd.trace.span.duration"
+SERVICE = "test-otlp-stats-svc"
+TRUTHY = ("yes", "true", "1")
+# OTLP AggregationTemporality enum: DELTA == 1.
+AGGREGATION_TEMPORALITY_DELTA = 1
+
+# Common env shared by every test. The native OTLP stats worker ticks (and closes stat buckets)
+# on the stats writer interval, and only releases buckets older than 2 intervals, so use a short
+# interval to keep client-computed stats exported within the test window.
```
mabdinur:
```suggestion
# interval to keep client-computed stats exported within the test window. The initial implementation only supports json export.
```

cbeauchesne (commented):
LGTM, but can you get a review from someone familiar with the tested feature ?

gnufede (approved):
Python manifest looks good to me

mabdinur (approved):

tests/parametric/test_otlp_trace_metrics.py
```diff
@@ -0,0 +1,1318 @@
...
+        assert headers.get("datadog-client-computed-stats", "").lower() in TRUTHY, (
+            f"Expected Datadog-Client-Computed-Stats to be truthy, got headers: {headers}"
+        )
+
+    @pytest.mark.parametrize(
+        "library_env",
+        [
+            {
+                **_BASE_ENVVARS,
+                "OTEL_TRACES_SPAN_METRICS_ENABLED": "false",
```
zacharycmontoya, line 423: What happens if both are true? Which one wins? This could use a test case

zacharycmontoya: Or maybe it's worth setting the env variable `OTEL_TRACES_SPAN_METRICS_ENABLED=true` in the `test_fr02_3_otlp_suppresses_native_stats` test case if OTLP wins

tests/parametric/test_otlp_trace_metrics.py
```diff
@@ -0,0 +1,1318 @@
...
+            t.dd_flush()
+
+        with pytest.raises(ValueError):
+            test_agent.wait_for_num_otlp_metrics(num=1)
+
+    @pytest.mark.parametrize(
+        "library_env",
+        [{**_BASE_ENVVARS, "OTEL_TRACES_EXPORTER": "none", "DD_METRICS_OTEL_ENABLED": "true"}],
+    )
+    def test_fr01_5_disabled_when_traces_exporter_not_otlp(
```
zacharycmontoya, line 326: The test name describes the scenario as "traces_exporter_not_otlp" but that is already tested in `test_fr01_1_enabled_explicit`. I suggest renaming this because what you're actually doing is disabling traces altogether with `OTEL_TRACES_EXPORTER=none`. You could also add a test that disables tracing via the DD environment variable, so `DD_APM_TRACING_ENABLED=false`

```suggestion
    def test_fr01_5_disabled_when_tracing_is_disabled(
```

tests/parametric/test_otlp_trace_metrics.py
```diff
@@ -0,0 +1,1318 @@
...
+    ):
+        """Span kind maps to the data-point attribute span.kind."""
+        with test_library as t:
+            with t.dd_start_span(name="web.request", service=SERVICE, typestr="web") as span:
+                span.set_meta("span.kind", "server")
+            t.dd_flush()
+
+        metrics = test_agent.wait_for_num_otlp_metrics(num=1)
+        attrs = _data_point_attrs(_duration_data_points(metrics)[0])
+        assert attrs.get("span.kind") == "server", f"Expected span.kind=server, got attrs: {attrs}"
```
zacharycmontoya, line 609: Should the value of the `span.kind` attribute match the casing of span metrics connector (in this case SERVER)? See the README [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/connector/spanmetricsconnector)

tests/parametric/test_otlp_trace_metrics.py
```diff
@@ -0,0 +1,1318 @@
...
+        """Http status code maps to the data-point attribute http.response.status_code."""
+        with test_library as t:
+            with t.dd_start_span(name="web.request", service=SERVICE, typestr="web") as span:
+                span.set_meta("http.status_code", "200")
+            t.dd_flush()
+
+        metrics = test_agent.wait_for_num_otlp_metrics(num=1)
+        attrs = _data_point_attrs(_duration_data_points(metrics)[0])
+        assert attrs.get("http.response.status_code") == 200, (
+            f"Expected http.response.status_code == 200 (typed int), got attrs: {attrs}"
```
zacharycmontoya, line 644: Are we mixing and matching attribute types? Or should all attributes on the time-series have string values?

zacharycmontoya: FWIW, I asked Claude to take a look at the [Span Metrics Connector](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/connector/spanmetricsconnector) to understand how the 'dimensions' configuration copies span attributes to the resulting time series. If an attribute with the matching name exists on the span, then it is copied as-is to the timeseries, so an example of `ip.port` with an IntValue would be copied as an IntValue onto the resulting metrics timeseries

tests/parametric/test_otlp_trace_metrics.py
```diff
@@ -0,0 +1,1318 @@
...
+            f"Expected service.version=1.2.3, got: {resource_attrs}"
+        )
+        # The deployment environment semantic convention was renamed in 1.27.0.
+        assert (
+            resource_attrs.get("deployment.environment") == "prod"
+            or resource_attrs.get("deployment.environment.name") == "prod"
+        ), f"Expected deployment environment=prod, got: {resource_attrs}"
+
+        # The span uses the configured default service, so its data point omits service.name.
+        assert SERVICE not in _data_point_services(metrics), (
```
zacharycmontoya, line 758: I think we should always emit `service.name` on the time series as a metric attribute, since that follows behavior of SMC. I think OpenTelemetry users would expect the metric itself to have all the metric attributes to fully describe the spans, which includes the service name

tests/parametric/test_otlp_trace_metrics.py
```diff
@@ -0,0 +1,1318 @@
...
+    ):
+        """Resource maps to the data-point attribute span.name."""
+        with test_library as t:
+            with t.dd_start_span(name="web.request", service=SERVICE, resource="/users", typestr="web"):
+                pass
+            t.dd_flush()
+
+        metrics = test_agent.wait_for_num_otlp_metrics(num=1)
+        attrs = _data_point_attrs(_duration_data_points(metrics)[0])
+        assert attrs.get("span.name") == "/users", f"Expected span.name=/users, got attrs: {attrs}"
```
zacharycmontoya, line 592: If we want to add assertions for `datadog.resource.name`, we should also assert that it's not included here. Otherwise, we can remove all assertions on this name since it's no longer mentioned in the RFC

```suggestion
        assert attrs.get("span.name") == "/users", f"Expected span.name=/users, got attrs: {attrs}"
        assert "datadog.resource.name" not in attrs, f"datadog.resource.name must be absent: {attrs}"
```

tests/parametric/test_otlp_trace_metrics.py
```diff
@@ -0,0 +1,1318 @@
...
+        """OTel semantic-convention attributes are still emitted in OTel-semantics mode."""
+        with test_library as t:
+            with t.dd_start_span(name="web.request", service=SERVICE, typestr="web") as span:
+                span.set_meta("http.method", "GET")
+                span.set_meta("http.route", "/users/{id}")
+            t.dd_flush()
+
+        metrics = test_agent.wait_for_num_otlp_metrics(num=1)
+        attrs = _data_point_attrs(_duration_data_points(metrics)[0])
+        assert attrs.get("http.request.method") == "GET", f"Expected http.request.method=GET, got attrs: {attrs}"
```
zacharycmontoya, line 943: By [default](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/connector/spanmetricsconnector#overview), the span metrics connector only emits the following attributes:
- service.name
- span.name
- span.kind
- status.code
- (feature-gate) collector.instance.id
- (feature-gate) otel.status_code

In our `DD_TRACE_OTEL_SEMANTICS_ENABLED` mode we are emitting additional fields like `env` and these HTTP ones. Is the intention that even in OTel semantics mode we emit the full set of attributes our own stats computation needs? If so, can we consider a configuration flag to not emit a user-specified set of attribute keys?

PR #6837, opened 2026-05-03
Verify compact OTLP histograms where consecutive empty buckets are collapsed

mcculls:
## Motivation

The Java SDK exports OTLP histograms in a compact form where consecutive empty buckets are collapsed into a single bucket using the last boundary. This saves bandwidth when exporting histograms that have lots of explicit boundaries, but only a few of them have counts.

## Changes

Creates a new `assert_histogram_buckets` method that correctly validates both non-compacted and compacted histogram buckets.

Also fixes an issue in the original test where the `aggregation_temporality` parameter was not actually tested due to shadowing.

## Workflow

## Reviewer checklist

### DIFF
--- a/tests/parametric/test_otel_metrics.py
+++ b/tests/parametric/test_otel_metrics.py
@@ -143,7 +143,7 @@ def assert_gauge_aggregation(gauge_aggregation: dict, value: int, attributes: di
 
 def assert_histogram_aggregation(
     histogram_aggregation: dict[str, str | list[dict]],
-    aggregation_temporality: str,
+    histogram_temporality: str,
     count: int,
     sum_value: int,
     min_value: int,
@@ -152,8 +152,8 @@ def assert_histogram_aggregation(
     bucket_counts: list[int],
     attributes: dict[str, str],
 ):
-    aggregation_temporality: str = histogram_aggregation["aggregation_temporality"]
-    assert aggregation_temporality.casefold() == aggregation_temporality.casefold()
+    aggregation_temporality = str(histogram_aggregation["aggregation_temporality"])
+    assert aggregation_temporality.casefold() == histogram_temporality.casefold()
 
     assert isinstance(histogram_aggregation["data_points"], list)
     data_points: list[dict] = histogram_aggregation["data_points"]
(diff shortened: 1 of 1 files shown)


tests/parametric/test_otel_metrics.py
```diff
@@ -164,14 +165,54 @@ def assert_histogram_aggregation(
             assert histogram_data_point["min"] == min_value
             assert histogram_data_point["max"] == max_value
 
-            assert histogram_data_point["explicit_bounds"] == bucket_boundaries
-            assert list(map(int, histogram_data_point["bucket_counts"])) == bucket_counts
+            if test_library.lang == "java":
```
cbeauchesne: Could you split the test logic in two, and use manifests to activate the relevant tests?

It's usually way better to use manifest as the single place of test activation/deactivation, as it's easier to track what is done, and why. Otherwise, we bake deep in the test logic those activations, making them harder to follow.

cbeauchesne (changes requested):

tests/parametric/test_otel_metrics.py
mcculls: Good point - I'm rewriting the test so it verifies the bucket content while being flexible about zero buckets. This means we don't need to tie a language to a particular approach.

mcculls: @cbeauchesne I refactored the approach to avoid needing to know the language

Also found a small bug in the original test which I fixed where `aggregation_temporality` wasn't being checked

cbeauchesne (approved):

tests/parametric/test_otel_metrics.py
```diff
@@ -163,15 +163,64 @@ def assert_histogram_aggregation(
...
+        assert actual_bounds == []
+        assert actual_counts == []
+        return
+    # otherwise we expect counts for every boundary plus an overflow count
+    # (some empty buckets may be collapsed into one on languages like Java)
+    assert len(bucket_counts) == len(bucket_bounds) + 1
+    assert len(actual_counts) == len(actual_bounds) + 1
+    idx = 0
+    last_collapsed_bound = None
+    for expected_bound, expected_count in zip(bucket_bounds, bucket_counts, strict=False):
```
zacharycmontoya, line 192: I'm convinced this looks right after working with AI to come up with a real example 🎉

```
  Worked example

  Expected: bounds [10, 20, 30, 40, 50], counts [3, 0, 0, 0, 2, 1] (last is overflow).
  Compacted actual: bounds [10, 40, 50], counts [3, 0, 2, 1].

  ┌──────┬────────────────┬────────────────────────────────────────────────────────────────────────────────────┬─────┬──────────────────────┐
  │ iter │ (bound, count) │                                       branch                                       │ idx │ last_collapsed_bound │
  ├──────┼────────────────┼────────────────────────────────────────────────────────────────────────────────────┼─────┼──────────────────────┤
  │ 1    │ (10, 3)        │ aligned → consume actual[0]=(10,3)                                                 │ 1   │ None                 │
  ├──────┼────────────────┼────────────────────────────────────────────────────────────────────────────────────┼─────┼──────────────────────┤
  │ 2    │ (20, 0)        │ empty, not aligned → stash                                                         │ 1   │ 20                   │
  ├──────┼────────────────┼────────────────────────────────────────────────────────────────────────────────────┼─────┼──────────────────────┤
  │ 3    │ (30, 0)        │ empty, not aligned → stash                                                         │ 1   │ 30                   │
  ├──────┼────────────────┼────────────────────────────────────────────────────────────────────────────────────┼─────┼──────────────────────┤
  │ 4    │ (40, 0)        │ empty, not aligned → stash                                                         │ 1   │ 40                   │
  ├──────┼────────────────┼────────────────────────────────────────────────────────────────────────────────────┼─────┼──────────────────────┤
  │ 5    │ (50, 2)        │ non-empty: assert actual[1]=(40,0) ✓, idx=2; then assert actual[2]=(50,2) ✓, idx=3 │ 3   │ None                 │
  └──────┴────────────────┴────────────────────────────────────────────────────────────────────────────────────┴─────┴──────────────────────┘

  Post-loop: idx == len(actual_bounds) == 3 ✓, overflow actual_counts[-1] == 1 == bucket_counts[-1] ✓.
```

zacharycmontoya (approved):
LGTM!

PR #6870, opened 2026-05-06
php weblog: detect new _automated API via function_exists

cataphract:
## Motivation

master of dd-trace-php renamed some internal functions that are unfortunately directly used by the weblog (they are in fact meant to be internally called by integrations). For now, just call the new functions.

## Reviewer checklist


cbeauchesne (commented):
Please do not merge yet, the base image mechanism is pretty new (yesterday) and I need to fix several thing before.

cbeauchesne:
and BTW, I'm currently working on having base images for apache in a separated PR : #6874

cataphract:
@cbeauchesne Please note these additional changes, if you want to take it over:

* replace dumb-init invocation with exec dumb-init, I think it's more correct because dumb-init is supposed to run as pid 1. Maybe it would be more straightforward to move it to ENTRYPOINT though
* The base image had the php files, meaning the base image would need to be rebuild each time a php file changes. This is not very practical. So I used ONBUILD to make it the child images add the files themselves. This also allowed for the simplification of the child dockerfiles (they only need a FROM line)
* I think this block is dead code:
   ```
   if [[ -f "/etc/php/98-ddtrace.ini" ]]; then
    grep -E 'datadog.trace.request_init_hook|datadog.trace.sources_path' /etc/php/98-ddtrace.ini >> /etc/php/php.ini
   fi
  ```

cbeauchesne:
> The base image had the php files, meaning the base image would need to be rebuild each time a php file changes. This is not very practical. So I used ONBUILD to make it the child images add the files themselves. This also allowed for the simplification of the child dockerfiles (they only need a FROM line)



yeah, it's an error, I'm fixing this

bwoebi (approved):

robertomonteromiguel (approved):

PR #6879, opened 2026-05-07
Improve performances by reducing RC poll interval on all E2E scenarios on python and node.js

rochdev:
## Summary

- Sets `DD_REMOTE_CONFIG_POLL_INTERVAL_SECONDS=0.2` as a default for all Node.js and Python scenarios in `EndToEndScenario.configure()`
- Node.js starts the poll timer only after receiving the previous response, so faster polling is safe
- Other languages use a fixed interval or have bugs that cause race conditions at shorter intervals, so the default does not apply to them
- Uses `setdefault` so any scenario that explicitly sets this value retains it
- This is a simpler version of #6851 and #6808 in hope that it will be easier to get approved.

## Test plan

- [ ] Run an RC-dependent scenario against the Node.js weblog and verify the poll interval is applied

## Human notes

Example gains from the worst offenders:

- `APPSEC_RUNTIME_ACTIVATION` goes from 9 minutes to 3 minutes.
- `APPSEC_API_SECURITY_RC` goes from 8 minutes to <3 minutes.

This is significant because for languages that parallelize the tests, our total test time of our entire CI ends up being our slowest jobs, which were those. So fixing this one problem alone will reduce our total run time significantly.

---

🤖 Generated with [Claude Code](https://claude.ai/code)


utils/_context/_scenarios/endtoend.py
```diff
@@ -309,6 +309,12 @@ def configure(self, config: pytest.Config):
 
         library = self.weblog_infra.library_name
 
+        if library == "nodejs":
```
brettlangdon:
```suggestion
        if library in ("nodejs", "python"):
```

seems viable for Python as well

brettlangdon (approved):

cbeauchesne (dismissed):
As we've discussed by zoom, my point about this is that it diverges from the standard configuration applied on our customers, and does so across all scenarios without distinction. It also introduces a different testing approach for Python/Node compared to the other languages.

As a result:

* From a QA perspective, this is not something I would recommend.
* That said, this is not my decision to make, **so I’m neither opposing nor approving it**. Each product team should weigh in on this. If there is alignment across teams, then that’s fine. 

Just one "request" change : the PR description does not include the expected time savings. This is important to help evaluate the trade-offs, so it would be helpful to add it.

rochdev:
> As we've discussed by zoom, my point about this is that it diverges from the standard configuration applied on our customers, and does so across all scenarios without distinction. It also introduces a different testing approach for Python/Node compared to the other languages.

Yes, it's always better to not have any special casing, but this is a problem that is impacting us significantly, so the trade-off is worth it. That is not necessarily the case for every language. For example, a language that already has a 1h CI would not case that system tests are slower. But for languages that have an efficient CI pipeline, this has a huge impact. So the trade-off of polling faster is worth more than the potential risk. That risk is also extremely minimal, polling should always work the same regardless of the interval. That is the nature of short-polling. Long-polling would be instant but that would be a design change.

> That said, this is not my decision to make, so I’m neither opposing nor approving it. Each product team should weigh in on this. If there is alignment across teams, then that’s fine.

Discussed this with the teams and everyone is on board with the change. Still limiting the scope to languages that are known to have a safe polling strategy.

> Just one "request" change : the PR description does not include the expected time savings. This is important to help evaluate the trade-offs, so it would be helpful to add it.

Added!

leoromanovsky (approved):

tylfin (approved):

crysmags (approved):

utils/_context/_scenarios/endtoend.py
```diff
@@ -309,6 +309,12 @@ def configure(self, config: pytest.Config):
 
         library = self.weblog_infra.library_name
 
+        if library in ("nodejs", "python"):
```
bm1549, line 312: If we want to consider this as the default behavior, it would be nice to invert the condition here so that it signals that a lower polling rate is the "default" and the higher one is anomalous

Not blocking for this PR, but definitely a nice to have

bm1549 (approved):
LGTM, one non-blocking comment

utils/_context/_scenarios/endtoend.py
rochdev: I agree with the sentiment. I'll open a follow-up PR to expand this to other languages and do some more testing, but for now the opt-in is fine I think to at least unblock Node and improve Python.

Julio-Guerra (approved):

PR #6921, opened 2026-05-13
ci: add workflow to mirror MCR images to GHCR

rochdev:
## Summary

- Adds a `workflow_dispatch` workflow to mirror container images from external registries (e.g. `mcr.microsoft.com`) to GHCR
- Mirrors images under `ghcr.io/datadog/system-tests/{source-image}:{tag}`
- Modelled after the same workflow in dd-trace-js

## Usage

Trigger manually with:
- **image**: source image (e.g. `mcr.microsoft.com/dotnet/sdk`)
- **tag**: image tag (e.g. `8.0`)

## Test plan

- [ ] Trigger the workflow manually with a .NET image and verify it appears in GHCR

## Human notes

Basically the same as https://github.com/DataDog/dd-trace-js/pull/8165. Images will then be available in [Packages](https://github.com/DataDog/system-tests/packages). Example in dd-trace-js: https://github.com/Datadog/dd-trace-js/packages

---

### DIFF
--- a/.github/workflows/mirror-image.yml
+++ b/.github/workflows/mirror-image.yml
@@ -1,0 +1,44 @@
+name: Mirror image to GHCR
+
+on:
+  workflow_dispatch:
+    inputs:
+      image:
+        description: "Source image to mirror (e.g. mcr.microsoft.com/dotnet/sdk)"
+        required: true
+        type: string
+      tag:
+        description: "Image tag to mirror (e.g. 8.0, 8.0-alpine)"
+        required: true
+        type: string
+
+permissions:
+  packages: write
+
+jobs:
+  mirror:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Log in to GHCR
+        uses: docker/login-action@4907a6ddec9925e35a0a9e82d7399ccc52663121 # v4.1.0
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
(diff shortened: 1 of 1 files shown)


.github/workflows/mirror-image.yml
```diff
@@ -0,0 +1,44 @@
...
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Resolve GHCR target
+        env:
+          IMAGE: ${{ inputs.image }}
+          REPOSITORY: ${{ github.repository }}
+        run: |
+          GHCR_TARGET="$(echo "ghcr.io/${REPOSITORY}/${IMAGE}" | tr '[:upper:]' '[:lower:]')"
```
cbeauchesne, line 34: If I understand correctly, this will replicate the image in ghcr.io/DataDog/system-tests 

But will this registry be accessible to other repositories ? 

If yes, it's perfect, we just need to replace mcr by ghcr in any reference. Otherwise, ghcr is not the good target. But we can use the same logic with docker hub.

cbeauchesne (changes requested):

.github/workflows/mirror-image.yml
rochdev: There is a setting once the image is published, for example:

image

cbeauchesne: Nice, I was not aware of this, let's try this out !

cbeauchesne (approved):

PR #6952, opened 2026-05-15
Add trace filter tests [APMSP-2763]

Eldolfin:
APMSP-2763
APMSP-3048

## Changes

Add system-tests coverage for the CSS V1.2 spec trace filters (filter_tags, filter_tags_regex, ignore_resources).

## Workflow

## Reviewer checklist


utils/_context/_scenarios/__init__.py
```diff
@@ -128,6 +128,144 @@ class _Scenarios:
         scenario_groups=[scenario_groups.appsec],
     )
 
+    trace_stats_computation_trace_filter_reject = EndToEndScenario(
```
cbeauchesne: I'm wondering if parametric test would be a better fit for this use case. WDYT @nccatoni  ?

nccatoni: Looks like it would be

Eldolfin: Noted, I'll change it ty

Eldolfin: It's now parametric tests, which indeed makes much more sense, if you want to take a look again.

manifests/cpp.yml
```diff
@@ -288,6 +288,7 @@ manifest:
   tests/parametric/test_telemetry.py::Test_TelemetrySCAEnvVar::test_telemetry_sca_enabled_propagated: missing_feature  # Created by easy win activation script
   tests/parametric/test_telemetry.py::Test_TelemetrySCAEnvVar::test_telemetry_sca_enabled_propagated_specifics: irrelevant
   tests/parametric/test_telemetry.py::Test_TelemetrySSIConfigs: missing_feature
+  tests/parametric/test_trace_filters.py::Test_Trace_Filters: missing_feature (agent /info trace filters not applied client-side)
```
VianneyRuhlmann: nit: I think the part in parenthesis is suppose to be a reason for the feature not being implemented rather than a description of the feature

tests/parametric/test_trace_filters.py
```diff
@@ -0,0 +1,213 @@
...
+        assert len(test_agent.traces()) == 1, "trace with the required tag should be kept"
+        # Not testing that CSS here so we don't check that we received stats
+
+    @enable_tracestats()
+    @enable_trace_filters(require=["require_tag:true"])
+    def test_trace_filters_require_missing(self, test_agent: TestAgentAPI, test_library: APMLibrary):
+        """A require filter drops traces whose root span lacks the required tag."""
+        with test_library, test_library.dd_start_span(name="web.request"):
+            pass
+        assert not test_agent.traces(), "trace lacking the required tag should be dropped"
```
VianneyRuhlmann, line 87: nit: why not check for stats here ?

tests/parametric/test_trace_filters.py
```diff
@@ -0,0 +1,213 @@
...
+        with (
+            test_library,
+            test_library.dd_start_span(name="web.request", tags=[("require_tag", "something true something")]),
+        ):
+            pass
+        assert len(test_agent.traces()) == 1, "trace should be kept (regex match)"
+
+    @enable_tracestats()
+    @enable_trace_filters(require_regex=["require_tag:.*true.*"])
+    def test_trace_filters_require_regex_missing(self, test_agent: TestAgentAPI, test_library: APMLibrary):
```
VianneyRuhlmann: You're missing the equivalent for reject `test_trace_filters_reject_regex_missing`

VianneyRuhlmann (approved):

tests/parametric/test_trace_filters.py
Eldolfin: These tests are focused on trace filters specifically, so they should not fail if CSS has a problem.

When I tried checking stats here using php it failed for the same reason as why CSS tests are disabled for php:

```
missing_feature (CSS v1.2.0: dd-trace-php ships CSS since v1.19.0, but the sidecar stats exporter retains buckets younger than ~20s (buffer_len=2 * bucket_size=10s) and dd_trace_synchronous_flush passes force=false, so short-lived parametric tests never observe a /v0.6/stats request)
```

cbeauchesne (commented):
From framework usage, all good.

I let someone familiar with the feature validate the test logic and approve the PR

tests/parametric/test_trace_filters.py
```diff
@@ -0,0 +1,229 @@
...
+def enable_trace_filters(
+    reject: list[str] | None = None,
+    require: list[str] | None = None,
+    reject_regex: list[str] | None = None,
+    require_regex: list[str] | None = None,
+    ignore_resources: list[str] | None = None,
+) -> pytest.MarkDecorator:
+    """Configure agent-side trace filters and pin the test agent version.
+
+    Filters are advertised in the test agent's /info response (via DD_AGENT_EXTRA_INFO) so the
```
ichinaski, line 19: Why is this behind an env var? The real agent always exposes them (since 7.75)

manifests/dotnet.yml
```diff
@@ -1055,6 +1055,7 @@ manifest:
   tests/parametric/test_telemetry.py::Test_TelemetrySCAEnvVar: v2.50.0
   tests/parametric/test_telemetry.py::Test_TelemetrySCAEnvVar::test_telemetry_sca_enabled_propagated_specifics: irrelevant
   tests/parametric/test_telemetry.py::Test_TelemetrySSIConfigs: v2.53.0
+  tests/parametric/test_trace_filters.py::Test_Trace_Filters: missing_feature
```
ichinaski, line 1068: I believe dotnet does have support for this feature? https://datadoghq.atlassian.net/wiki/spaces/APM/pages/6378587288/Implementation+Status#6.-Trace-Filters-(root-span%2C-applied-BEFORE-stats)

cc/ @andrewlock could you confirm whether this should be marked as implemented?

Eldolfin: Looking at the test results, I see that 2 tests are failing for dotnet:
`test_trace_filters_ignore_resources` -> surprised this fails, it's a basic case
`test_trace_filters_trim` -> edge case behavior where the tag filter ` reject_tag : true  ` is trimmed to be equivalent to `reject_tag:true`

tests/parametric/test_trace_filters.py
Eldolfin: I'm not sure I understand what you're asking.
To clarify the env var `DD_AGENT_EXTRA_INFO` is used to add the trace filter configuration to the test agent `/info` response, AFAIK there is no other way to configure trace filters in the test agent.

And with the real agent we could've used the official parameters

tests/parametric/test_trace_filters.py
```diff
@@ -0,0 +1,229 @@
...
+    tracer applies them client-side. The agent version is pinned to the minimum that supports
+    client-side stats according to the spec.
+    """
+    agent_version = MIN_AGENT_VERSION_FOR_CSS
```
mabdinur, line 23: Should we hard code the agent version here? It would be nice to have at least one test case that runs against the latest version of the agent

mabdinur (approved):

tests/parametric/test_trace_filters.py
Eldolfin: I'm not sure how I would go about it tbh.
I can't just let the default version of the test agent (`test` won't work because it's not in the `\d+.\d+.\d+` format)

This only seems possible by using the real agent, so with end-to-end tests _(or by using a quick hack where I fetch it from the github release page...)_

These tests were previously end-to-end so if you want I can partially revert my deletion of end-to-end tests ?

tests/parametric/test_trace_filters.py
```diff
@@ -0,0 +1,229 @@
+import json
+import pytest
+
+from .conftest import APMLibrary
+from .test_library_tracestats import MIN_AGENT_VERSION_FOR_CSS, enable_tracestats
```
cbeauchesne: Can you move `MIN_AGENT_VERSION_FOR_CSS` and `enable_tracestats` into `tests/parametric/utils.py` ? (long story short, because of the CI orchestrator that determine which scneario to run, a test file should never import something from another test file.)

cbeauchesne (dismissed):

PR #6954, opened 2026-05-15
feat(parametric): add synchronous /trace/remote-config/apply endpoint

bm1549:
## Motivation

Parametric RC tests call `set_and_wait_rc()`, which waits for the test agent's `APM_TRACING ACKNOWLEDGED`. That ACK fires when the tracer's RC client *validates* a payload, not when its subscribers have *applied* it. Reading tracer state in that gap is racy, and it is the cause of the Python flaky gates on `TestDynamicConfigSamplingRules` (APMAPI-857, APMAPI-1051, APMAPI-1860), which were previously masked with retry loops.

This closes the gap at the source: a synchronous "apply pending RC now" endpoint the test can call after the ACK, so Python reads tracer state only once the new config is live.

## Changes

- New `POST /trace/remote-config/apply` on the Python parametric `apm_test_client`: drains pending RC synchronously (10s timeout → `504`; `500` if the drain raises), plus a `flush_remote_config()` client helper. The helper returns `None` on a missing endpoint (`404`); on any other non-2xx it logs a warning and returns `None`, so a transient transport error or server-side timeout doesn't kill the test.
- `set_and_wait_rc()` calls `flush_remote_config()` unconditionally after the ACK (no per-language branching). It's a no-op on tracers that haven't implemented the endpoint yet, so they keep their existing retry loops.
- Unflags the 3 Python sampling tests that were carrying the race.
- Endpoint-parity tests live in `test_parametric_endpoints.py` (`@features.parametric_endpoint_parity`). Non-Python parametric apps don't implement the endpoint yet, so they're declared `incomplete_test_app` in their manifests.

## Workflow

## Reviewer checklist

### DIFF
--- a/utils/build/docker/python/parametric/apm_test_client/server.py
+++ b/utils/build/docker/python/parametric/apm_test_client/server.py
@@ -9,6 +9,7 @@ from typing import Union
 import logging
 import os
 import enum
+import threading
 from fastapi import FastAPI
 from fastapi import Request
 from fastapi.responses import JSONResponse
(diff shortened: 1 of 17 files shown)


bwoebi:
If you want to point your AI to it for PHP too, `dd_trace_internal_fn("get_loaded_remote_configs")` returns an array with the keys being the _config id_.

PHP has currently some hardcoded hacks over the place in various tests/apply methods to ensure RC data is applied, because it suffers the same problem.

utils/docker_fixtures/_test_clients/_test_client_parametric.py
```diff
@@ -492,6 +492,25 @@ def dd_flush(self) -> bool:
...
+        other non-2xx responses (504 timeout, 5xx error).
+        """
+        resp = self._session.post(
+            self._url("/trace/remote-config/apply"),
+            json={},
+            timeout=timeout,
+        )
+        if resp.status_code == HTTPStatus.NOT_FOUND:
+            return None
+        resp.raise_for_status()
```
mabdinur: Instead of raising an exception here, can we log a descriptive error. There might be cases were the flush succeeded but the connection timed/failed. We shouldn't kill the test prematurely.

tests/parametric/test_remote_config_apply_endpoint.py
```diff
@@ -0,0 +1,103 @@
...
+    non-synchronous drain would then visibly fail rather than being masked by a
+    fast background poll.
+
+    Languages that haven't implemented the endpoint are declared missing_feature
+    in their manifests, so this class only runs against Python today.
+    """
+
+    def test_apply_returns_empty_when_no_rc_received(self, test_agent: TestAgentAPI, test_library: APMLibrary) -> None:
+        """When no RC has been sent, the endpoint returns 200 with applied_configs=[]."""
+        test_agent.clear()
```
mabdinur: Do we need this clear operation? We may lose events queued on start up.

tests/parametric/test_remote_config_apply_endpoint.py
```diff
@@ -0,0 +1,103 @@
...
+    DEFAULT_ENVVARS,
+    _create_rc_config,
+    _set_rc,
+    set_and_wait_rc,
+)
+
+
+@scenarios.parametric
+@features.dynamic_configuration
+class TestRemoteConfigApplyEndpoint:
```
mabdinur: Can we move this test case to: tests/parametric/test_parametric_endpoints.py and add `@features.parametric_endpoint_parity`. We can probably remove `features.dynamic_configuration` feature. It mostly tests the app.

utils/build/docker/python/parametric/apm_test_client/server.py
```diff
@@ -434,6 +435,106 @@ def trace_stats_flush(args: TraceStatsFlushArgs) -> TraceStatsFlushReturn:
...
+@app.post("/trace/remote-config/apply")
+def trace_remote_config_apply(
+    args: TraceRemoteConfigApplyArgs,
+) -> JSONResponse:
+    """Synchronously drain pending Remote Config and return the applied set.
+
+    Contract: docs/parametric/remote-config-apply-contract.md. The drain runs in
+    a worker thread joined with a 10s deadline (504 on timeout).
+    """
+    from ddtrace.internal.remoteconfig.worker import remoteconfig_poller
```
mabdinur, line 466: Do we need to line this import (if so can we add a comment).

tests/parametric/test_remote_config_apply_endpoint.py
mabdinur: ALso claude left this comment: ` test_agent.clear() resets the test agent buffer but does not reset the
  tracer's in-memory _applied_configs (server.py:528). The drain calls
  client.request() which fetches from the now-empty test agent, but
  _applied_configs still contains whatever prior tests in the same container
  session applied. The assertion will fail if any other test in
  TestRemoteConfigApplyEndpoint runs first (e.g.,
  test_apply_returns_applied_config_after_set sets a config). Tests within a
  parametric class share a container.
`

tests/parametric/test_remote_config_apply_endpoint.py
```diff
@@ -0,0 +1,103 @@
...
+        test asserts the documented behavior path by confirming the endpoint
+        URL is reachable and returns within the server-side timeout window
+        under normal load. The full hang path is covered by the unit test in
+        the parametric server itself (TODO when we have one).
+        """
+        test_agent.clear()
+        start = time.monotonic()
+        test_library.flush_remote_config(timeout=10.0)
+        elapsed = time.monotonic() - start
+        assert elapsed < 10.0, f"endpoint took {elapsed:.2f}s, expected <10s under normal load"
```
mabdinur: From claude: ` The timeout=10.0 argument is the HTTP client read timeout, not the server-side drain deadline. The server's 10s timeout is hardcoded and
  cannot be controlled from outside. Under normal load the endpoint returns in milliseconds, so elapsed < 10.0 is a trivially true
  connectivity check -- it does not validate the 504 path at all. If a 504 were returned, raise_for_status() raises before the assertion is
  reached. The test name and docstring overstate what is actually being tested.`

If this test were to fail, an exception would be raised. Instead of using a timeout we should use `pytest.rasies(....)`

docs/parametric/remote-config-apply-contract.md
```diff
@@ -0,0 +1,196 @@
+# Parametric Harness: `/trace/remote-config/apply` Contract
```
mabdinur: This feels like overkill for a new parametric endpoint. A doc string in the parametric parity test class should be enough. I am not against adding this doc but we should simplify the contents. It's a lot to read.

mabdinur (approved):
left some nits and high level suggestions

bm1549:
@bwoebi I'm going to merge this in as-is to get the win for Python, but happy to approve if you introduce one for PHP!

PR #7000, opened 2026-05-21
fix: switch mssql image to ghcr mirror to avoid flaky mcr registry

rochdev:
## Summary

- Replaces `mcr.microsoft.com/mssql/server:2022-latest` with the GHCR mirror `ghcr.io/datadog/system-tests/mcr.microsoft.com/mssql/server:2022-latest`
- The MCR registry has been flaky, causing intermittent test failures; the GHCR mirror follows the same convention used by the existing `mirror-image.yml` workflow

## Test plan

- [ ] Verify mssql-dependent integration tests pass in CI

> Generated by [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/utils/_context/containers.py
+++ b/utils/_context/containers.py
@@ -1373,7 +1373,7 @@ class MsSqlServerContainer(SqlDbTestedContainer):
         }
 
         super().__init__(
-            image_name="mcr.microsoft.com/mssql/server:2022-latest",
+            image_name="ghcr.io/datadog/system-tests/mcr.microsoft.com/mssql/server:2022-latest",
             name="mssql",
             cap_add=["SYS_PTRACE"],
             user="root",


cbeauchesne (changes requested):
Could you run a test PR on dd-trace-js to be sure it's working. I'd be more confortable since some repo are using direclty system-tests `main` branch without any pin

rochdev:
@cbeauchesne https://github.com/DataDog/dd-trace-js/pull/8744

cbeauchesne (approved):

cbeauchesne (changes requested):
side question, are those images reachable from azure / gitlab ?

rochdev:
> side question, are those images reachable from azure / gitlab ?

Yes but that requires authentication. In that sense it's no different than Docker Hub (they have a deal with GitHub but otherwise require authentication elsewhere)

cbeauchesne:
Well, then there are two solutions : 

* either check with dd-trace-php and dd-trace-dotnet to see if they can authenticate their CI to github
* or else we won't be able to use GHCR, but we can do the same and publish it to docker hub. It's already heavily used so it won't change the stability rate.

PR #7006, opened 2026-05-22
Add test to remotely disable Stripe rules

simon-id:
## Changes

Test that remote config is capable of disabling Stripe integrations


manifests/nodejs.yml
```diff
@@ -1043,6 +1043,10 @@ manifest:
         express4: *ref_5_99_0
         express5: *ref_5_99_0
         "*": irrelevant (one weblog is enough coverage)
+  tests/appsec/test_automated_payment_events.py::Test_Automated_Payment_Events_Stripe_RC_disablement:
+    - weblog_declaration:
+        express4: *ref_5_99_0
+        express5: *ref_5_99_0
```
simon-id, line 1049:
```suggestion
        express5: *ref_5_99_0
        "*": irrelevant (one weblog is enough coverage)
```

PR #7013, opened 2026-05-22
feat(nodejs): add base images for Node.js weblogs

rochdev:
## Summary

- Adds `*.base.Dockerfile` files for express4, express5, fastify, express4-typescript, and nextjs weblogs
- Base images use bun for dependency installation with frozen lockfiles
- Adds `docker-bake.hcl` for building all base images at once
- Adds `build_nodejs_base_images.sh` script
- Updates CI (`ci.yml`, `system-tests.yml`) with a `build-nodejs-base-images` label trigger
- Updates `get_pr_merged_labels.sh` to handle the new label
- Main Dockerfiles are simplified to `FROM datadog/system-tests:<weblog>.base-v1`

## Dependencies

- Should be merged after [#7011](https://github.com/DataDog/system-tests/pull/7011) (gzip→zst) since it uses `zstd` in `build.sh`

## Notes

- This PR includes the gzip→zst changes from PR #7011 in `build.sh` for self-consistency. Those lines will need to be removed once #7011 merges.

## Test plan

- [ ] Add the `build-nodejs-base-images` label and verify base images build in CI
- [ ] Verify weblogs using `FROM datadog/system-tests:express4.base-v1` build and run correctly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #7014, opened 2026-05-22
feat(ci): add _build_weblog_images option to skip weblog build jobs

rochdev:
## Summary

- Adds a `_build_weblog_images` parameter to `compute-workflow-parameters.yml`
- When `false`, weblogs skip the dedicated build job and are rebuilt inline during the run job
- For Node.js, this is automatically set based on whether `_build_nodejs_base_images` is enabled
- Adds `build_in_run` field to the `Weblog` data class in `workflow_data.py`

## Dependencies

- Should be merged after [#7013](https://github.com/DataDog/system-tests/pull/7013) (nodejs base images), since it references `inputs._build_nodejs_base_images` in `system-tests.yml`

## Test plan

- [ ] Verify that running without the `build-nodejs-base-images` label skips the build step and weblogs rebuild in-run
- [ ] Verify that other library weblogs still get their own build jobs

### DIFF
--- a/.github/workflows/compute-workflow-parameters.yml
+++ b/.github/workflows/compute-workflow-parameters.yml
@@ -52,6 +52,11 @@ on:
         default: ''
         required: false
         type: string
+      _build_weblog_images:
+        description: "When true, weblogs get a build job to save artifacts. When false, parallel_weblogs is empty and run jobs rebuild from Docker Hub."
+        default: true
+        required: false
+        type: boolean
 
     # Map the workflow outputs to job outputs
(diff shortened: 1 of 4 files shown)


nccatoni (commented):
I'm not sure I understand, when a weblog doesn't get a build job where is the tracer coming from ?

rochdev:
> I'm not sure I understand, when a weblog doesn't get a build job where is the tracer coming from ?

From the artifact.

rochdev:
Superseded by #7130

PR #7034, opened 2026-05-28
fix: remove tests checking telemetry otlp configs which can be sensitive

paullegranddc:
# Motivation

We want to redact these values in libraries. These tests prevent these changes, but changing them outright will break libraries that haven't done the change

### DIFF
--- a/tests/parametric/test_otel_logs.py
+++ b/tests/parametric/test_otel_logs.py
@@ -653,7 +653,7 @@ class Test_FR11_Telemetry:
     """Test OTLP Logs generated via OpenTelemetry API generate telemetry configurations and metrics."""
 
     @pytest.mark.parametrize(
-        ("library_env", "test_agent_otlp_http_port"),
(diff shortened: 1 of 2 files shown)


tests/parametric/test_otel_logs.py
```diff
@@ -653,7 +653,7 @@ class Test_FR11_Telemetry:
     """Test OTLP Logs generated via OpenTelemetry API generate telemetry configurations and metrics."""
 
     @pytest.mark.parametrize(
-        ("library_env", "endpoint_env", "test_agent_otlp_http_port"),
+        ("library_env", "test_agent_otlp_http_port"),
```
zacharycmontoya: Please restore the `endpoint_env` pytest parameter and the value in the corresponding argument, that should allow the test cases to be wired up correctly and pass

bm1549 (approved):

BridgeAR (approved):

zacharycmontoya (approved):

PR #7066, opened 2026-06-03
feat(java): move Jetty download into prebuilt base image

AlexeyKuznetsov-DD:
# What Does This Do

Moves the Jetty distribution download out of per-test image builds and into a single prebuilt, pushed base image (`datadog/system-tests:java-jetty-9.4.58.v20250814.base-v1`), mirroring the PHP approach from #6861. Also bumps Jetty to `9.4.58.v20250814` (latest 9.4.x with security fixes).

- New `jetty-classpath.base.Dockerfile` + `docker-bake.hcl` + `build_java_base_images.sh` build the classpath once (with `curl --retry`).
- Java weblog Dockerfiles (`multialpine`, `multicontainer`, SSI `jetty-app`) now `COPY --from=jetty_classpath /opt/jetty-classpath` and compile `*.java` in-image instead of downloading Jetty at build time.
- Alpine images switched from JRE-only packages to `*-jdk` (so `javac` is available); `openjdk8` already bundles `javac`.
- VM provisioning (`alpine`/`container`/`multi*`) drops the `compile_app.sh` step. The bare-VM `provision_test-app-java.yml` keeps it, now with a `curl`/`wget` retry loop.
- CI wiring for a new `build-java-base-images` label (GitHub Actions + GitLab `get_pr_merged_labels.sh`) and docs.

# Motivation

Many GitLab jobs failed because the Jetty download from Maven Central hit **`429: Too Many Requests`** (confirmed via DDCI job `1736247977`). Downloading Jetty in every weblog build multiplies requests to `repo1.maven.org`; building it once into a base image removes that load from the test path. The version bump to `9.4.58.v20250814` picks up the latest security fixes.

# Additional Notes

- Requires the `build-java-base-images` label so the base image is built and pushed to Docker Hub before tests consume it — ping Reliability & Performance to approve the push.
- Verified locally on `9.4.58.v20250814`: base image builds, `Dockerfile.jdk11` compiles and runs (**HTTP 200 on :18080**, log reports `jetty-9.4.58.v20250814`), `jdk21-alpine` (`openjdk21-jdk`) and `jdk8-alpine` (`openjdk8`) build with working `javac`. `shellcheck` passes.
- The `test-app-java-buildpack` 429 (Gradle/Spring Boot deps) is a separate path not addressed here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,6 +88,7 @@ jobs:
       _build_proxy_image: ${{ contains(github.event.pull_request.labels.*.name, 'build-proxy-image') }}
       _build_python_base_images: ${{ contains(github.event.pull_request.labels.*.name, 'build-python-base-images') }}
       _build_php_base_images: ${{ contains(github.event.pull_request.labels.*.name, 'build-php-base-images') }}
+      _build_java_base_images: ${{ contains(github.event.pull_request.labels.*.name, 'build-java-base-images') }}
       _enable_replay_scenarios: true
       _system_tests_dev_mode: ${{ matrix.version == 'dev' }}
       _system_tests_library_target_branch_map: ${{ needs.compute_libraries_and_scenarios.outputs.target-branch-map }}
--- a/.github/workflows/system-tests.yml
+++ b/.github/workflows/system-tests.yml
@@ -54,6 +54,11 @@ on:
         default: false
         required: false
         type: boolean
+      _build_java_base_images:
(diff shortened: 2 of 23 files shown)


smola (approved):

robertomonteromiguel (changes requested):
This PR should fix this problem: https://github.com/DataDog/system-tests/pull/7075

AlexeyKuznetsov-DD:
Fixed in #7075

PR #7079, opened 2026-06-03
perf(nodejs): reduce Docker image file count with nft static analysis

rochdev:
## Summary

Reduces the file count in Node.js base images using [@vercel/nft](https://github.com/vercel/nft) static analysis. nft traces the import graph from the app entry point and deletes everything in `node_modules` that is statically unreachable, giving ~4,500 files per image vs ~21,600 from a raw `bun install` (−79%).

nft works at two levels simultaneously: it eliminates entire unreachable packages (eslint and its transitive deps, `@types` in JS images, unused AWS SDK v3 packages, etc.) and within kept packages it retains only the specific files that are imported, not the full package tree.

### How it works

A single `nft-prune.mjs` script calls `nodeFileTrace()` from `@vercel/nft`, then deletes every `node_modules` file not in the result. `@vercel/nft` is added as a devDependency so it installs with `bun install` and is then pruned away by its own trace (the app never imports it). App source is `COPY`'d into the base image before `bun install` so nft has an entry point to trace from; both steps run in the same `RUN` layer so deleted files never appear in the image.

The script accepts a `--keep-types` flag used by `express4-typescript`, which preserves `node_modules/@types` — needed because `tsc` scans for type declarations dynamically rather than importing them statically.

### Other changes

- **Next.js build moved into base image** — compiled during base image creation; stable across dd-trace versions since dd-trace is loaded via `NODE_OPTIONS` at runtime.
- **App source now lives in base images** — required for nft tracing; final Dockerfiles no longer need to re-copy source.

> **Note:** New base image tags (v2) need to be pushed to Docker Hub by the R&P team via `#apm-shared-testing` before this PR can merge.

## Test plan

- [ ] CI builds base images via `build-nodejs-base-images` label
- [ ] All five weblogs start correctly with dd-trace loaded
- [ ] express4-typescript compiles successfully (verifies `--keep-types` keeps `@types` and `typescript`)
- [ ] Next.js healthcheck and routes return 200


rochdev:
@cbeauchesne Did you re-ship the image?

cbeauchesne (approved):

PR #7111, opened 2026-06-08
test(debugger): budget limit for capture expressions

tylfin:
## Motivation

Extends the Live Debugger budget tests to ensure that probes configured with **capture expressions** respect the 1 snapshot per second limit, the same budget already enforced for full-snapshot log probes.

## Changes

- New probe fixture `probe_capture_expressions_budgets.json`: a line probe on the `Budgets` method with `captureSnapshot: false` and a `captureExpressions` entry on the `loops` parameter.
- New test `Test_Debugger_Probe_Budgets::test_log_line_capture_expression_budgets`: fires 150 requests and asserts that only 1-20 snapshots carrying capture expressions are emitted (matching the existing `test_log_line_budgets` tolerance for the ~1/s budget).
- Marked the test as a known bug for Java (`DEBUG-5730`) in `manifests/java.yml`: capture-expression probes currently emit on every invocation (150/150) and are not yet rate limited.

## Testing

Ran `DEBUGGER_PROBES_SNAPSHOT` locally for Java:
- Without the manifest entry, the test correctly fails (`got 150 ... should be 1-20`), confirming the budget is bypassed.
- With the `DEBUG-5730` bug declaration, the test is skipped as expected.

### DIFF
--- a/tests/debugger/test_debugger_probe_budgets.py
+++ b/tests/debugger/test_debugger_probe_budgets.py
@@ -89,6 +89,41 @@ class Test_Debugger_Probe_Budgets(debugger.BaseDebuggerTest):
                 f"Expected 1-20 snapshot with captures, got {snapshots_with_captures} in {self.total_request_time} seconds"
             )
 
+    def setup_log_line_capture_expression_budgets(self):
+        self._setup(
+            "probe_capture_expressions_budgets",
+            "/debugger/budgets/150",
+            lines=self.method_and_language_to_line_number("Budgets", self.get_tracer()["language"]),
+        )
+
+    def test_log_line_capture_expression_budgets(self):
+        self._assert()
+        self._validate_snapshots()
+
+        snapshots_with_captures = 0
+        for _id in self.probe_ids:
+            for span in self.probe_snapshots[_id]:
+                captures = span.get("debugger", {}).get("snapshot", {}).get("captures", None)
+                if captures is None:
+                    continue
+
+                # Capture expressions live under captures.lines.{line}.captureExpressions for line probes.
+                has_capture_expressions = any(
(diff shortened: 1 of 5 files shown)


jpbempel (approved):

tests/debugger/test_debugger_probe_budgets.py
```diff
@@ -89,6 +89,41 @@ def test_log_line_budgets(self):
...
+                    for line_data in captures.get("lines", {}).values()
+                )
+                if not has_capture_expressions:
+                    continue
+
+                snapshots_with_captures += 1
+
+            # Probe budgets aren't exact and can take time to be applied, so we allow a range of 1-20 snapshots with
+            # capture expressions for 150 requests, matching the 1/s snapshot limit.
+            assert 1 <= snapshots_with_captures <= 20, (
```
cbeauchesne, line 122: I'm slightly concerned about the stability of this test. From experience, every time we;ve introduced time-based assertions, we've eventually been hit by the fact that we don;t control the underlying machine the test runs on, which introduces a lot of non-determinism.

In the end, we often had to expand the asserted window significantly to reduce flakiness, which defeats the purpose of the assertion.

Let's give this a try, but if it turns out that we get too many failures that cannot be correlated with a real issue, we'll need to rethink this approach.

cbeauchesne (approved):
Approved, with one comment :)

tests/debugger/test_debugger_probe_budgets.py
tylfin: AFAIK the lack of stability for the debugger tests was always a missing waiting signal, we should be fairly deterministic as long as we gate on the proper signals

PR #7116, opened 2026-06-09
APPSEC-63236 RFC-1103: normalized route tests for multi-params-in-segment and optional params

christophe-papazian:
APPSEC-63236

## Summary

- Add `Test_NormalizedRouteMultiParamsInSegment`: validates RFC-1103 rule 5 for two mandatory path parameters in the same URL segment — the tracer must combine them as `{id+format}` using the `+` combining marker.
- Add `Test_NormalizedRouteOptionalParams`: validates optional-element resolution (rules 5 + 6) — `{id+format}` when the optional intra-segment parameter is present, `{id}` when absent, and the two normalized routes are distinct.
- Add the two new endpoints (`/api_security/multi-params-in-segment/{id}-{format}` and `/api_security/optional-params/{id}[-{format}]`) to Flask, FastAPI, Django, and Tornado Python weblogs.
- Document both endpoints in the end-to-end weblog spec.

## Test plan

- [ ] Run `APPSEC_API_SECURITY` scenario against Python Flask/FastAPI/Django/Tornado weblogs and verify the two new test classes pass.
- [ ] Verify `test_optional_routes_are_distinct` confirms the two normalized routes differ.


tests/appsec/api_security_testing/test_normalized_route.py
```diff
@@ -127,8 +124,84 @@ def setup_query_string_not_in_route(self):
...
+    When a route template places two dynamic parameters inside the same
+    path segment (e.g. `/<id>-<format>`), the tracer must collapse them into a
+    single atomic element using the `+` combining marker, in declaration order:
+    `{id+format}`.  The resulting normalized route is deterministic and
+    identical across all weblogs because the parameter names `id` and `format`
+    are fixed in the endpoint declaration.
+    """
+
+    def setup_multi_params_in_segment(self):
+        self.r = weblog.get("/api_security/multi-params-in-segment/123-json")
```
Strech: I have a question, why don't we make it real-world URL, like `123.json` because that's how it will be in real life and `123-json` is a cumbersome to Ruby standard frameworks to parse it, also it require manual intervention to define it (we artificially coding the solution to work)

christophe-papazian: I can change the dash to a point, yes.

Strech (approved):
👏🏼

_P.S Please don't forget to update documentation and all mentions of `{id}-{format}` style_

PR #7100, opened 2026-06-05
Enriching & Enabling OTLP Runtime Metrics for JS

link04:
## Motivation

So that we can test that the work is done correctly when collecting and sending OTLP Runtime Metrics using the DD SDK.

## Changes

Added attributes to the test assertions, removed unexpected metrics and enabled it.

## Workflow

1. ⚠️ Create your PR as draft ⚠️
2. Work on you PR until the CI passes
3. Mark it as ready for review
    * Test logic is modified? -> Get a review from RFC owner.
    * Framework is modified, or non obvious usage of it -> get a review from [R&P team](https://dd.enterprise.slack.com/archives/C025TJ4RZ8X)

:rocket: Once your PR is reviewed and the CI green, you can merge it!

🛟 [#apm-shared-testing](https://dd.enterprise.slack.com/archives/C025TJ4RZ8X) 🛟

## Reviewer checklist

* [ ] Anything but `tests/` or `manifests/` is modified ? I have the approval from [R&P team](https://dd.enterprise.slack.com/archives/C025TJ4RZ8X)
* [ ] A docker base image is modified?
    * [ ] the relevant `build-XXX-image` label is present
* [ ] A scenario is added, removed or renamed?
    * [ ] Get a review from [R&P team](https://dd.enterprise.slack.com/archives/C025TJ4RZ8X)

PR #7119, opened 2026-06-09
Fix flaky v8js.resource.type validation in OTLP runtime metrics test

link04:
## Summary

Follow-up to #7100 (Enriching & Enabling OTLP Runtime Metrics for JS). The Node.js OTLP runtime metrics test enforced a curated allow-list for `v8js.resource.type`, which flaked in the dd-trace-js nightly when an active HTTP keep-alive socket reported `TCPSocketWrap`.

- `v8js.resource.type` values come from each handle/request class's C++ `MemoryInfoName()` override (surfaced by `process.getActiveResourcesInfo()`), **not** the `async_wrap` provider enum — e.g. `PROVIDER_TCPWRAP` surfaces as `"TCPSocketWrap"`, not `"TCPWrap"`. Node's docs state the set "can change in any Node.js release", and it is per-scrape non-deterministic (depends on which handles are live). So it is now validated for **key presence only**, not value.
- Kept value validation for the closed/bounded attributes: `nodejs.eventloop.state`, `v8js.gc.type`, and the `v8js.heap.space.name` superset (deterministic per Node version — only changes on a V8 version bump). Added the missing `code_range_space` and clarified the comments.

This mirrors OpenTelemetry's own `instrumentation-runtime-node` posture: it asserts known values are *present* and never rejects unknown V8/Node values (see open-telemetry/semantic-conventions#2832).

## Test plan

- [x] `TEST_LIBRARY=nodejs ./run.sh OTLP_RUNTIME_METRICS tests/test_otlp_runtime_metrics.py` passes locally (`2 passed`) against dd-trace-js master (the merged #8357 code).
- [x] `format.sh` mypy/ruff/yaml clean.
- [ ] Re-run the dd-trace-js cross-repo nightly against this branch to confirm the `TCPSocketWrap` failure is gone.

### DIFF
--- a/tests/test_otlp_runtime_metrics.py
+++ b/tests/test_otlp_runtime_metrics.py
@@ -106,30 +106,36 @@ EXPECTED_METRICS: dict[str, dict[str, dict[str, list[str]]]] = {
     },
 }
 
-# Valid value domains for attributes. For closed enums (jvm.memory.type, jvm.thread.*,
-# nodejs.eventloop.state, v8js.gc.type) these are exhaustive. For open-ended attributes
-# (pool names, GC names, V8 heap space names, libuv resource types) these are supersets
-# covering all known implementations — the assertion is that observed values fall within
-# the known universe.
+# Valid value domains for attributes. Closed enums (jvm.memory.type, jvm.thread.*,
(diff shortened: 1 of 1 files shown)


link04:
Superseded by #7120 (same fix on branch maximo/fix-flake-otlp-runtime-metrics).

PR #7120, opened 2026-06-09
Fix flaky v8js.resource.type validation in OTLP runtime metrics test

link04:
## Summary

Follow-up to #7100 (Enriching & Enabling OTLP Runtime Metrics for JS). The Node.js OTLP runtime metrics test enforced a curated allow-list for `v8js.resource.type`, which flaked in the dd-trace-js nightly when an active HTTP keep-alive socket reported `TCPSocketWrap`.

- `v8js.resource.type` values come from each handle/request class's C++ `MemoryInfoName()` override (surfaced by `process.getActiveResourcesInfo()`), **not** the `async_wrap` provider enum — e.g. `PROVIDER_TCPWRAP` surfaces as `"TCPSocketWrap"`, not `"TCPWrap"`. Node's docs state the set "can change in any Node.js release", and it is per-scrape non-deterministic (depends on which handles are live). So it is now validated for **key presence only**, not value.
- Kept value validation for the closed/bounded attributes: `nodejs.eventloop.state`, `v8js.gc.type`, and the `v8js.heap.space.name` superset (deterministic per Node version — only changes on a V8 version bump). Added the missing `code_range_space` and clarified the comments.

This mirrors OpenTelemetry's own `instrumentation-runtime-node` posture: it asserts known values are *present* and never rejects unknown V8/Node values (see open-telemetry/semantic-conventions#2832).

## Test plan

- [x] `TEST_LIBRARY=nodejs ./run.sh OTLP_RUNTIME_METRICS tests/test_otlp_runtime_metrics.py` passes locally (`2 passed`) against dd-trace-js master (the merged #8357 code).
- [x] `format.sh` mypy/ruff/yaml clean.
- [ ] Re-run the dd-trace-js cross-repo nightly against this branch to confirm the `TCPSocketWrap` failure is gone.

### DIFF
--- a/tests/test_otlp_runtime_metrics.py
+++ b/tests/test_otlp_runtime_metrics.py
@@ -6,19 +6,26 @@ jvm.*, go.*, v8js.*, etc.) instead of DD-proprietary naming (runtime.dotnet.*,
 runtime.go.*, runtime.node.*, etc.).
 """
 
+from typing import TypedDict
+
 from utils import context, features, interfaces, scenarios, weblog
 
 
-# Maps each expected metric to its attribute constraints:
-#   "all"  — attribute keys that must appear on every emitted data point.
-#   "some" — attribute keys that must appear on at least one data point.
-#
-# Use "some" when a metric emits both per-dimension points (which carry the attribute)
-# and aggregate rollup points (which don't). For example, jvm.memory.used emits one
-# point per memory pool (with jvm.memory.pool.name) as well as heap/non-heap totals
-# (without it), so pool.name goes in "some" while jvm.memory.type, which is present
-# on every point, goes in "all".
-EXPECTED_METRICS: dict[str, dict[str, dict[str, list[str]]]] = {
+class MetricConstraints(TypedDict, total=False):
+    """Attribute constraints for an expected metric.
+
+    all: keys required on every data point.
+    some: keys required on at least one data point.
+    present_values: attribute -> values that must each appear on at least one data point.
+    """
+
+    all: list[str]
+    some: list[str]
+    present_values: dict[str, list[str]]
+
+
+# Maps each expected metric to its attribute constraints (see MetricConstraints).
+EXPECTED_METRICS: dict[str, dict[str, MetricConstraints]] = {
     "dotnet": {
         "dotnet.assembly.count": {"all": []},
         "dotnet.exceptions": {"all": []},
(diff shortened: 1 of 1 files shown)


tests/test_otlp_runtime_metrics.py
```diff
@@ -106,30 +106,36 @@
...
+        # Closed in dd-trace-js: GC_ATTR_BY_KIND maps perf_hooks GC kinds to these four OTel values.
         # Kind 2 (V8 MinorMarkSweep on Node 20+) is mapped to "minor" upstream.
         "v8js.gc.type": frozenset({"minor", "major", "incremental", "weakcb"}),
-        # V8 heap space names: OTel well-known set plus additional spaces V8 exposes in
-        # Node 18+ (read_only, *_large_object) and Node 20+ multi-isolate/sandbox spaces.
+        # V8 heap space names (best-effort superset). dd-trace-js passes space.space_name from
+        # v8.getHeapSpaceStatistics() through unchanged, so the set tracks what V8 emits across
+        # supported Node versions. Per the OTel spec custom values are also allowed, so add new
+        # entries here when V8 introduces a new space rather than treating it as a hard regression.
         "v8js.heap.space.name": frozenset(
```
bm1549, line 134: What's the reason to re-order these? I don't notice any logical/alphabetical benefits

tests/test_otlp_runtime_metrics.py
```diff
@@ -138,28 +144,13 @@
                 "shared_trusted_large_object_space",
             }
         ),
-        # libuv handle types reported by process.getActiveResourcesInfo(); superset across Node 18+.
-        "v8js.resource.type": frozenset(
```
bm1549, line 142: This was a nice guard to have - @BridgeAR is it possible to get an enumerated list from Node.js here?

tests/test_otlp_runtime_metrics.py
link04: No specific reason, just got moved around while doing tests locally.

tests/test_otlp_runtime_metrics.py
link04: Checked the Node source and there's no enumerable list that matches `getActiveResourcesInfo()`. The strings come from each class's C++ `MemoryInfoName()` (e.g. `TCPSocketWrap`), not the async_wrap provider enum (which only has the uppercase TCPWRAP), the live set changes per scrape, and the docs say it can change in any release. So instead of a reject-unknowns allow-list I assert `v8js.resource.active` always emits `TCPServerWrap` (our weblog's listening server) — a real guard that won't flake when Node adds/renames a handle type. 

@BridgeAR lmk if there's a better Node-side hook here.

bm1549 (approved):

PR #7146, opened 2026-06-16
Add EVP flagevaluation system tests

leoromanovsky:
## Motivation

Server-side FFE EVP flagevaluation needs system-test coverage for the worker-visible event contract and SDK-side aggregation behavior before language-by-language rollout.

## Changes

- Adds `tests/ffe/test_flag_eval_evp.py` for basic EVP emission, repeated-count aggregation, bounded context, runtime default, load aggregation, bounded burst aggregation, high-cardinality aggregation, and production-cap degradation.
- Adds batched `/ffe` request support to weblog fixtures so degradation is exercised with 10,050 evaluations without making the test an HTTP throughput benchmark.
- Adds `feature_flags_evp_flagevaluation` as a distinct feature marker from dynamic evaluation, exposures, and OTel evaluation metrics.
- Wires manifests so SDK languages keep this new test disabled for first review with `missing_feature (FFL-2446)`; non-applicable proxy/lambda/OTel manifests are marked `irrelevant`.

## Decisions

- Tests validate visible EVP fields from `/api/v2/flagevaluations` without vendoring dd-source schemas.
- OpenFeature `reason` is neither accepted in payloads nor used as an aggregation dimension.
- `targeting_rule` is expected only when real targeting-rule metadata exists; current no-rule fixtures assert it is omitted.
- Degradation is reached naturally by one request with 10,050 distinct evaluations for one flag; degraded events must omit context and targeting key while preserving visible flag, variant, allocation, and count fields.
- System-tests do not force private SDK queue caps. The bounded burst and degradation tests exercise async handoff, aggregation, and degradation end-to-end; SDK-internal queue-overflow accounting remains SDK unit-test territory.
- Language enablement happens in separate draft sibling PRs after each SDK has local passing evidence.

## Evidence

Base PR checks on current head `2e7ad616bc525ee76698f34a686da549701b1914`:

```bash
./run.sh TEST_THE_TEST tests/test_the_test/test_conventions.py::test_utils
```

Result: `1 passed in 0.74s`.

Current manifest state:

- `golang`, `nodejs`, `python`, `ruby`, `java`, `.NET`, `php`: `tests/ffe/test_flag_eval_evp.py: missing_feature (FFL-2446)`.
- Proxy, lambda, and OTel-only manifests are marked `irrelevant` where the server-side EVP weblog scenario does not apply.

Validated sibling evidence includes PHP #7160: `./run.sh FEATURE_FLAGGING_AND_EXPERIMENTATION --library php --weblog apache-mod-8.2 -k "test_flag_eval_evp"` - PASS, 8 passed, 2577 deselected in 63.53s. Degradation payload from that run: `events=10001`, `total=10050`, `full_total=10000`, `degraded_count=1`, `degraded_total=50`.


manifests/ruby.yml
```diff
@@ -1886,6 +1886,7 @@ manifest:
     - weblog_declaration:
         "*": irrelevant
         rails72: v2.23.0-dev
+  tests/ffe/test_flag_eval_evp.py: missing_feature (FFL-2446)
```
sameerank, line 1889: I see you put `irrelevant` in all the lambda manifests. Should there be one in `ruby_lambda.yml` too?

utils/_features.py
```diff
@@ -2705,6 +2705,14 @@ def feature_flags_eval_metrics(test_object):
         """
         return _mark_test_object(test_object, feature_id=548, owner=_Owner.ffe)
 
+    @staticmethod
+    def feature_flags_evp_flagevaluation(test_object):
+        """Feature Flags EVP Flagevaluation
+
+        https://feature-parity.us1.prod.dog/#/?feature=548
```
sameerank: I wonder if this should have its own feature id in the parity dashboard

The one it currently links to was meant for OTEL and indicates that all the languages are already implemented https://feature-parity.us1.prod.dog/?products=14&feature=548&runDateFilter=7d which isn't true for this EVP direction

tests/ffe/test_flag_eval_evp.py
```diff
@@ -0,0 +1,521 @@
...
+        for index, response in enumerate(self.responses):
+            assert response.status_code == 200, f"Request {index + 1} failed: {response.text}"
+
+        wait_for_evp_flagevaluation_event(self.flag_key)
+        events = find_evp_flagevaluation_events(self.flag_key)
+        assert events, f"Expected EVP flagevaluation events for flag {self.flag_key}"
+
+        degraded_events = [event for _, event in events if "context" not in event and "targeting_key" not in event]
+        if not degraded_events:
+            pytest.skip("No SDK/test cap override is available to force degraded EVP flagevaluation buckets")
```
sameerank: Is there a plan to not skip this? Otherwise, this might get forgotten here

tests/ffe/test_flag_eval_evp.py
```diff
@@ -0,0 +1,521 @@
...
+
+RC_PRODUCT = "FFE_FLAGS"
+RC_PATH = f"datadog/2/{RC_PRODUCT}"
+EVP_FLAGEVALUATIONS_PATH = "/api/v2/flagevaluations"
+EVP_WAIT_TIMEOUT_SECONDS = 30
+
+JSON = dict[str, Any]
+
+
+def make_ufc_fixture(
```
sameerank: This method is also in 

https://github.com/DataDog/system-tests/blob/68ef6f9007041d5889a6d68049cbd2317ba1c45b/tests/ffe/test_exposures.py#L449

and

https://github.com/DataDog/system-tests/blob/68ef6f9007041d5889a6d68049cbd2317ba1c45b/tests/ffe/test_flag_eval_metrics.py#L16

With a 3rd copy here, it might be time to refactor to a shared method, e.g. in a `tests/ffe/_fixtures.py` file maybe. Also fine to leave as-is if the methods are sufficiently different.

sameerank (approved):

utils/_features.py
leoromanovsky: Ah thanks yes good idea; I went into that dashboard to register the new env vars (https://github.com/DataDog/dd-trace-go/pull/4886/changes#diff-0bd192ef12ad91c9074d8742d18605bf312b6651289121ddd14ad05cff50a472) but forgot to do this.

leoromanovsky: I believe its this one: https://feature-parity.us1.prod.dog/#/?feature=540

tests/ffe/test_flag_eval_evp.py
leoromanovsky: I noticed that too; thanks for calling it out.

tests/ffe/test_flag_eval_evp.py
leoromanovsky: Thanks for mentioning! I needed to work through this; got the system test to generate high volume traffic:

```
def setup_ffe_evp_flagevaluation_degradation(self) -> None:
        config_id = "ffe-evp-degradation"
        self.flag_key = "evp-degradation-flag"
        self.eval_count = EVP_FULL_TIER_PER_FLAG_CAP + EVP_DEGRADATION_OVERFLOW_EVALS
        rc.tracer_rc_state.reset().set_config(f"{RC_PATH}/{config_id}/config", make_ufc_fixture(self.flag_key)).apply()

        with ThreadPoolExecutor(max_workers=64) as executor:
            self.responses = list(
                executor.map(
                    lambda index: evaluate_flag(
                        self.flag_key,
                        targeting_key=f"evp-degradation-user-{index}",
                        attributes={},
                    ),
                    range(self.eval_count),
                )
            )
```

nccatoni (approved):

PR #7169, opened 2026-06-18
test(parametric): make OTel span assertions v0.4/v1-agnostic.

AlexeyKuznetsov-DD:
# What Does This Do

Makes the parametric OTel span and span-event assertions protocol-agnostic for trace payloads that may arrive in either `v0.4`-style or `v1.0`-style shapes.

- Adds `is_same_boolean(..., is_otel_boolean=True)` support for comparing boolean values represented as `"true"`/`"false"`, native `True`/`False`, or OTel numeric `1`/`0`.
- Updates OTel span tests to read boolean attributes from merged `meta`/`metrics` data and to tolerate empty optional `tracestate`/`attributes` fields.
- Uses `retrieve_span_events()` in span event tests instead of directly parsing `span["meta"]["events"]`.
- Extends native span-event attribute flattening so protocol `v1` array attributes become plain Python lists.
- Adds `TEST_THE_TEST` coverage for the boolean comparison helper.

# Motivation

`dd-trace-java` protocol `v1` span events serialize some OTel data differently from older trace formats. These changes keep the shared parametric assertions focused on behavior instead of the exact wire-format representation used by a tracer/protocol combination.


AlexeyKuznetsov-DD:
Implemented in other PRs: #7403 & #7404

PR #6464, opened 2026-03-10
Special weblogs infra for go proxies

e-n-0:
## Motivation

The Envoy and HAProxy proxy weblogs (Go security processor, via External Processing / Stream Processing Offload) previously required their own dedicated end-to-end scenarios (`GO_PROXIES_DEFAULT`, `GO_PROXIES_APPSEC_BLOCKING`). That duplicated scenario wiring, kept proxy test activation locked inside scenario-specific decorators, and made it impossible to exercise the proxies against the same scenarios already used for every other weblog.

This PR turns the proxies into **first-class end-to-end weblogs** so they run under the standard scenarios (e.g. `APPSEC_BLOCKING`, `APPSEC_BLOCKING_FULL_DENYLIST`) instead of bespoke ones, with test activation driven by manifests like every other weblog.

## Changes

**Weblog infrastructure refactor** — `utils/_context/weblog_infrastructure.py` (single module) → `utils/_context/weblog_infrastructure/` package:
- `base.py` — abstract `WeblogInfra` / `EndToEndInfra` interfaces + shared container type aliases
- `library_end_to_end.py` — `LibraryEndToEndInfra` for standard tracer weblogs
- `go_proxies.py` — `GoProxiesEndToEndInfra` for Envoy/HAProxy proxies + the `PROXY_WEBLOGS` registry
- `selector.py` — `EndToEndWeblogInfra` picks the library vs proxy infra from the weblog variant (incl. replay-mode discovery)

**Scenario model**
- Removed the dedicated `GoProxiesScenario`, the `GO_PROXIES_DEFAULT` / `GO_PROXIES_APPSEC_BLOCKING` scenarios, and the `go_proxies` scenario group.
- Proxies now run under standard scenarios; `utils/scripts/ci_orchestrators/workflow_data.py` gates them via `PROXY_WEBLOGS` + `PROXY_SUPPORTED_SCENARIOS`.

**Test activation** — moved proxy-specific deactivations out of `@scenarios.go_proxies_*` decorators and into `manifests/envoy.yml` / `manifests/haproxy.yml` (manifest-first, per repo conventions).

**Framework**
- `EndToEndScenario.weblog_container` now resolves to the *library* container (proxies have distinct HTTP vs library containers); added `get_image_list` aggregation.
- `ExternalProcessingContainer` / `StreamProcessingOffloadContainer` gained `trace_agent_port` + `warmup_request` so they fit the standard e2e lifecycle.
[...]

PR #7179, opened 2026-06-19
Use weblog infra for haproxy and envoy integrations

cbeauchesne:
## Motivation

Will replace #6464 

## Workflow

## Reviewer checklist


manifests/haproxy.yml
e-n-0, line 1: I know this isn't finished but I'm not a big fan of removing the specific manifests for envoy/haproxy. I don't want the golang manifest to be the source of truth for those integrations as they behave differently. Also there is difference in some tests/features between haproxy and envoy, even if they shares the same go tracer (different platforms limitations).
I also want to add that there is "sub integration" of proxies, like for envoy, we have the kubernetes gateways like istio or envoy-gateway that use the same envoy security processor (same deployed container image between those sub integrations) but can have also different supported/unsupported tests due to the platform.

Overall what I wanted to do in the previous PR was to do:
- Integrations (envoy, haproxy)
- "weblogs" (envoy proxy, envoy-gateway, istio // haproxy)

cbeauchesne: See with Falvien and Eliott : we'll move forward with a single manifest.

If it turns out it's a pain to use, we'll consider a splitted one.

kakkoyun (approved):

utils/_context/_scenarios/endtoend.py
```diff
@@ -528,6 +527,6 @@ def get_junit_properties(self) -> dict[str, str]:
         result["dd_tags[systest.suite.context.library.name]"] = self.library.name
         result["dd_tags[systest.suite.context.library.version]"] = self.library.version
         result["dd_tags[systest.suite.context.weblog_variant]"] = self.weblog_variant
-        result["dd_tags[systest.suite.context.appsec_rules_file]"] = self.weblog_container.appsec_rules_file or ""
+        result["dd_tags[systest.suite.context.appsec_rules_file]"] = self.weblog_infra.appsec_rules_file or ""
```
eliottness: With `appsec_rules_file` moving onto `weblog_infra` and the scenario no longer exposing its own `appsec_rules_file` property, does `context.appsec_rules_file` still behave the same? It reads through `_get_scenario_property("appsec_rules_file", "")`, which returns the `""` default when the scenario has no such attribute. So for a scenario with no custom rules it looks like it now returns `""` instead of `None`, which flips `context.appsec_rules_file is not None` from False to True.

The case I could trace is Java's `Test_RemoteConfigurationUpdateSequenceASMDD::test_tracer_update_sequence`, which runs in `REMOTE_CONFIG_MOCKED_BACKEND_ASM_DD` (where `DD_APPSEC_RULES` is `None`). With the old `None` it ran; with `""` the `@irrelevant(context.appsec_rules_file is not None)` guard would mark it irrelevant and skip it. golang CI wouldn't surface this since it only shows up on Java. Would adding a passthrough property back on the scenario keep the old behavior?

eliottness (commented):
Went through the go-proxies revamp. It reads cleanly, and the wiring between `SYSTEM_TESTS_WEBLOG`, the conftest weblog resolution, the artifact names, and `run.sh` all lines up. Most of my notes are small. The `appsec_rules_file` one is the only thing I'd look at before merge, since it touches shared scenario code rather than just the proxy path. Specifics are inline.

On the CI ordering from the description: dd-trace-go #4964 checks out system-tests at a ref that resolves to `main` on `pull_request` events, so it looks like this PR needs to land in `main` first before #4964's envoy/haproxy jobs can find the weblog variants. Does that match your rollout plan? I left a separate question on #4964 about the nightly image tag.

One non-blocking thing I couldn't anchor to a changed line: `_get_weblog_system_info` calls `weblog_container.exec_run`, which resolves to the http_container. For proxy weblogs that container isn't started, so it may throw and log a traceback plus "can't get weblog system info" on every envoy/haproxy run. It's caught by `except BaseException`, so non-fatal, just noisy.

utils/_context/weblog_infrastructure.py
```diff
@@ -67,23 +96,122 @@ def __init__(
...
+    def library_container(self) -> TestedContainer:
+        """Container whose library version and metadata represent the tested library.
+
+        In go-proxies mode this is the security processor, not the HTTP entry point.
+        """
+        if self._is_proxy_weblog:
+            return self._processor_container
+        return self.http_container
+
+    def get_image_list(self, library: str, weblog: str) -> list[str]:
```
eliottness, line 191: `get_image_list` gets called from the pull-images path (`get-image-list.py`) before `configure()` runs, so `_activate_proxy_weblog()` hasn't happened yet and `get_containers()` still returns the http_container. For envoy/haproxy that means this returns `[]`, so the envoy/haproxy base images and the dummy server don't make it into the pre-pull list. That matches how GoProxiesScenario worked before, so it may be intended, but would keying off the `weblog` argument here instead of the container state make the pre-pull list complete?

tests/test_semantic_conventions.py
```diff
@@ -287,13 +284,16 @@ def validator(span: DataDogLibrarySpan):
             if span.get("type") != "web":  # do nothing if is not web related
                 return
 
+            logger.info(f"span mane is {span['name']}")
```
eliottness: Are these two `logger.info` lines (this one and the `actual_component` one a few lines down) meant to stay in? They look like leftover debugging, and this one has a small typo ("span mane"). They'd log for every span on each `Test_Meta` run.

utils/_context/weblog_infrastructure.py
```diff
@@ -32,6 +41,23 @@ class EndToEndWeblogInfra(WeblogInfra):
     This class is meant to work with EndToEndScenario
     """
 
+    _is_proxy_weblog: bool
```
eliottness: Minor: is this `_is_proxy_weblog: bool` class annotation still needed? There's a `@property _is_proxy_weblog` further down that shadows it, so this line looks dead.

tests/test_semantic_conventions.py
cbeauchesne: In term of log in debugging, if it has been useful once, it means it'll be usefull later 😄   

I've corrected the typo.

utils/_context/weblog_infrastructure.py
cbeauchesne: good catch

utils/_context/_scenarios/endtoend.py
cbeauchesne: fixed

PR #7225, opened 2026-06-27
fix(weblog,parametric): make Node.js OTel apps work with dd-trace v6's optional @opentelemetry/api peer

BridgeAR:
## Summary

[dd-trace-js#9077](https://github.com/DataDog/dd-trace-js/pull/9077) makes `@opentelemetry/api` an optional peer dependency in v6: dd-trace no longer bundles its own copy and instead shares the single copy the application installs. `express4-typescript` and `nextjs` never declared it.

The `OTLP_RUNTIME_METRICS` scenario runs these weblogs with `DD_METRICS_OTEL_ENABLED=true` and asserts dd-trace emits OTel-native runtime metric names (`nodejs.eventloop.delay.max`, ...) and none of the DD-proprietary ones (`runtime.node.event_loop.delay.max`, ...). On v6 the API can no longer be resolved, so dd-trace disables OTel metrics and falls back to the DD-proprietary names, failing both assertions. Adding the dependency restores the OTel metrics pipeline.

## Why

`^1.9.0`, matching the other Node.js weblogs. The open range is deliberate: it lets these jobs surface a future `@opentelemetry/api` release that drifts outside the range a dd-trace major supports, rather than masking that behind a pin.

## Test plan

- [ ] `OTLP_RUNTIME_METRICS` End-to-end jobs for `express4-typescript` and `nextjs` go green against dd-trace-js#9077 (both currently fail on `test_otel_metrics_are_present_and_attributed` and `test_dd_metrics_are_absent`).

### DIFF
--- a/utils/build/docker/nodejs/express4-typescript/bun.lock
+++ b/utils/build/docker/nodejs/express4-typescript/bun.lock
@@ -5,6 +5,7 @@
     "": {
       "name": "express4-typescript",
       "dependencies": {
+        "@opentelemetry/api": "^1.9.0",
         "@types/express-session": "^1.18.1",
         "@types/ldapjs": "^3.0.6",
         "@types/passport": "^1.0.17",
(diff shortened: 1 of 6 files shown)


mabdinur (approved):
We just need to address the bot comments but the motivation looks good to me.

PR #7230, opened 2026-06-29
[rust] resolve otel dependencies versions dynamically

iunanua:
## Motivation

`opentelemetry` crates versions were hardcoded in the `Cargo.toml` of the parametric app which prevented us from testing `datadog-opentelemetry` depending on different versions of opentelemetry.

## Changes

Now, `opentelemetry` versions are managed by `install_ddtrace.sh` during the docker build. It resolves and reads the `opentelemetry` versions `datadog-opentelemetry` pulls in and updates the versions included in the Cargo.toml


utils/build/docker/rust/parametric/Cargo.lock
```diff
@@ -328,7 +335,7 @@ dependencies = [
 
 [[package]]
 name = "datadog-opentelemetry"
-version = "0.3.2"
+version = "0.4.0"
```
ekump, line 338: non-blocking: I know the lock file gets regenerated before tests run, but it feels slightly odd that we commit a lock file that won't actually compile.

ekump (approved):

PR #7231, opened 2026-06-29
Fix flaky runtime-metrics tests with condition-based collection wait

bm1549:
## Motivation

`Test_Config_RuntimeMetrics_Enabled[*]` and `Test_OtlpRuntimeMetrics.test_otel_metrics_are_present_and_attributed[*]` are flaky: they pass overall but get flagged on retry in the Flaky Test Registry (owned by `@DataDog/apm-sdk-capabilities`).

The cause is timing. These tests assert that runtime metrics are present, but metrics are emitted on an interval and the `RUNTIME_METRICS_ENABLED` / `OTLP_RUNTIME_METRICS` scenarios collect them in a fixed window. When the window closes before a metric arrives, the assertion fails, then passes on the next run.

## Changes

Wait for the runtime metrics to reach the agent before the scenario tears down, instead of relying on the fixed window. The wait lives in each test's `setup_` method (a small `wait_for_runtime_metrics` helper), so it runs while the weblog is still up and returns as soon as the metrics arrive:

- `Test_Config_RuntimeMetrics_Enabled[*]` waits until a `runtime.*` / `jvm.*` metric is collected.
- `Test_OtlpRuntimeMetrics` waits until every expected metric for the library is collected.

Only `tests/` changes; no scenario or framework changes.

## Workflow

## Reviewer checklist

### DIFF
--- a/tests/test_config_consistency.py
+++ b/tests/test_config_consistency.py
@@ -662,6 +662,7 @@ class Test_Config_RuntimeMetrics_Enabled:
 
     def setup_main(self):
         self.req = weblog.get("/")
+        wait_for_runtime_metrics()
 
     def test_main(self):
         assert self.req.status_code == 200
@@ -696,6 +697,7 @@ class Test_Config_RuntimeMetrics_Enabled_WithRuntimeId:
 
     def setup_main(self):
         self.req = weblog.get("/")
+        wait_for_runtime_metrics()
 
     def test_main(self):
         assert self.req.status_code == 200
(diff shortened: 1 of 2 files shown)


link04 (approved):

utils/_context/_scenarios/endtoend.py
```diff
@@ -481,6 +492,21 @@ def _wait_interface(self, interface: ProxyBasedInterfaceValidator, timeout: int)
 
         interface.wait(timeout)
 
+    def _wait_for_metrics(self, prefixes: tuple[str, ...]):
```
cbeauchesne: I feel that logic would be better placed in the setup function, the logic will be closer to the test method, and we won't add complexity to the scenario API, WDYT ?

bm1549: This is how I tried to design this from the beginning, but Claude and Codex review fought me hard on it because of [these rules](https://github.com/DataDog/system-tests/blob/087766177a04489b7be05e549065770d2233d481/.cursor/rules/pr-review.mdc#L26-L34) so I decided to change it up to match the expectation

I'll make the switch to them since imo it is way cleaner

```
## 4. Setup Methods Must Not Fail

`setup_*` methods must never contain assertions, raise exceptions, or call code that can fail. They exist only for observability (trigger trace generation). If one fails, the entire scenario stops. Assertions belong in `test_*` methods.

- Ref: [add-new-test.md](mdc:docs/edit/add-new-test.md), [test-flow.md](mdc:docs/understand/test-flow.md)

## 5. Use Scenario-Level Timeouts, Not Test-Level Sleeps

Do not add `time.sleep()` or explicit waits inside `setup_` method in end-to-end scenarios. If a longer wait is needed, use the scenario parameters `agent_interface_timeout` or `library_interface_timeout` defined in `utils/_context/_scenarios/endtoend.py`.
```

PR #7257, opened 2026-07-03
[python] PoC: AI Guard test under INTEGRATION_FRAMEWORKS

avara1986:
## What

Proof-of-concept exploring whether AI Guard coverage can be consolidated under the `INTEGRATION_FRAMEWORKS` scenario instead of the dedicated `AI_GUARD` scenario.

Adds `tests/integration_frameworks/llm/openai/test_openai_ai_guard.py`, which drives AI Guard evaluation through the OpenAI framework weblog using the `test_client`/`test_agent` fixtures, and asserts on the `ai_guard` span (`meta.ai_guard.*`, `meta_struct.ai_guard`) — reusing the payloads and span shape from `tests/ai_guard/test_ai_guard_sdk.py`.

## Status: DRAFT / investigation

This is a scaffold for the investigation tracked in APPSEC-68977. It is **not runnable yet** — it depends on prerequisites still to be validated:
- an `/ai_guard/evaluate` endpoint in `utils/build/docker/python/openai_app/openai.py` that calls the tracer's AI Guard client;
- the test-agent VCR proxy serving an `/vcr/aiguard` upstream (or reusing the standalone `VCRCassettesContainer`);
- `DD_AI_GUARD_ENABLED` / `DD_AI_GUARD_ENDPOINT` wired into `IntegrationFrameworksScenario`.

## Jira

Investigation task: https://datadoghq.atlassian.net/browse/APPSEC-68977 (under epic APPSEC-68823)


tests/integration_frameworks/llm/openai/test_openai_ai_guard.py
```diff
@@ -0,0 +1,118 @@
...
+                "POST",
+                "/chat/completions",
+                dict(
+                    model="gpt-4o-mini",
+                    messages=[{"role": "user", "content": "Tell me a short story about a robot."}],
+                    parameters=dict(max_tokens=35, stream=True),
+                ),
+            )
+
+        guard_spans = _ai_guard_spans(test_agent.wait_for_num_traces(num=1))
```
avara1986:
```suggestion
        guard_spans = _ai_guard_spans(test_agent.wait_for_num_traces(num=2)
```

cbeauchesne (commented):
Framwork usage : all good

But I let someone familiar with the feature to do the real review.

tests/integration_frameworks/llm/openai/test_openai_ai_guard.py
```diff
@@ -0,0 +1,191 @@
...
+                "/chat/completions",
+                dict(
+                    model="gpt-4o-mini",
+                    messages=[{"role": "user", "content": "Tell me a short story about a robot."}],
+                    parameters=dict(max_tokens=35, stream=True),
+                ),
+            )
+
+        guard_spans = _wait_for_ai_guard_spans(test_agent)
+        assert guard_spans, "expected an ai_guard span from the after-model (streamed) evaluation"
```
christophe-papazian: **Finding 1 (High) — `test_after_model_validation` is a false positive.**

`_wait_for_ai_guard_spans(test_agent)` is called with no `target`, so it matches *any* `ai_guard` span. Every AI-Guard-enabled call already emits a before-model `ai_guard.target == "prompt"` span, so this assertion passes even if the streamed after-model evaluation never runs.

**Confirming evidence — the cassettes prove the after-model eval was never recorded:**
- The **before-model** (non-stream) test shipped *two* aiguard cassettes: `...before_model_validation...ca44130a.json` = messages `[user]` (before-model) **and** `...before_model_validation...d13cc790.json` = messages `[user, assistant "I'm unable to provide..."]` (after-model).
- The **after-model (streaming)** test shipped only *one* aiguard cassette: `...after_model_validation_stream_True...5dec5d6a.json` = messages `[user "Tell me a short story about a robot."]` — prompt only. There is **no** cassette containing the reconstructed assistant story.

So on replay this test can only serve a before-model eval, and the assertion is satisfied by the before-model `prompt` span. Note the irony: the non-streaming before-model test is the only one that actually exercised an after-model eval, which suggests the streamed after-model path (`DD_AI_GUARD_ANALYZE_STREAM_RESPONSES_ENABLED=true`) did not fire during recording — worth chasing, not just an assertion tweak.

**Fix:** match an after-model-specific span (by target, or by `meta_struct.ai_guard.messages` containing an assistant message) and re-record so the after-model cassette exists. (Same as the resolved Codex-bot P2 thread.)

tests/integration_frameworks/llm/openai/test_openai_ai_guard.py
```diff
@@ -0,0 +1,191 @@
...
+        env["DD_API_KEY"] = os.environ["DD_API_KEY"]
+        env["DD_APP_KEY"] = os.environ["DD_APP_KEY"]
+    else:
+        env["DD_API_KEY"] = "mock_api_key"
+        env["DD_APP_KEY"] = "mock_app_key"
+    return env
+
+
+def _ai_guard_spans(traces: list[list[dict]]) -> list[dict]:
+    return [span for trace in traces for span in trace if span.get("resource") == "ai_guard"]
```
christophe-papazian, line 53: **Finding 2 (Medium) — "attached to the OpenAI trace" is asserted nowhere.**

The module docstring promises an `ai_guard` span "attached to the OpenAI trace," but this helper (and `_ai_guard_event_root_spans`) flattens **all** traces (`for trace in traces for span in trace`) and never requires an `openai.request` span nor compares `trace_id`. A fully detached `ai_guard` trace passes every assertion in this file.

**Confirming:** none of the three tests reference an OpenAI span or a shared `trace_id`; they only check span presence / `ai_guard.target`. The linkage is assertable — the recorded OpenAI cassettes carry `x-datadog-trace-id`, so the `ai_guard` span and the `openai.request` span can be required to share a `trace_id`.

Related: the resolved Codex-bot P2 on the tool-call test is the same class of gap — `target == "tool"` can also come from an ordinary after-model eval of an assistant response, so asserting on `meta_struct.ai_guard.messages` content (the `tool_calls` entry) is what actually proves the tool-call path reached AI Guard.

tests/integration_frameworks/llm/openai/test_openai_ai_guard.py
```diff
@@ -0,0 +1,191 @@
...
+        "DD_AI_GUARD_ENABLED": "true",
+        # after-model evaluation of streamed responses is opt-in
+        "DD_AI_GUARD_ANALYZE_STREAM_RESPONSES_ENABLED": "true",
+    }
+    # The AI Guard client needs an API key + app key. Real keys are required when recording
+    # cassettes (the client calls the real AI Guard API); mock keys are fine on replay since
+    # the VCR proxy matches on the request, not on auth.
+    if request.config.option.generate_cassettes:
+        env["DD_API_KEY"] = os.environ["DD_API_KEY"]
+        env["DD_APP_KEY"] = os.environ["DD_APP_KEY"]
```
christophe-papazian: **Finding 4 (Medium/High) — real Datadog keys via `library_env` are both leak-prone and can silently skip cassette generation.**

Two distinct problems on lines 46-47:

1. **Leak (matches the resolved Codex-bot P1).** `tests/integration_frameworks/conftest.py:64` copies `library_env` into scenario metadata:
   ```python
   context.scenario.parametrized_tests_metadata[request.node.nodeid] = dict(library_env)
   ```
   which the top-level conftest serializes into the JSON report — so during `--generate-cassettes` the real `DD_API_KEY`/`DD_APP_KEY` can land in logs/artifacts. Confirmed by that copy line.

2. **Silent skip.** `tests/integration_frameworks/conftest.py:13-22` marks every integration_frameworks test `xfail(strict=False)` under `--generate-cassettes`. `os.environ["DD_API_KEY"]` here is a hard index, and the scenario only pre-validates `OPENAI_API_KEY` (`utils/_context/_scenarios/integration_frameworks.py:23-26`). So if the DD keys are absent, the `KeyError` in fixture setup becomes a non-strict xfail — the aiguard cassettes are silently not generated while `generate-integration-framework-cassettes.sh` still prints "Cassettes generated."

**Fix:** pass these secrets outside `library_env` (so they aren't copied into report metadata), and add `DD_API_KEY`/`DD_APP_KEY` to `_required_cassette_generation_api_keys["openai"]` so a missing key fails fast via `pytest.exit`.

tests/ai_guard/test_ai_guard_sdk.py
```diff
@@ -652,6 +652,11 @@ def test_standalone_keeps_ai_guard_trace(self):
             assert root_span.get("meta", {}).get("_dd.p.dm") == "-" + str(SamplingMechanism.AI_GUARD), (
                 "Decision maker (_dd.p.dm) must match AI_GUARD sampling mechanism in standalone mode"
             )
+            # ai_guard.event:true is the tracer-emitted root-span marker that AI Guard ran on this
+            # trace. (The _dd.ai_guard.enabled facet seen in the UI is not present in raw payloads.)
+            assert root_span.get("meta", {}).get("ai_guard.event", False) in (True, "true"), (
```
christophe-papazian: **Finding 6 (Low) — this assertion is redundant and out of scope for the PoC (but not a regression).**

`ai_guard.event:true` is already covered by the dedicated `Test_AIGuardEvent_Tag::test_ai_guard_event` in this same file (lines ~574-594), which asserts the stronger property: set on the root span **and not** on children. Bolting the same check onto `test_standalone_keeps_ai_guard_trace` (whose purpose is USER_KEEP + AI_GUARD decision-maker) duplicates coverage and mixes concerns — and it lands in the `AI_GUARD`/`AI_GUARD_STANDALONE` scenario inside a PR whose subject is the `INTEGRATION_FRAMEWORKS` PoC.

**Confirming it is NOT a version-gating regression:** `Test_AIGuardStandalone` runs only on `flask-poc` at `v4.10.0` (`manifests/python.yml`), and `ai_guard.event` is supported from `v4.9.0` (`Test_AIGuardEvent_Tag` is `missing_feature (<4.9.0)`). Since 4.9.0 < 4.10.0 the tag is always present where this runs, so the assertion is safe — just unnecessary here.

**Suggestion:** drop it, or if the intent is to cover the tag *in standalone mode specifically*, add a standalone variant of `Test_AIGuardEvent_Tag` in its own commit rather than appending to the sampling test.

christophe-papazian (commented):
Reviewed the AI Guard PoC. Four inline notes below (findings 1, 2, 4, 6), each with evidence to confirm. Two of these overlap with the resolved Codex-bot threads and, in my reading, are still substantive — flagging so they aren't lost when this moves from draft toward runnable.

tests/integration_frameworks/llm/openai/test_openai_ai_guard.py
avara1986: **Partially fixed.** The docstring over-promised, so I reworded it to state explicitly that we do **not** assert trace linkage — the tracer may emit the `ai_guard` span as its own trace (as the helper comments already note), so a shared `trace_id` isn't guaranteed. For the tool-call half (same class as the resolved Codex P2), I tightened `test_tool_call_validation` to assert the assistant `tool_calls` entry is present in `meta_struct.ai_guard.messages` — that proves the tool-call path reached AI Guard rather than relying on `target == 'tool'` (which an ordinary after-model eval can also produce). (fbb137301)

tests/integration_frameworks/llm/openai/test_openai_ai_guard.py
avara1986: **Fixed both problems.** Moved `DD_API_KEY`/`DD_APP_KEY` out of `library_env` (so they're no longer copied into `parametrized_tests_metadata` → the JSON report) and into `_required_cassette_generation_api_keys['openai']`. They now reach the container via the scenario environment, and a missing key fails fast via `pytest.exit` at configure time instead of the silent non-strict xfail. Heads-up: this makes **all** openai cassette generation require the DD keys, not just AI Guard — happy to scope it tighter if you'd prefer. (fbb137301)

tests/ai_guard/test_ai_guard_sdk.py
avara1986: **Fixed — dropped it.** Removed the `ai_guard.event` assertion from `test_standalone_keeps_ai_guard_trace`. Agreed it duplicated `Test_AIGuardEvent_Tag::test_ai_guard_event` (which asserts the stronger root-only property) and was out of scope for this INTEGRATION_FRAMEWORKS PoC. (fbb137301)

tests/integration_frameworks/llm/openai/test_openai_ai_guard.py
```diff
@@ -0,0 +1,143 @@
...
```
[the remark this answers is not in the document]
avara1986: **Fixed.** `DD_API_KEY`/`DD_APP_KEY` are no longer placed in `library_env` (which is copied into the JSON report metadata) — they're now injected via the scenario environment through `_required_cassette_generation_api_keys['openai']`, and a missing key fails fast instead of a silent xfail. (fbb137301)

tests/integration_frameworks/llm/openai/test_openai_ai_guard.py
```diff
@@ -0,0 +1,143 @@
...
+                "/chat/completions",
+                dict(
+                    model="gpt-4o-mini",
+                    messages=[{"role": "user", "content": "Tell me a short story about a robot."}],
+                    parameters=dict(max_tokens=35, stream=True),
+                ),
+            )
+
+        guard_spans = _wait_for_ai_guard_spans(test_agent)
+        assert guard_spans, "expected an ai_guard span from the after-model (streamed) evaluation"
```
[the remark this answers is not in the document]
avara1986: **Assertion fixed; test skipped pending a cassette re-record.** Now requires a span whose `meta_struct.ai_guard.messages` includes the assistant response, so the before-model prompt span can't satisfy it. The current cassette only recorded the prompt eval, so the test is `@pytest.mark.skip` until it's regenerated (APPSEC-68977). (fbb137301)

tests/integration_frameworks/llm/openai/test_openai_ai_guard.py
```diff
@@ -0,0 +1,143 @@
...
+                            "role": "user",
+                            "content": "Bob is a student at Stanford University. He is studying computer science.",
+                        }
+                    ],
+                    parameters=dict(tool_choice="auto", tools=TOOLS),
+                ),
+            )
+
+        guard_spans = _wait_for_ai_guard_spans(test_agent, target="tool")
+        assert guard_spans, "expected a tool-call ai_guard span with target 'tool'"
```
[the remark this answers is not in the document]
avara1986: **Fixed.** `test_tool_call_validation` now asserts the assistant `tool_calls` entry is present in `meta_struct.ai_guard.messages`, proving the tool-call path forwarded the tool calls to AI Guard rather than relying on `target == 'tool'` alone. (fbb137301)

tests/integration_frameworks/llm/openai/test_openai_ai_guard.py
avara1986: **Update — resolved by removing the test instead of re-recording.** Streamed after-model evaluation (`DD_AI_GUARD_ANALYZE_STREAM_RESPONSES_ENABLED`) isn't implemented across the other tracer libraries yet, so an after-model test can't run at cross-language parity — it would only ever be the false positive you flagged. Removed `test_after_model_validation`, the stream env var from `library_env`, and the two orphaned streaming cassettes (`...5dec5d6a`, `...5c14b241`). before-model + tool-call coverage stays. No cassette re-record needed. Docstring now states after-model is intentionally out of scope. (bc915c933)

tests/integration_frameworks/llm/openai/test_openai_ai_guard.py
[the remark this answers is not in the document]
avara1986: **Update — test removed rather than tightened.** The streamed after-model path isn't implemented across the other tracer libraries yet, so this test can't run at cross-language parity. Removed `test_after_model_validation` + the stream env var + the orphaned streaming cassettes; before-model and tool-call coverage remain. (bc915c933)

avara1986:
### Update — Finding 1 resolved (commit bc915c933)

Rather than re-recording the after-model cassette, we **removed the streamed after-model test entirely**: `DD_AI_GUARD_ANALYZE_STREAM_RESPONSES_ENABLED` (streamed-response reconstruction) is not implemented across the other tracer libraries yet, so an after-model test can't run at cross-language parity — it could only ever be the false positive that was flagged.

Changes:
- Removed `test_after_model_validation` and the `DD_AI_GUARD_ANALYZE_STREAM_RESPONSES_ENABLED` env var from `library_env`.
- Deleted the two orphaned streaming cassettes (`...after_model_validation_stream_True...5dec5d6a.json`, `...5c14b241.json`).
- Docstring updated to state after-model is intentionally out of scope.
- **No cassette re-record needed.** before-model + tool-call coverage is unchanged.

Updated status table:

| Finding | Status |
|---|---|
| **F4 / Codex P1** — DD keys leak via `library_env` + silent xfail | ✅ Fixed |
| **Codex P2 / F2 (tool half)** — tool-call payload not verified | ✅ Fixed |
| **F2 (docstring)** — trace linkage asserted nowhere | ✅ Fixed |
| **F6** — redundant `ai_guard.event` assert | ✅ Fixed |
| **F1 / Codex P2 (after-model)** — false positive | ✅ **Fixed** — after-model streaming test removed (not implemented cross-language yet); no re-record needed |

christophe-papazian (approved):

PR #7277, opened 2026-07-07
test(ffe): cover agentless CDN configuration-source contract

leoromanovsky:
## Motivation

Server SDK agentless flag delivery needs one shared system-test contract before the SDK implementations land. This PR defines the cross-SDK behavior for configuration-source selection, HTTP UFC polling, cold failures, warm last-known-good preservation, and non-overlapping polls.

Tracking: `FFL-2731`, under epic `FFL-2662`.

## Changes

- Adds parametric FFE coverage for explicit `remote_config`, explicit and default `agentless`, invalid source handling, auth and malformed-payload failures, timeout, retry recovery, ETag/`304`, last-known-good preservation, and poller concurrency.
- Uses the current provider gate, `DD_EXPERIMENTAL_FLAGGING_PROVIDER_ENABLED=true`.
- Pins the existing Remote Config-based FFE evaluation and span-enrichment suites to `DD_FEATURE_FLAGS_CONFIGURATION_SOURCE=remote_config`, so the new agentless default does not change their transport under test.
- Uses the Java agentless configuration contract:
  - `DD_FEATURE_FLAGS_CONFIGURATION_SOURCE`
  - `DD_FEATURE_FLAGS_CONFIGURATION_SOURCE_AGENTLESS_BASE_URL`
  - `DD_FEATURE_FLAGS_CONFIGURATION_SOURCE_AGENTLESS_POLL_INTERVAL_SECONDS`
  - `DD_FEATURE_FLAGS_CONFIGURATION_SOURCE_AGENTLESS_REQUEST_TIMEOUT_SECONDS`
- Passes the mock UFC URL through the public agentless base-URL option; a custom HTTP endpoint remains an endpoint choice under `agentless`, not a separate source mode.
- Adds a controllable mock agentless UFC backend that records request metadata, response status, ETag, auth presence, and concurrency without exposing UFC bytes through its status API.
- Adds `host.docker.internal:host-gateway` automatically when a library environment contains the mock endpoint.
- Gates the new parametric suite as `missing_feature` for the seven v1 SDKs and `irrelevant` for C++ and Rust.
- Keeps SDK-independent mock-backend coverage under `tests/test_the_test/`.

## Decisions

- `agentless` is the default direct-HTTP source. The default first-party Datadog CDN-backed endpoint and a configured custom HTTP endpoint use the same mode and UFC polling semantics.
- `remote_config` is explicit opt-in and remains distinct because its Agent-mediated security, signing, targeting, capability, and subscription lifecycle differs from direct HTTP polling.
- `offline` remains reserved for future startup-provided UFC bytes with no network source.
- This PR does not introduce `DD_FEATURE_FLAGS_ENABLED`; it preserves the existing experimental provider gate used by current SDKs and test applications.
- Remote Config assertions start from the fixture-owned clean backend state. Tests do not reset after the SDK container starts, so an incorrect early agentless request cannot be erased.
- The mock suppresses client-disconnect errors while sending headers and bodies, keeping timeout tests free of fixture-side broken-pipe noise.

## Coverage

The suite defines 15 SDK-gated parametric cases across:

- source selection and no-fallback behavior
- cold auth, malformed payload, timeout, retry, and `304` behavior
- warm last-known-good preservation
- ETag propagation
- non-overlapping delayed polls

The mock backend itself has three SDK-independent tests covering UFC delivery and metadata recording, Docker host-gateway mapping, and metadata-only status output.

## Validation

```text
$ ./run.sh TEST_THE_TEST tests/test_the_test/test_mock_ffe_agentless_backend.py -q
3 passed in 2.32s
```

```text
$ ./format.sh --check
Success: no issues found in 514 source files
598 files already formatted
All checks passed!
All good, the system-tests CI will be happy!
```


tests/parametric/test_ffe/test_mock_ffe_cdn_reachability.py
```diff
@@ -0,0 +1,49 @@
...
+
+from utils import features, scenarios
+from utils._context.docker import get_docker_client
+from utils.docker_fixtures import TestAgentAPI
+from utils.docker_fixtures._mock_ffe_cdn import CONFIG_PATH, EXPECTED_API_KEY, MockFFECDNServer
+
+pytest_plugins = ["utils.docker_fixtures._mock_ffe_cdn"]
+
+
+@scenarios.parametric
```
cbeauchesne: This test is not related to any library, but rather a "test the test" stuff.

Though, as it, we cannot move it away from the parametric scenario. Two possible solutions : 

1. Since the calling script is a bare python script not really related to the test agent, we can do a simpler unit test case inside tests/test_the_tests. The rest of the behavior will be indirectly (but completely) tested through other real tests, so in term of coverage, we're fine. It's my favorite solution.
2. if you'd like to keep the test as it, we must not execute in any other CI, as failures will be essentially noise for users. The idea is to run it only on system-tests CI, if ever it runs on a CI. You can achieve that by using the `irrelevant` decorator, and checking env var values (three use case to consider : gitlab, github and azure)

tests/parametric/test_ffe/test_mock_ffe_cdn_reachability.py
```diff
@@ -0,0 +1,49 @@
...
+@features.feature_flags_dynamic_evaluation
```
cbeauchesne: As we're testing system-tests internal, it does not make sense to attach it to any feature : 
```suggestion
@features.not_reported
```

tests/parametric/test_ffe/test_configuration_sources.py
```diff
@@ -0,0 +1,470 @@
+"""Parametric FFE configuration-source coverage for mocked FFE CDN validation."""
```
cbeauchesne: Could you add a comment on what does the tested feature, and what is the test strategy ?

utils/docker_fixtures/_core.py
```diff
@@ -10,6 +11,10 @@
 from utils._context.docker import get_docker_client
 
 
+HOST_DOCKER_INTERNAL = "host.docker.internal"
```
cbeauchesne, line 14: Not a request change, but rather a question : I'm a little bit surprised to see that working. If I understand correctly, the test agent container will make request to a process running inside the host ? 

Few years ago, I tried this, and I failed to find a solution that works everywhere (Mac, Linux, CI...).

Do you checked this works on all those setup?

If ever the answer is no, unfortunately we'll need to find another solution, as this is a core requirement for system-tests (but fortunately, I do have solutions 😃  )

cbeauchesne (dismissed):
I didn't looked closely the test logic, but as I know nothing aboutthe tested feature, a review from someone else will be a better idea :) 

For the framework usage, I do have few questions, the most important one is the compatibility with all setup where system-tests runs. Then, just a few request change on the test-the-test stuff, and all good.

utils/docker_fixtures/_core.py
leoromanovsky: I wrote a "test a test" to validate this. It passed on my Mac laptop and looks like it here in CI as well.

tests/parametric/test_ffe/test_mock_ffe_cdn_reachability.py
leoromanovsky: Went with 1

tests/parametric/test_ffe/test_configuration_sources.py
```diff
@@ -0,0 +1,470 @@
...
+    @parametrize("library_env", [{"configuration_source": "remote_config", "response": "valid"}], indirect=True)
+    def test_remote_config_positive_ignores_cdn_env(
+        self, test_agent: TestAgentAPI, test_library: APMLibrary, mock_ffe_cdn: MockFFECDNServer
+    ) -> None:
+        mock_ffe_cdn.reset()
+        apply_state = _set_and_wait_ffe_rc(test_agent, UFC_VALID_DATA)
+        assert apply_state["apply_state"] == RemoteConfigApplyState.ACKNOWLEDGED.value
+        assert apply_state["product"] == RC_PRODUCT
+
+        assert test_library.ffe_start(UFC_VALID_DATA), "failed to start FFE provider in remote_config mode"
```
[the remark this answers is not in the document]
leoromanovsky: addressed in 4b67703c5

utils/docker_fixtures/_mock_ffe_cdn.py
```diff
@@ -0,0 +1,333 @@
...
+
+    def record_request(self, headers: Mapping[str, str], path: str) -> str:
+        parsed = urlparse(path)
+
+        with self._lock:
+            self.requests_total += 1
+            self.in_flight += 1
+            self.max_in_flight = max(self.max_in_flight, self.in_flight)
+            self.last_path = parsed.path
+            self.last_if_none_match = headers.get("If-None-Match")
```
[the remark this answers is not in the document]
leoromanovsky: addressed in 4b67703c5

tests/parametric/test_ffe/test_configuration_sources.py
```diff
@@ -0,0 +1,470 @@
...
+        self,
+        test_library: APMLibrary,
+        mock_ffe_cdn: MockFFECDNServer,
+        expected_status_codes: list[int],
+    ) -> None:
+        assert test_library.ffe_start(), "failed to start FFE provider for bad_to_good"
+
+        status = _wait_for_status(
+            mock_ffe_cdn,
+            lambda current: current["status_codes"][-2:] == expected_status_codes,
```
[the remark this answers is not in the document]
leoromanovsky: addressed in 4b67703c5

utils/docker_fixtures/_core.py
leoromanovsky: Test the test from last CI run.
https://github.com/DataDog/system-tests/actions/runs/28940954291/job/85862711479

Relevant log snippet:

```
  2026-07-08T12:01:15.7985641Z /bin/bash run.sh TEST_THE_TEST
  2026-07-08T12:01:18.2086735Z Scenario: TEST_THE_TEST
  2026-07-08T12:01:18.2087055Z Logs folder: ./logs_test_the_test

  2026-07-08T12:02:33.3884000Z tests/test_the_test/test_mock_ffe_cdn.py ..                              [ 50%]

  2026-07-08T12:03:14.8169872Z - generated xml file: /home/runner/work/system-tests/system-tests/logs_test_the_test/reportJunit.xml
  2026-07-08T12:03:14.8172640Z ========= 323 passed, 2374 deselected, 1 xfailed in 116.59s (0:01:56) ==========

  The .. on tests/test_the_test/test_mock_ffe_cdn.py is our two new fixture tests passing in CI.
```

leoromanovsky:
> I didn't looked closely the test logic, but as I know nothing aboutthe tested feature, a review from someone else will be a better idea :)
> 
> For the framework usage, I do have few questions, the most important one is the compatibility with all setup where system-tests runs. Then, just a few request change on the test-the-test stuff, and all good.

@cbeauchesne Thanks for the suggestions; normal and expected that FFE will approve changes to features, I will hold the PR for their review.

cbeauchesne (commented):
AGTM for framwork usage. I let the real review for the test logic :)

tests/test_the_test/test_mock_ffe_agentless_backend.py
```diff
@@ -0,0 +1,45 @@
...
+@features.not_reported
+def test_mock_ffe_agentless_backend_host_gateway_mapping(monkeypatch: pytest.MonkeyPatch, worker_id: str) -> None:
+    monkeypatch.delenv("SYSTEM_TESTS_MOCK_FFE_AGENTLESS_BACKEND_BASE_URL", raising=False)
+    monkeypatch.delenv("SYSTEM_TESTS_MOCK_AGENTLESS_BACKEND_BASE_URL", raising=False)
+    monkeypatch.delenv("SYSTEM_TESTS_MOCK_FFE_AGENTLESS_BACKEND_HOST", raising=False)
+    monkeypatch.delenv("SYSTEM_TESTS_MOCK_AGENTLESS_BACKEND_HOST", raising=False)
+
+    server = MockFFEAgentlessBackendServer(worker_id)
+    try:
+        env = {"DD_FEATURE_FLAGS_CONFIGURATION_SOURCE_AGENTLESS_BASE_URL": server.library_config_url}
```
sameerank: Don't forget to update the `DD_FLAGGING_CONFIGURATION_SOURCE*` references in the PR description. Also references to "cdn" are now "agentless"

tests/parametric/test_ffe/test_configuration_sources.py
```diff
@@ -0,0 +1,535 @@
...
+    )
+
+
+@scenarios.parametric
+@features.feature_flags_agentless
+class Test_Feature_Flag_Configuration_Source_Selection:
+    """Validate source selection for Agent RC, agentless, and customer endpoint overrides."""
+
+    @parametrize("library_env", [{"configuration_source": "remote_config", "response": "valid"}], indirect=True)
+    def test_remote_config_positive_ignores_agentless_env(
```
sameerank, line 193: The PR description needs some updating:
1. `test_remote_config_positive_ignores_cdn_env` should be changed to this
2. there's a reference to `tests/parametric/test_ffe/test_mock_ffe_cdn_reachability.py` in this folder which is no longer in the diff

tests/parametric/test_ffe/test_configuration_sources.py
```diff
@@ -0,0 +1,535 @@
...
+        assert status["max_in_flight"] == 1
+        assert status["last_path"] == CONFIG_PATH
+
+
+@scenarios.parametric
+@features.feature_flags_agentless
+class Test_Feature_Flag_Configuration_Source_Mock_Fixture:
+    """Validate that the mock backend exposes only metadata needed by the tests."""
+
+    def test_mock_ffe_agentless_backend_status_is_metadata_only(
```
sameerank: Your manifest file updates gate this entire file, e.g.

https://github.com/DataDog/system-tests/blob/299f69144f27933b999ca1f180070a9f648b5004/manifests/java.yml#L3737

But this at least looks like a test that is independent of the SDKs and doesn't need to be gated. Maybe this could set elsewhere like `tests/test_the_test/`

tests/parametric/test_ffe/test_configuration_sources.py
```diff
@@ -0,0 +1,535 @@
...
+    """Validate source selection for Agent RC, agentless, and customer endpoint overrides."""
+
+    @parametrize("library_env", [{"configuration_source": "remote_config", "response": "valid"}], indirect=True)
+    def test_remote_config_positive_ignores_agentless_env(
+        self,
+        test_agent: TestAgentAPI,
+        test_library: APMLibrary,
+        mock_ffe_agentless_backend: MockFFEAgentlessBackendServer,
+    ) -> None:
+        mock_ffe_agentless_backend.reset()
```
sameerank: Some feedback from Claude that might be helpful here:

>`mock_ffe_agentless_backend.reset()` runs after the test_library container is already up. It's fine today because provider polling starts at `ffe_start()` (called later), but if any SDK begins agentless polling at process init rather than provider-set, an early erroneous request would be wiped by `reset()` and `_assert_no_mock_requests` would pass falsely. Capturing the clean baseline after `ffe_start()`, or documenting the lifecycle assumption, would harden it.

utils/docker_fixtures/_mock_ffe_agentless_backend.py
```diff
@@ -0,0 +1,336 @@
...
+            self._write_json(HTTPStatus.BAD_REQUEST, {"error": str(error)})
+            return
+
+        self.server.state.set_responses(validated_responses)
+        self._write_json(HTTPStatus.OK, self.server.state.status())
+
+    def _write_json(self, status_code: HTTPStatus, payload: dict[str, Any] | MockFFEAgentlessBackendStatus) -> None:
+        self.send_response(status_code)
+        self.send_header("Content-Type", "application/json")
+        self.end_headers()
```
sameerank: `.send_response` and `.end_headers` are wrapped in a block in `_handle_config`

https://github.com/DataDog/system-tests/blob/51372e948740ba58425088289d45da44897b8be0/utils/docker_fixtures/_mock_ffe_agentless_backend.py#L179-L191

Would it make sense to do the same here?

utils/build/docker/dotnet/weblog/Controllers/FeatureFlagEvaluatorController.cs
```diff
@@ -21,7 +21,10 @@ public class FeatureFlagEvaluatorController : Controller
 
         private static global::OpenFeature.FeatureClient? InitClient()
         {
-            var activation = Environment.GetEnvironmentVariable("DD_EXPERIMENTAL_FLAGGING_PROVIDER_ENABLED")?.ToLowerInvariant() ?? "0";
+            var activation = (
+                Environment.GetEnvironmentVariable("DD_FEATURE_FLAGS_ENABLED")
+                ?? Environment.GetEnvironmentVariable("DD_EXPERIMENTAL_FLAGGING_PROVIDER_ENABLED")
+                ?? "0").ToLowerInvariant();
```
sameerank: This adds `DD_FEATURE_FLAGS_ENABLED` as an alias for the legacy enable flag, which makes sense for updating system-tests for using the new env variables. 

But fyi, this is the e2e weblog controller, and the parametric suite this PR adds runs against `utils/build/docker/dotnet/parametric/Endpoints/ApmTestApi.cs ` so this edit isn't touched by what the new tests run against

sameerank (approved):
You may want to make further adjustments when enabling these tests against an SDK, but this looks like a good start and seems aligned with the RFC so far

tests/test_the_test/test_mock_ffe_agentless_backend.py
leoromanovsky: Thank you; yes, performing a careful synchronization before merging:

RFC -> java trace, run against local update system test -> push trace -> push system test 🚀

cbeauchesne (approved):

PR #7287, opened 2026-07-09
[Debugger] Add system-test coverage for the debugger guardrails

dudikeleti:
## Motivation

Add focused system-test coverage for the debugger guardrails required for GA.

## Changes

- Add method- and line-probe evaluation-timeout tests covering an expensive regex and a large collection filter. Aborted evaluations must produce exactly one evaluation-error snapshot with `evaluationErrors[]` populated and no captured data.
- Add snapshot-size guardrail coverage verifying that the request containing the snapshot is ≤1 MB and includes a payload-pruning marker.
- Add capture-timeout coverage verifying that the affected collection reports `notCapturedReason=timeout`, without incorrectly satisfying the snapshot-size guardrail.
- Add the dedicated `DEBUGGER_CAPTURE_TIMEOUT` scenario with a 10 ms capture budget, CI orchestration, and compatible weblog filtering.
- Add the `/debugger/snapshot/capture-timeout` fixture to supported .NET, Java, and Node.js weblogs and document the endpoint.
- Extend `test_default_max_length` to verify that truncated strings report `truncated=true`.
- Extract the shared `captures_contain_data` helper into `tests.debugger.utils`.
- Enable capture-timeout coverage for supported .NET, Java, and Node.js tracer versions. Keep unsupported weblogs and the evaluation-timeout and snapshot-size guardrails disabled through manifests until support lands.

## Workflow

## Reviewer checklist


dudikeleti:
@codex review

tests/debugger/test_debugger_guardrails.py
```diff
@@ -0,0 +1,268 @@
+# Unless explicitly stated otherwise all files in this repository are licensed under the the Apache License Version 2.0.
```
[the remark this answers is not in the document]
dudikeleti: fixed

tests/debugger/test_debugger_guardrails.py
```diff
@@ -0,0 +1,268 @@
...
+
+    def test_snapshot_size_cap(self) -> None:
+        _, snapshot = self._get_single_snapshot()
+        serialized_size = len(json.dumps(snapshot, separators=(",", ":")).encode())
+        assert serialized_size <= MAX_SNAPSHOT_BYTES, (
+            f"Snapshot should be trimmed to <= {MAX_SNAPSHOT_BYTES} bytes, got {serialized_size} bytes"
+        )
+
+        large_collection = self._get_captured_local(snapshot, "largeCollection")
+        assert self._is_partially_captured(large_collection, self.SNAPSHOT_SIZE_COLLECTION_ITEMS), (
```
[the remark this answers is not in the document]
dudikeleti: fixed

tests/debugger/test_debugger_guardrails.py
```diff
@@ -0,0 +1,268 @@
...
+            return False
+
+        if value.get("notCapturedReason") == "payloadTooLarge" or value.get("pruned") is True:
+            return True
+
+        elements = value.get("elements") or value.get("entries")
+        if isinstance(elements, list) and len(elements) < expected_items:
+            return True
+
+        return any(self._is_partially_captured(child, expected_items) for child in value.values())
```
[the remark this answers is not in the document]
dudikeleti: fixed

tests/test_the_test/test_ci_orchestrator.py
```diff
@@ -51,6 +51,37 @@ def _is_supported(weblog: WeblogMetaData, scenario: Scenario) -> bool:
...
+        ("java", "spring-boot-wildfly"),
+        ("java", "uds-spring-boot"),
+        ("nodejs", "express4"),
+        ("nodejs", "express4-typescript"),
+        ("nodejs", "express5"),
+        ("nodejs", "fastify"),
+        ("nodejs", "uds-express4"),
+    }
+    libraries = ("dotnet", "golang", "java", "nodejs", "php", "python", "ruby")
+    available_weblogs = {(library, weblog_name) for library in libraries for weblog_name in get_weblogs(library)}
```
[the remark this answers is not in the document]
dudikeleti: fddb75d74d686298c8d3099876b62a94e485720b

utils/_context/_scenarios/__init__.py
```diff
@@ -893,6 +893,18 @@ class _Scenarios:
...
+    debugger_capture_timeout = DebuggerScenario(
+        "DEBUGGER_CAPTURE_TIMEOUT",
+        weblog_env={
+            "DD_DYNAMIC_INSTRUMENTATION_CAPTURE_TIMEOUT": "10",
+            "DD_DYNAMIC_INSTRUMENTATION_CAPTURE_TIMEOUT_MS": "10",
+            "DD_DYNAMIC_INSTRUMENTATION_MAX_TIME_TO_SERIALIZE": "10",
+            "DD_DYNAMIC_INSTRUMENTATION_ENABLED": "1",
+        },
+        doc="Test that debugger snapshot capture reports when its time budget is exceeded",
+        weblog_categories=[],
```
cbeauchesne: This means the scenario will never be executed. is it what you want ?

cbeauchesne: Ok, I see you've added it in weblog_metadata. In theory, those metadata should not be used for test activation, as it's the purpose of manifests. Do you have a reason to do so ?

cbeauchesne: There is a cleaner approach if ever you'd like to run only on some weblog, it's using the other way around : 

*  the scenario applies for weblog shipping a `ddtrace` library, so we must keep this property : keep weblog categories as it.
* Add the scenario in weblog_metadata `excluded_scenarios` property.

With that, we don't hack the `weblog_categories` semantic.

cbeauchesne (changes requested):

utils/_context/_scenarios/__init__.py
dudikeleti: e7786f6f79856ad4bcd0cd91cd6ac9a6b45e6f0f

tests/debugger/test_debugger_guardrails.py
```diff
@@ -0,0 +1,273 @@
...
+            )
+
+            envelope = snapshots[0]
+            snapshot = envelope.get("debugger", {}).get("snapshot") or envelope.get("debugger.snapshot") or {}
+
+            evaluation_errors = snapshot.get("evaluationErrors") or []
+            assert evaluation_errors, (
+                "Evaluation-timeout snapshot has an empty evaluationErrors[]; the tracer must "
+                "report that expression evaluation was aborted."
+            )
```
grantseltzer, line 92: This only checks that evaluationErrors is non-empty. Do we want to check the specific evaluation failures?

tests/debugger/test_debugger_guardrails.py
```diff
@@ -0,0 +1,273 @@
...
+
+
+@rfc(GUARDRAILS_RFC)
+@features.debugger_line_probe
+@scenarios.debugger_probes_snapshot
+@slow
+class Test_Debugger_Snapshot_Size_Guardrail(_DebuggerSnapshotGuardrailTest):
+    """RFC guardrail for the completed snapshot size."""
+
+    # Before enabling, verify this exceeds 1 MB without timing out; otherwise use a dedicated scenario or remove it.
```
grantseltzer, line 205: Has this been handled?

tests/debugger/test_debugger_guardrails.py
```diff
@@ -0,0 +1,273 @@
...
+            self.setup_failures.append("Snapshot was not received within 30s")
+
+    def _get_single_snapshot(self) -> tuple[str, dict[str, Any]]:
+        self.collect()
+        self.assert_setup_ok()
+        self.assert_rc_state_not_error()
+        self.assert_all_probes_are_emitting()
+        self.assert_all_weblog_responses_ok()
+
+        for probe_id in self.probe_ids:
```
grantseltzer: The for loop returns on the first probe, so any second probe is silently ignored, and the trailing raise AssertionError is only reachable when probe_ids is empty. Every probe file here has exactly one probe. Just assert that and index directly.

utils/_context/_scenarios/debugger.py
```diff
@@ -8,7 +8,12 @@
 
 
 class DebuggerScenario(EndToEndScenario):
-    def __init__(self, name: str, doc: str, weblog_env: dict[str, str | None]) -> None:
+    def __init__(
+        self,
+        name: str,
+        doc: str,
+        weblog_env: dict[str, str | None],
+    ) -> None:
```
grantseltzer: Autolint?

grantseltzer (approved):
LGTM, but please take a look at few small comments.

utils/build/docker/cpp_httpd/weblog_metadata.yml
```diff
@@ -2,4 +2,7 @@
 
 httpd:
   categories: [dd_trace]
-  excluded_scenarios: [APPSEC_MISSING_RULES, APPSEC_CORRUPTED_RULES] # C++ 1.2.0 freeze when the rules file is missing
\ No newline at end of file
+  excluded_scenarios:
+    - APPSEC_MISSING_RULES  # C++ 1.2.0 freezes when the rules file is missing
+    - APPSEC_CORRUPTED_RULES  # C++ 1.2.0 freezes when the rules file is corrupt
+    - DEBUGGER_CAPTURE_TIMEOUT
```
nccatoni: If I'm not mistaken this scenario is used by only one test and is the only scenario of that test so this should be replaced by a manifest entry

nccatoni (changes requested):

utils/build/docker/cpp_httpd/weblog_metadata.yml
dudikeleti: I may be missing some context here, since I’m not very familiar with this code. Based on the AI agent’s explanation, though, my understanding is that the current approach is preferable because the manifest already marks all debugger tests as irrelevant for `cpp_httpd`.

As I understand it, this intentionally prevents the orchestrator from scheduling the dedicated weblogs scenario without the `capture-timeout` fixture, while manifests only control pytest activation. Adding another manifest entry would therefore be redundant and would increase the CI matrix.

Does that reasoning make sense, or am I misunderstanding how this works?

tests/debugger/test_debugger_guardrails.py
dudikeleti: We intentionally avoid checking the exact failure message because timeout wording isn’t standardized across tracers. The shared contract currently requires a populated evaluationErrors[] and no captured data.

tests/debugger/test_debugger_guardrails.py
dudikeleti: Yes, theoretically. The fixture is designed to exceed 1 MiB using a large string and collection. The test also requires a payload-size pruning marker, so a timeout alone cannot produce a false pass. However, it remains disabled pending per-tracer activation and end-to-end verification.

tests/debugger/test_debugger_guardrails.py
dudikeleti: Fixed in be6355af1256c8707b5637384d101cba5ac0337e

utils/_context/_scenarios/debugger.py
dudikeleti: 62aa7d8d2fc6caea4148e6ac907b499dc8f42866

utils/build/docker/cpp_httpd/weblog_metadata.yml
nccatoni: My bad, I didn't see that there was already a manifest entry covering those tests. I don't think I fully understand what you want to do here. Could you please clarify what the objective of deactivating the scenario is?

dudikeleti: The objective is not to deactivate the test; manifests already control test support. The exclusion prevents the CI matrix from pairing DEBUGGER_CAPTURE_TIMEOUT with weblogs that do not implement its dedicated fixture. Without it, CI would schedule the scenario and start its infrastructure before pytest applied the manifest and deselected the disabled tests.

Debugger tests currently do not apply to C++. Existing debugger scenarios predate metadata-based compatibility filtering and still rely on manifests to produce empty runs. For this new fixture-specific scenario, we restrict the matrix upfront so weblogs without the endpoint never start it. We could apply the same optimization to all debugger scenarios for C++ in a separate cleanup, but I would prefer not to broaden this PR.

nccatoni: In general, deactivated tests should still be run. This allows us to detect when they start passing and can be reactivated once the feature is implemented.

If the feature is irrelevant for a specific weblog, you should use an `irrelevant` declaration in the manifest so the test is skipped entirely.

If the feature might be supported in the future, but running the test causes undesirable side effects, you can force-skip it while it is deactivated by using the `@slow` or `@scenario_crash` decorator.

For more details, take a look at the documentation:
https://github.com/DataDog/system-tests/blob/main/docs/edit/skip-tests.md

dudikeleti: Thanks for the explanation and doc link. I’ll remove the scenario exclusions and rely on the manifests.

dudikeleti: 83397a42c8c68a6ba65d31e2d797b89eef893b8d

nccatoni (approved):

cbeauchesne (approved):

PR #7288, opened 2026-07-10
feat(rust): axum weblog

MilanGarnier:
## Motivation

Rust is lacking a weblog to run end to end tests. The goal is to provide a canonical weblog based on axum and community otel integrations. 

## Changes

Introduces a new Axum-based Rust weblog and wires it into the end-to-end test suite:

- **New weblog app** (`utils/build/docker/rust/axum/`): an Axum HTTP server instrumented with `dd-trace-rs`, exposing the core endpoints needed by end-to-end tests (`/`, `/healthcheck`, `/headers`, `/status`, `/spans`, `/stats-unique`, `/params/{value}`,  `/sample_rate_route/{i}`, `/flush`, `/createextraservice`,  `/e2e_otel_span`, `/e2e_single_span`, `/returnheaders`,  `/make_distant_call`, and the OTel drop-in propagator endpoints).
- **Docker build** (`utils/build/docker/rust/axum.Dockerfile`): a  multi-stage build that compiles the weblog and installs `dd-trace-rs` via  `install_ddtrace.sh`, and adds the corresponding mirror image entry.
- **Manifest updates** (`manifests/rust.yml`): mark tests that exercise  endpoints or features not yet implemented in the Axum weblog or in  `dd-trace-rs` as `missing_feature`/`incomplete_test_app`/`irrelevant`,  while removing entries that no longer apply now that this weblog exists.
- Deliberately keeps `client_ip` tagging, referer-hostname extraction,
  and query-string obfuscation unimplemented in the weblog rather than
  faking them, since those are tracer responsibilities.

### ⚠️ OpenTelemetry dependency resolution (open for discussion)

`datadog-opentelemetry`, `tracing-opentelemetry`,`opentelemetry-instrumentation-tower` and the `opentelemetry*` crates must all resolve to the same `opentelemetry` minor version, or Rust links two copies side by side and half the app silently gets a no-op tracer/propagator (no server spans, no context extraction).

`install_ddtrace.sh` currently enforces this at build time rather than via static `Cargo.toml` pins: it installs `datadog-opentelemetry`, inspects `cargo metadata` to find the OTel minor it resolved to, re-pins the other OTel/contrib crates to match, then fails the build if `cargo metadata` still shows more than one `opentelemetry` version (usually because `opentelemetry-instrumentation-tower`, which is git-pinned, has drifted).

This self-adjusts as `dd-trace-rs` moves, but it's a lot of logic standing in for what should be static dependency info, and it only fails at build time, not at edit time. Not settled — open to alternatives (vendoring/forking the tower crate, manual re-pinning).

## Workflow

## Reviewer checklist


MilanGarnier:
@codex review

utils/build/docker/rust/axum/Cargo.toml
```diff
@@ -0,0 +1,60 @@
...
+# of truth (unbounded latest from crates.io, or a local /binaries checkout) and
+# then align_opentelemetry() repins the published family — opentelemetry*,
+# reqwest-tracing, tracing-opentelemetry — to the minor it resolves to.
+#
+# opentelemetry-instrumentation-tower is NOT on crates.io, so it cannot be
+# repinned by version/feature: its OpenTelemetry minor is whatever the git rev
+# below depends on. When datadog-opentelemetry moves to a new OpenTelemetry
+# minor, bump this rev to a commit on the matching minor; otherwise
+# check_single_opentelemetry_version fails the build with an actionable message.
+opentelemetry-instrumentation-tower = { git = "https://github.com/open-telemetry/opentelemetry-rust-contrib", rev = "66bfea4", package = "opentelemetry-instrumentation-tower", features = ["axum"] }
```
iunanua: `opentelemetry-instrumentation-tower` depends on `opentelemetry`, what if prod and dev use different otel versions?
Could we avoid using it? and just create the span in `datadog_specific_axum_layer`

utils/build/docker/rust/axum/src/integration.rs
```diff
@@ -0,0 +1,211 @@
...
+        let host = req.url().host_str().unwrap_or_default().to_owned();
+        let query_suffix = req
+            .url()
+            .query()
+            .map(|q| format!("?{q}"))
+            .unwrap_or_default();
+        let host_port = match req.url().port() {
+            Some(p) => format!("{host}:{p}"),
+            None => host.clone(),
+        };
```
yannham:
```suggestion
        let host_port = req.url().port()
            .map(|p| format!("{host}:{p}")
            .unwrap_or_else(|| host.clone());
```

utils/build/docker/rust/axum/src/main.rs
```diff
@@ -0,0 +1,501 @@
...
+        .filter_map(|(name, value)| {
+            value
+                .to_str()
+                .ok()
+                .map(|value| (name.as_str().to_owned(), value.to_owned()))
+        })
+        .collect()
+}
+
+async fn returnheaders(headers: HeaderMap) -> Json<HashMap<String, String>> {
```
yannham: Nit: unconventional naming

```suggestion
async fn header_map_to_json(headers: HeaderMap) -> Json<HashMap<String, String>> {
```

utils/build/docker/rust/axum/src/main.rs
```diff
@@ -0,0 +1,501 @@
...
+    Json(header_map_to_string_map(&headers))
+}
+
+async fn requestdownstream(State(state): State<AppState>, method: Method) -> Response {
```
yannham:
```suggestion
async fn request_downstream(State(state): State<AppState>, method: Method) -> Response {
```

utils/build/docker/rust/axum/src/main.rs
```diff
@@ -0,0 +1,501 @@
...
+    let client = ClientBuilder::new(reqwest::Client::new())
+        .with(TracingMiddleware::<DatadogClientSpanBackend>::new())
+        .build();
+
+    match client.request(method, &state.downstream_url).send().await {
+        Ok(response) if response.status().is_success() => match response.bytes().await {
```
yannham: Nit: could use `map`, `and_the`, `map_error` or similar combinators here to avoid tripe nesting. Another possibility is to make a small local closures that return an error type and take advantage of `?`; it will be most likely eliminated by rustc anyway.

utils/build/docker/rust/axum/Cargo.toml
```diff
@@ -0,0 +1,60 @@
+[package]
+name         = "weblog"
```
yannham: Nit: I wouldn't try to align things like that, just use one space. Diffs can become annoyingly big once we have to update one value and everything gets misaligned.

yannham (approved):
My knowledge of system tests is limited. I focused on the Rust part and nothing particularly stands out 👍

utils/build/docker/rust/axum/src/main.rs
MilanGarnier: I refactored this into [8343f9e](https://github.com/DataDog/system-tests/pull/7288/commits/8343f9eb2af0c8bd0984d5bb93db6782474a6410), it's more concise

utils/build/docker/rust/install_ddtrace.sh
```diff
@@ -6,44 +6,18 @@ cd /usr/app
 
 REPO_URL=https://github.com/DataDog/dd-trace-rs
 
-OTEL_DEPS=(
```
genesor, line 9: A comment explaining what the script is doing to resolve all the versions would be helpful as it's not straightforward

utils/build/docker/rust/axum/src/integration.rs
```diff
@@ -0,0 +1,211 @@
+//! Tracing and span-enrichment utilities for Datadog compatibility.
+//!
+//! This module adds Datadog-specific semantic-convention attributes to both
+//! inbound (server) and outgoing (client) HTTP spans so they satisfy the
+//! system-tests' `validate_all_spans` checks.
+//!
+//! Future work: If a proper datadog integration for axum gets implemented,
```
genesor: 💯 As we're validating against this integration to make sure the Rust tracer is compliant with what we expect this integration should be made public and properly supported at some point.

utils/build/docker/rust/axum/src/integration.rs
```diff
@@ -0,0 +1,211 @@
...
+
+    next.run(request).await
+}
+
+/// [`reqwest_tracing::ReqwestOtelSpanBackend`] that names outgoing spans
+/// `http.client.request` and enriches them with the same Datadog
+/// semantic-convention attributes used for inbound server spans.
+pub struct DatadogClientSpanBackend;
+
+impl ReqwestOtelSpanBackend for DatadogClientSpanBackend {
```
genesor: If possible it would be nice to split Axum and Reqwest integrations into separate files

genesor (commented):
Great work! 3 nits and we should be good to go

utils/build/docker/rust/install_ddtrace.sh
iunanua, line 1: This script is getting complicated and how we infer the versions is a bit fragile. But there is no easy solution. 
@bantonsson proposed to move these test apps to the dd-trace-rs repo but we might as well do it in the furure...

genesor (approved):
LGTM functionally speaking

utils/build/docker/rust/install_ddtrace.sh
MilanGarnier: I guess the integrations would eventually belong to the dd-trace-rs repo where they would be publicly documented (as @genesor points out) and versionned. But keeping the endpoints definitions in system tests would be more in line with the other tracers. If dd-trace-rs starts providing the integrations then it'll be easier to keep the versions aligned.

utils/build/docker/rust/axum/Cargo.toml
MilanGarnier: @iunanua Ideally we would remove this pin when `opentelemetry-instrumentation-tower` makes a new release, but they haven't released for the lastest opentelemetry minor yet. We could avoid using it altogether and do everything ourselves. Would you prefer having the content of this crate pretty much inlined here ?

iunanua: As the crate is just a single `lib.rs` I guess we can inline it  :+1:

MilanGarnier: @iunanua  Since this is the official otel contribution, I guess it still makes more sense to use it rather than reimplement it ourselves. I contacted them and they're currently preparing a release that we'll be able to use instead of the pinned commit

iunanua: but then as it depends on otel crates, its version should be handled by `install_ddtrace.sh` like the rest...

MilanGarnier: I agree that it should. However, right now they have not released the interface we need in this integration. (`HttpLayer`). They are currently working on releasing so we'll soon be able to apply the same strategy as the others.
Depending on the approach they have to managing opentelemetry versions, in the future we might have to rely on choosing between a set of versions/commits when resolving...

MilanGarnier: 9ba907083 I've updated `install_ddtrace.sh` to take it into account and I've made the comments clearer. When a new release of datadog-opentelemetry version will be released, it will be necessary to add the strategy for picking the right version for this crate to have system tests to pass (another commit hash or fixed version) Or we'll have our own integrations that contain the version handling on their side

iunanua: I have tried droping `opentelemetry-instrumentation-tower` dep [here](https://github.com/DataDog/system-tests/pull/7399) and I think I prefer that option.

It is just adding some more attributes in `datadog_specific_axum_layer` function and helps us avoid future problems.

wdyt @MilanGarnier?

MilanGarnier: I think the change is short enough to justify having one less issue with version handling. I'm good with it, do you want to merge it on branch or get a review there ? Other than that we should be good to go

iunanua: :tada: 
do whatever you prefer, merge it if you want

PR #7298, opened 2026-07-11
test(ffe): add agentless end-to-end harness

leoromanovsky:
## Motivation

Server SDK Feature Flags are moving to authenticated CDN delivery as the default configuration path, with Agent Remote Configuration remaining an explicit alternative. The existing FFE end-to-end scenario can exercise only the Agent/RC lifecycle: it starts an Agent and weblog, then supplies UFC after the application is running.

That cannot prove the agentless startup contract. An agentless SDK may request UFC while the application is starting, so the test backend must already exist before the weblog starts. The scenario must also run without an Agent; otherwise a passing test cannot rule out an accidental Agent dependency.

This PR adds that missing, dormant test foundation. It starts a deterministic CDN-style UFC backend before a no-Agent weblog, injects the endpoint and credentials, and verifies authenticated configuration fetch plus evaluation. No SDK manifest is enabled here, so ordinary SDK CI behavior does not change until a later activation PR.

```mermaid
flowchart LR
    subgraph BEFORE["Existing end-to-end coverage"]
        B1["Start Agent and weblog"] --> B2["SDK starts"]
        B3["RC fixture"] --> B4["Agent Remote Configuration"]
        B4 --> B2
        B2 --> B5["Evaluate flag"]
    end

    subgraph THIS["Agentless harness added here"]
        A1["1. Start mock CDN backend"] --> A2["2. Start weblog and SDK<br/>without an Agent"]
        A2 --> A3["3. Fetch UFC with API key"]
        A3 --> A4["4. Evaluate flag"]
    end

    GAP["Missing on main:<br/>pre-start UFC service<br/>and no-Agent lifecycle"] -.-> A1

    classDef current fill:#fcbf49,stroke:#8a5a00,stroke-width:3px,color:#111;
    class A1,A2,A3,A4 current;
```

## Stack Position

This is the inactive foundation of the stack. Later PRs add the telemetry topology and then activate individual Java capabilities only when their implementation is green.

```mermaid
flowchart TD
    ST1["#7298 · this PR<br/>no-Agent CDN configuration harness<br/>all manifests disabled"]
    ST2["#7299 · next stacked PR<br/>side-effect transport contracts<br/>sidecar preferred, direct fallback"]
    STM["#7300 and small follow-ups<br/>enable Java capabilities one at a time"]

    J1["dd-trace-java #11892<br/>agentless configuration source"]
    J2["dd-trace-java side-effects work<br/>exposures · evaluation EVP · OTLP"]

    ST1 --> ST2 --> STM
    J1 --> STM
    J2 --> STM

    classDef current fill:#fcbf49,stroke:#8a5a00,stroke-width:3px,color:#111;
    class ST1 current;
```

## Architecture Boundary

Configuration source and telemetry transport are independent. The Agent belongs only to the explicit Remote Configuration branch. Default agentless configuration has no Agent dependency. Post-evaluation side effects are a stacked follow-up: prefer the Datadog serverless sidecar when available and fall back to direct authenticated intake otherwise.

```mermaid
flowchart LR
    subgraph CONFIG["Configuration delivery"]
        RC["RC fixture"] --> RCA["Agent<br/>RC only"]
        RCA --> RCS["SDK remote_config source"]

        CDN["Datadog CDN<br/>or controlled mock"] -->|"authenticated HTTP"| AGS["SDK agentless source<br/>default"]

        RCS --> EVAL["Same UFC / OpenFeature evaluator"]
        AGS --> EVAL
    end

    subgraph FOLLOWUP["Telemetry transport · stacked follow-up"]
        EVAL --> FX["Exposure EVP · evaluation EVP · OTLP metrics"]
        FX --> PICK{"Serverless sidecar available?"}
        PICK -->|"yes"| SIDECAR["Datadog serverless sidecar"]
        PICK -->|"no"| DIRECT["Direct authenticated intake"]
        SIDECAR --> INTAKE["Datadog backend"]
        DIRECT --> INTAKE
    end

    classDef current fill:#fcbf49,stroke:#8a5a00,stroke-width:3px,color:#111;
    class CDN,AGS,EVAL current;
```

## Changes

- Add a deterministic CDN-style UFC backend with API-key validation, `ETag`/`If-None-Match`, and metadata-only status inspection.
- Add a dormant `FEATURE_FLAGGING_AND_EXPERIMENTATION_AGENTLESS` scenario that starts the backend before the weblog.
- Allow the shared end-to-end lifecycle to omit the Datadog Agent and its proxy/interface handling.
- Run the agentless scenario without an Agent or Agent proxy and explicitly disable Remote Configuration.
- Add the default-agentless configuration/evaluation contract test.
- Gate the test in every SDK manifest and keep the dormant scenario out of broad `all`, `end_to_end`, and `tracer_release` selections.
- Add regression coverage for backend lifecycle, scenario topology, and CI scenario selection.

## Decisions

- Keep this PR deliberately inactive until an SDK-specific manifest PR opts in.
- Treat no-Agent execution as part of the agentless contract, not merely as a future dogfooding concern.
- Keep side-effect assertions out of this configuration-foundation PR.
- Put the serverless-sidecar/direct-fallback telemetry topology in the next stacked contract PR.
- Keep the Agent/RC scenario unchanged as the explicit alternative configuration path.

## Validation

- `./format.sh --check` — passes.
- Focused lifecycle and scenario tests — 6 passed.
- Forced Java run against `java@1.65.0-SNAPSHOT+f9e70e67c4` — 1 passed in 2.11s.
- The Java run starts no Agent or Agent proxy and waits on no Agent interface; it verifies an authenticated mock-CDN request and the expected flag evaluation.

## Next Steps

- Rework stacked PR #7299 around the serverless-sidecar-preferred/direct-fallback telemetry contract instead of `interfaces.agent`.
- Keep #7300 as the small Java configuration-manifest activation after #11892 and this harness are green together.
- Add separate Java manifest activations for each side-effect contract after the Java transport implementation passes the revised #7299 tests.


tests/ffe/test_agentless_configuration.py
```diff
@@ -0,0 +1,32 @@
...
+            "/ffe",
+            json={
+                "flag": "empty-targeting-key-flag",
+                "variationType": "STRING",
+                "defaultValue": "default",
+                "targetingKey": "user-1",
+                "attributes": {},
+            },
+        )
+        self.backend_status = scenarios.feature_flagging_and_experimentation_agentless.mock_backend_status()
```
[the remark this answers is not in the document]
leoromanovsky: Addressed in 5d5e341f9. Setup now only triggers the /ffe request; the backend status request and assertions run inside test_default_agentless_source, so control-endpoint errors are reported as test failures.

tests/ffe/test_agentless_configuration.py
```diff
@@ -0,0 +1,32 @@
+"""Test default agentless UFC delivery before FFE evaluation."""
+
+import json
+
+from utils import features, scenarios, weblog
+from utils.docker_fixtures._mock_ffe_agentless_backend import CONFIG_PATH
+
+
+@scenarios.feature_flagging_and_experimentation_agentless
```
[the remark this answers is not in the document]
leoromanovsky: Addressed in 5d5e341f9. The checked-in scenario map now includes this nodeid, and test_agentless_ffe_test_file verifies that changing this file selects FEATURE_FLAGGING_AND_EXPERIMENTATION_AGENTLESS. Production CI regenerates its map from the PR head, but the static regression fixture is now consistent too.

utils/_context/_scenarios/__init__.py
```diff
@@ -722,6 +727,21 @@ class _Scenarios:
         scenario_groups=[scenario_groups.ffe],
     )
 
+    feature_flagging_and_experimentation_agentless = FeatureFlaggingAgentlessEndToEndScenario(
+        "FEATURE_FLAGGING_AND_EXPERIMENTATION_AGENTLESS",
```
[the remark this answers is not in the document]
leoromanovsky: Addressed in 5d5e341f9. EndToEndScenario now has an explicit default-group opt-out, and the dormant agentless scenario registers only with the FFE group. The group-rules test verifies it is absent from all, end_to_end, and tracer_release while remaining explicitly selectable.

utils/_context/_scenarios/endtoend.py
```diff
@@ -21,6 +22,12 @@
     TestedContainer,
 )
 from utils._context.weblog_infrastructure import EndToEndWeblogInfra
+from utils.docker_fixtures._core import extra_hosts_for_environment
+from utils.docker_fixtures._mock_ffe_agentless_backend import (
```
cbeauchesne: There is a mechanism in our CI that control which scenario is executed based on changes. As now, `utils.docker_fixtures` triggers a change only on INTEGRATION_FRAMEWORKS and PARAMETRIC scenarios.

As `_mock_ffe_agentless_backend` will be a utilitity of both scenario, we must move this module in an dedicated module. 

I also feel that it should be generic, as I expect FFE won't be the only feature to use it.

WDYT about `utils.mocked_backend` ? For the CI orchestrator, it happens here : `utils/scripts/libraries_and_scenarios_rules.yml`

utils/_context/_scenarios/endtoend.py
```diff
@@ -374,8 +398,10 @@ def _start_interfaces_watchdog(self):
         open_telemetry_interfaces: list[ProxyBasedInterfaceValidator] = (
             [interfaces.open_telemetry] if self.include_opentelemetry else []
         )
+        agent_interfaces = [interfaces.agent] if self.include_agent else []
         super().start_interfaces_watchdog(
-            [interfaces.library, interfaces.agent]
+            [interfaces.library]
+            + agent_interfaces
```
cbeauchesne, line 398: The type of agent_interfaces is `list | None`, if I'm not wrong, you can't add `None` to a list here. The fox is just above, by setting `[]` if `include_agent==False`

utils/_context/_scenarios/endtoend.py
```diff
@@ -446,7 +474,7 @@ def _wait_and_stop_containers(self, *, force_interface_timout_to_zero: bool):
                 interfaces.library, 0 if force_interface_timout_to_zero else self.library_interface_timeout
             )
 
-            self.weblog_infra.stop()
+            self.weblog_infra.stop(flush=self._flush_weblog_on_stop)
```
cbeauchesne: Why do you need to disable the flush here ?

tests/test_the_test/test_mock_ffe_agentless_backend.py
```diff
@@ -66,3 +70,114 @@ def test_mock_ffe_agentless_backend_status_is_metadata_only(worker_id: str) -> N
         assert "body" not in status
     finally:
         server.close()
+
+
+@scenarios.test_the_test
+@features.not_reported
```
cbeauchesne: You don't need to add `@features.not_reported` here, same goes for other `test_the_test` files

utils/_context/_scenarios/endtoend.py
```diff
@@ -542,6 +573,8 @@ def __init__(
         backend_interface_timeout: int = 0,
         client_drop_p0s: bool | None = None,
         iast_enabled: bool = True,
+        include_agent: bool = True,
+        include_default_scenario_groups: bool = True,
```
cbeauchesne: IIUC, it's a temporary situation? If yes, could you rather force the groups in `FeatureFlaggingAgentlessEndToEndScenario.__init__`, rather adding an argument to the constructor? (it's already too big).

utils/_context/_scenarios/__init__.py
```diff
@@ -722,6 +727,24 @@ class _Scenarios:
...
+            "DD_REMOTE_CONFIGURATION_ENABLED": "false",
+        },
+        doc="Validate default agentless UFC delivery and evaluation without a Datadog Agent.",
+        include_agent=False,
+        include_default_scenario_groups=False,
+        flush_weblog_on_stop=False,
+        library_interface_timeout=0,
+        scenario_groups=[scenario_groups.ffe],
+        use_proxy_for_agent=False,
+        use_proxy_for_weblog=False,
```
cbeauchesne: If ever some of those value will be constant for the majority of `FeatureFlaggingAgentlessEndToEndScenario`, it's better to set the directly inside `FeatureFlaggingAgentlessEndToEndScenario` constructor, or inside `FeatureFlaggingAgentlessEndToEndScenario` construtor prototype as optional argument with defaults.

It keeps this file shorter and easier to maintain.

utils/_context/_scenarios/endtoend.py
```diff
@@ -574,16 +612,72 @@ def __init__(
...
 
+class FeatureFlaggingAgentlessEndToEndScenario(DdTraceEndToEndScenario):
+    """FFE end-to-end scenario with UFC available before the weblog starts."""
+
+    _mock_backend: MockFFEAgentlessBackendServer | None = None
+    _last_mock_backend_status: MockFFEAgentlessBackendStatus | None = None
+
+    def configure(self, config: pytest.Config) -> None:
+        try:
+            if not self.replay:
```
cbeauchesne, line 650: I'm not sure it's usefull, we never use xdist with end-to-end scenario.

cbeauchesne (changes requested):
Great job, congrats. I've few request changes and question. There are also two missing bits : 

* add rules in `utils/scripts/libraries_and_scenarios_rules.yml` for this new scenario
* add the scenario in the CI

utils/_context/_scenarios/endtoend.py
```diff
@@ -578,12 +607,94 @@ def __init__(
...
             weblog_categories=[WeblogCategory.dd_trace],
             weblog_env=weblog_env,
             weblog_volumes=weblog_volumes,
         )
 
 
+class FeatureFlaggingAgentlessEndToEndScenario(DdTraceEndToEndScenario):
+    """FFE end-to-end scenario with UFC available before the weblog starts."""
+
+    _default_scenario_groups: tuple[ScenarioGroup, ...] = ()
```
cbeauchesne, line 617: So, with that, the scenario is not activated in the official system-test workflow, unless you explicitly require it. 

In  system-tests CI if you modify a test file that runs on this scenario, or this present file, or something in `utils/mocked_backend/ffe.py`, it will be executed. For any other modification (many system-tests internals), it won't be executed.

Is it what you want ?

cbeauchesne (approved):
A small non blocking comment.

Great piece of work, congrat !

PR #7308, opened 2026-07-14
Fix system-tests parametric server for python RC changes

bwoebi:
The RC client was rewritten on top of libdatadog: the native client owns the current file set and no longer keeps a Python-side applied-config cache, hence we reconstruct that manually here.

Detect which client API is in use at runtime (native client exposes _ensure_native) and keep the old client.request()/_applied_configs path as a fallback, so the parametric server works against tracer versions from both before and after the rewrite.


emmettbutler (approved):
ack

PR #7310, opened 2026-07-14
Add retry for downloading install_script_agent

sarahchen6:
## Motivation

`dd-trace-java` has had multiple flaky failures recently due to the install_script_agent failing to download, likely due to flaky infra issues. e.g.
* https://gitlab.ddbuild.io/DataDog/apm-reliability/dd-trace-java/-/jobs/1857656432
* https://gitlab.ddbuild.io/DataDog/apm-reliability/dd-trace-java/-/jobs/1851603017
* https://gitlab.ddbuild.io/DataDog/apm-reliability/dd-trace-java/-/jobs/1858417690

## Changes

Add a retry mechanism when pulling the install script. Fail if the script is not pulled - there's no point in continuing with the test if there's no agent installed.

## Workflow

## Reviewer checklist


robertomonteromiguel (approved):

utils/build/ssi/base/install_script_ssi.sh
```diff
@@ -42,11 +42,25 @@ fi
 
 if [ -f "install_script_agent7.sh" ]; then
     echo "[TRACE] install_script_agent7.sh exists"
-    DD_REPO_URL=${DD_injection_repo_url}  DD_INSTALL_ONLY=true DD_APM_INSTRUMENTATION_ENABLED=host  bash -c "$(cat install_script_agent7.sh)"
```
maattdd, line 45: Could we extract a "download_with_retry(url)" function and use it in both places?

utils/build/ssi/base/install_script_ssi.sh
```diff
@@ -42,11 +42,25 @@ fi
...
 else
-    DD_REPO_URL=${DD_injection_repo_url}  DD_INSTALL_ONLY=true DD_APM_INSTRUMENTATION_ENABLED=host  bash -c "$(curl -L https://dd-agent.s3.amazonaws.com/scripts/install_script_agent7.sh)"
+    for attempt in 1 2 3 4 5; do
+        echo "[TRACE] downloading install_script_agent7.sh (attempt ${attempt})"
+        if curl --fail --retry 3 --retry-delay 2 -sSL -o install_script_agent7.sh https://dd-agent.s3.amazonaws.com/scripts/install_script_agent7.sh && [ -s install_script_agent7.sh ]; then
+            break
+        fi
+        echo "[TRACE] download failed or produced an empty file; retrying"
+        rm -f install_script_agent7.sh
+        sleep 2
```
bric3: **suggestion:** I suggest to sleep a "exponential backoff **with jitter**"

* exponential backoff, will add delay on each retry (no point in hammering the server if it's not available on a retry)
* jitter, if all jobs are stuck for the same reason, let's not hammer the server that the same time, hence a add a small random number

utils/build/ssi/base/install_script_ssi.sh
```diff
@@ -42,11 +42,25 @@ fi
 
 if [ -f "install_script_agent7.sh" ]; then
     echo "[TRACE] install_script_agent7.sh exists"
-    DD_REPO_URL=${DD_injection_repo_url}  DD_INSTALL_ONLY=true DD_APM_INSTRUMENTATION_ENABLED=host  bash -c "$(cat install_script_agent7.sh)"
 else
-    DD_REPO_URL=${DD_injection_repo_url}  DD_INSTALL_ONLY=true DD_APM_INSTRUMENTATION_ENABLED=host  bash -c "$(curl -L https://dd-agent.s3.amazonaws.com/scripts/install_script_agent7.sh)"
+    for attempt in 1 2 3 4 5; do
+        echo "[TRACE] downloading install_script_agent7.sh (attempt ${attempt})"
+        if curl --fail --retry 3 --retry-delay 2 -sSL -o install_script_agent7.sh https://dd-agent.s3.amazonaws.com/scripts/install_script_agent7.sh && [ -s install_script_agent7.sh ]; then
```
mhdatie: Why do we need to rely on the `attempt` loop when `--retry 3 --retry-delay 2` of `curl` offers the same mechanism?

bric3 (approved):
Pre-approving, but I agree having dedicated functions could be useful. The backoff with jitter might be useful as well.

mhdatie:
The premise of the PR makes sense to me - retrying for the install step to make this step less error prone. 

The motivation however will not always be the case going forward:
> ... - there's no point in continuing with the test if there's no agent installed.
 
I believe the FFE team are adding a mechanism to run system tests without an agent installed. I just want to call this out in case it interferes with the new system tests setup being worked on this week https://github.com/DataDog/system-tests/pull/7298

utils/build/ssi/base/install_script_ssi.sh
mhdatie: Reading up on retry and retry-delay, `retry` seems to bake exponential backoff by default, but when using `retry-delay` it will fixate it to 2 seconds (seeing @bric3's comment on this now too)

sarahchen6: Yeah the attempt loop is an extra safe-guard! It'll try the `curl --fail --retry 3 --retry-delay 2` command five times with a debug line to document each time.

utils/build/ssi/base/install_script_ssi.sh
sarahchen6: Makes sense! Done in 5f44ae2

utils/build/ssi/base/install_script_ssi.sh
sarahchen6: Sure! Done in 5f44ae2

utils/build/ssi/base/install_script_ssi.sh
mhdatie: What do we gain from the extra guard compared to retrying the curl command 3 times (totaling 4 attempts) only? Do we know that it constantly fails after 4 attempts?

utils/build/ssi/base/install_script_ssi.sh
mhdatie: My second comment here https://github.com/DataDog/system-tests/pull/7310#discussion_r3587615149 addresses the exponential backoff part without having to implement it yourself.

You can leverage it by removing `--retry-delay 2` from the curl command. If you intend to keep the outer loop, then I think the backoff should be present on the curl call, and the outer retry loop can exist with the simple sleep. It will make the code simpler too.

mhdatie (approved):
LGTM, thanks for addressing all the comments!

utils/build/ssi/base/install_script_ssi.sh
sarahchen6: Sounds good, exponential backoff should be enough, so I'll remove the jitter implementation as well - done in 8badeca

utils/build/ssi/base/install_script_ssi.sh
sarahchen6: The loop guard emits a debug line with each retry, but the `--retry 3` choice was arbitrary - just taken from other examples throughout the repo 🙂

PR #7311, opened 2026-07-14
[codex] build Docker SSI images with Rebuildr

pawelchcki:
## Summary

- replace manual Docker SSI image handling with separate cached and per-run Rebuildr DAGs
- pin reusable image identity to the base registry digest and tracked build inputs
- enforce CI/local publication policy and update CLI/wizard behavior
- capture Buildx Bake output and add focused DAG, cache, and publication tests

## Why

The previous flow used mutable tags, manual existence checks, Docker SDK builds, and teardown publication. That made cache invalidation implicit and allowed stale reusable images. Rebuildr provides content-addressed cache identities while keeping tracer and injector installation intentionally movable.

## Impact

GitLab CI always publishes cache-aware reusable images and now fails early without `PRIVATE_DOCKER_REGISTRY`. Local runs only read or build by default; `-P` explicitly opts into publication. The removed `-B` option is no longer needed because tracked inputs invalidate the cached graph automatically.

## Validation

- `./format.sh`
- `venv/bin/pytest -q --scenario TEST_THE_TEST tests/test_the_test/test_docker_ssi_image_builder.py` (12 passed)
- cold and warm Python 3.12.7 / Ubuntu 22.04 AMD64 `DOCKER_SSI` smoke runs (3 passed each, with expected skip/xfail/xpass outcomes)
- warm run confirmed the reusable DAG was skipped while SSI/weblog rebuilt with a fresh nonce


robertomonteromiguel:
**File:** `utils/docker_ssi/image_builder.py` (line 70)

🗿 🤖 🔴

**[Sphinx Review — HIGH]** The downloaded installer script is deleted when configure()'s `with self._installer_script()` block exits (line 70/172), but the rebuildr graphs that reference that file are not built until the separate build_weblog() call. In runs that must download the installer (the default local-dev path, since binaries/ is gitignored and only GitLab CI pre-populates it), the file is gone before cached_root.build/push and movable_root.build run, so the buildx bake context can reference a non-existent path.

**Suggestion:** Keep the installer file alive for the whole build lifetime, not just the load() call: have configure() enter the _installer_script() context and store it (e.g. via an ExitStack on the instance closed after build_weblog()), or download into a stable location with the same lifetime as cached_root/movable_root. Add a test covering the download branch that asserts the file still exists when the (mocked) build runs.

robertomonteromiguel:
**File:** `tests/test_the_test/test_docker_ssi_image_builder.py` (line 145)

🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** The build/publish execution path is fully mocked in every test: rebuildr's DockerBakeBuilder, pull_image, image_exists_locally, ImageHandle._verify_local/_materialize_local/cache_location and get_docker_client are stubbed, so no Dockerfile COPY/RUN or install script is ever executed under CI. The suite verifies graph structure and identity hashing only. A regression in a Dockerfile COPY path, the rebuildr File(src,dst) mapping, or an install script (e.g. install_script_ssi_installer.sh's new relative `cat install_script_agent7.sh`) would not be caught by TEST_THE_TEST and would surface only at DOCKER_SSI runtime.

**Suggestion:** Add at least one integration-style check that a real (or lightweight) build of the SSI/weblog Dockerfiles succeeds for one language/arch, or a static assertion that every File(dst) referenced by the two .rebuildr.py graphs maps to a path the corresponding Dockerfile COPY resolves, so context/path regressions are caught in CI rather than only in manual smoke runs.

robertomonteromiguel:
**File:** `utils/docker_ssi/image_builder.py` (line 108)

🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** The builder communicates with the third-party rebuildr graph files exclusively by snapshotting, mutating and restoring ~16 global os.environ keys around load(). The contract is a set of duplicated bare string literals split across three files with no shared constants; a rename/typo on either side fails silently (KeyError or empty-string build arg), the `value or ''` coalescing collapses None into '' during the load window, and the pattern is not reentrant/thread-safe if two builders ever run in one process. No test exercises the restore or load-failure cleanup path.

**Suggestion:** Prefer passing configuration to rebuildr explicitly (a params/config object) instead of process-global env. If env is the only supported input, at minimum define the DOCKER_SSI_* key names as shared module-level constants imported by both image_builder.py and the .rebuildr.py files, and add a test asserting the pre-existing environment is fully restored after a load() failure.

robertomonteromiguel:
**File:** `utils/docker_ssi/image_builder.py` (line 162)

🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** The download branch of _installer_script writes a temp file into the git working tree root (dir=self._root_dir) and cleans up only via try/finally, which is skipped on hard kills (SIGKILL/OOM) during the long build that runs inside the yielded `with` block. Because the local-fixture candidates live under gitignored binaries/* and are absent from a clean checkout, this download branch is the default runtime path, yet no test covers download-or-cleanup (tests inject installer.sh straight into _load_projects).

**Suggestion:** Materialize the downloaded installer under a dedicated gitignored/scratch location (e.g. host_log_folder or a tempdir) and expose it to the rebuildr context via that path rather than the repo root, or add the temp prefix to .gitignore and register an atexit/signal-safe cleanup. Add a test for the download + cleanup branch (including download failure).

robertomonteromiguel:
**File:** `utils/docker_ssi/image_builder.py` (line 157)

🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** The installer 'local fixture' fast-path in _installer_script never matches on a normal CI run: utils/build/ssi/base/binaries/ is not committed and root binaries/ is gitignored; that directory is only created transiently by the interactive wizard (docker_ssi_wizard.sh:287/297). So the fall-through branch (mkstemp in the repo root + urlretrieve from S3) is the production default, not an edge case. This moves the installer fetch from a cacheable Docker build layer (old install_script_ssi_installer.sh curl) to an uncached host-side download on every scenario configure, and it is the branch exercised by zero tests.

**Suggestion:** Either make the build pipeline populate utils/build/ssi/base/binaries/install_script_agent7.sh so CI hits the cached fast-path, or acknowledge the download branch is the default and add a test that exercises _installer_script's download+cleanup lifecycle. Also consider caching the downloaded script across runs rather than re-fetching on every configure().

robertomonteromiguel:
**File:** `utils/docker_ssi/image_builder.py` (line 178)

🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** _run wraps rebuildr's build/push in contextlib.redirect_stdout/redirect_stderr targeting docker_build.log, but these only rebind Python's sys.stdout/sys.stderr — they do not redirect the OS file descriptors of the `docker buildx bake` subprocess rebuildr spawns. Actual build/push output therefore bypasses docker_build.log (only the header lines land there), so the 'inspect {log}' error message points at a nearly-empty log, making CI build failures hard to triage — a regression from the prior code that streamed build logs to file.

**Suggestion:** Capture the subprocess output at the fd level (pass explicit stdout/stderr targets or the log file's fd through rebuildr, or duplicate fds 1/2 to the log). If rebuildr can't do that, at minimum include the captured output / str(error.__cause__) (and a tail of the build log) in the raised DockerSSIImageError so the failure reason survives to the console.

robertomonteromiguel:
**File:** `utils/docker_ssi/image_builder.py` (line 139)

🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** On _resolve_base_image inspect failure the captured stderr is discarded. subprocess.run uses capture_output=True so the real docker error is in error.stderr, but the raised DockerSSIImageError only contains a generic 'Failed to resolve registry digest' message and CalledProcessError's str does not include stderr. CI failures here will be hard to diagnose.

**Suggestion:** Include the captured stderr in the raised message, e.g. append getattr(error, 'stderr', '') (guarding for OSError which has no stderr) so the docker/buildx failure reason is visible without inspecting __cause__.

robertomonteromiguel:
**File:** `utils/docker_ssi/image_builder.py` (line 74)

🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** The publication decision '("GITLAB_CI" in os.environ or self.push_base_images)' is duplicated: configure() (line 66) uses it to decide whether PRIVATE_DOCKER_REGISTRY is mandatory, and build_weblog() (line 74) independently recomputes it to choose push vs build. There is no single source of truth, so the validation guard and the action it guards can silently drift apart (e.g. a future edit adding another CI marker to one but not the other).

**Suggestion:** Extract the policy into one place, e.g. a `@property _should_publish(self) -> bool: return "GITLAB_CI" in os.environ or self.push_base_images`, and have both configure()'s guard and build_weblog() consume it.

robertomonteromiguel:
**File:** `utils/docker_ssi/image_builder.py` (line 167)

🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** The installer download branch is the DEFAULT path for local runs, not a rare fallback: on a clean checkout neither candidate exists (only GitLab CI populates binaries/ from S3 pinned to DD_INSTALL_SCRIPT_VERSION). Local runs therefore always urlretrieve the UNVERSIONED latest install_script_agent7.sh, while CI bakes a pinned version — so local DOCKER_SSI outcomes are not reproducible against CI, and the fetched-then-baked script (executed as root at build time) is neither version-pinned nor integrity-checked.

**Suggestion:** Pin the local download to the same DD_INSTALL_SCRIPT_VERSION CI uses (make _INSTALLER_URL version-aware), and/or verify a known checksum of the downloaded script before baking it into the image so local and CI runs exercise an identical, verified installer.

robertomonteromiguel:
**File:** `utils/docker_ssi/image_builder.py` (line 129)

🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** _resolve_base_image unconditionally shells out to `docker buildx imagetools inspect` against the remote registry on every scenario configure, with a 120s timeout and no fallback to a locally-present base image. Any registry/network flakiness (or offline development) now hard-fails configuration via DockerSSIImageError, whereas the prior implementation only attempted a registry pull as an existence probe and tolerated its absence. No test covers the success or failure branch.

**Suggestion:** Add a local fallback: if the remote inspect fails, resolve the digest from a locally available image (e.g. `docker image inspect`) before raising, or make the remote pin conditional on publication/CI so pure-local runs proceed offline. Add tests covering the digest-parsing success path and the network-failure path.

robertomonteromiguel:
🗿 🤖 🔴

**[Sphinx Review — LOW]** This PR is not framework-only: it rewrites the Docker SSI image build and publication semantics (cache identity, movable nonce, base-image digest pinning, CI/local publication gating), changes how the DD installer is obtained (`cat` of a context-copied script vs curl at build time), adds an SSI_BUILD_NONCE cache-buster, and removes the -B/--ssi-force-build run option. A framework/scenario-level approval is insufficient; a reviewer who owns the SSI installer/injector build flow should validate the new Rebuildr DAG behavior end-to-end.

**Suggestion:** Request explicit sign-off from an owner of the Docker SSI image-building / injector install flow (R&P team / #apm-shared-testing), covering the removed -B option, the installer-fetch change, the cache-invalidation semantics, and the new fail-early publication policy; confirm warm/cold cache and publish/no-publish behavior on a real registry before merge.

robertomonteromiguel:
**File:** `utils/docker_ssi/image_builder.py` (line 125)

🗿 🤖 🔴

**[Sphinx Review — LOW]** No test in this PR would detect removing the `+ "_"` concatenation in `resolve_runtime_version(...) + "_"` in _base_tag(), which would produce a malformed tag like "ubuntu_24_04_3.12.7linux_amd64" instead of "ubuntu_24_04_3.12.7_linux_amd64".

**Suggestion:** Add an assertion (e.g. to test_rebuildr_graphs_and_cached_identity) validating the computed base tag format, i.e. that proper separators appear between the base image, runtime, and architecture components.

robertomonteromiguel:
**File:** `tests/test_the_test/test_docker_ssi_image_builder.py` (line 133)

🗿 🤖 🔴

**[Sphinx Review — LOW]** Unlike the other publication tests in this file, test_warm_cache_skips_cached_graph_but_runs_movable_graph does not neutralize the GITLAB_CI environment variable. build_weblog() computes `publish = "GITLAB_CI" in os.environ or self.push_base_images`, so if this test runs where GITLAB_CI is set, the cached graph is pushed instead of built and the bake-definition assertions no longer hold — making the test environment-dependent.

**Suggestion:** Add `monkeypatch.delenv("GITLAB_CI", raising=False)` (and ensure push_base_images is False) at the top of the test so it exercises the build (non-publish) path deterministically regardless of the ambient environment.

robertomonteromiguel:
**File:** `utils/_context/_scenarios/docker_ssi.py` (line 92)

🗿 🤖 🔴

**[Sphinx Review — LOW]** Comment block still describes the removed manual 3-step Docker build flow ('2. Build the ssi installer image ... This image will be push in the registry'), which no longer matches the Rebuildr cached/movable DAG implementation introduced by this PR.

**Suggestion:** Update the comment to describe the two Rebuildr graphs (cached reusable installer image and per-run movable SSI+weblog image) and the CI/local publication policy, or remove the outdated step list.

robertomonteromiguel:
**File:** `utils/scripts/ssi_wizards/docker_ssi_wizard.sh` (line 225)

🗿 🤖 🔴

**[Sphinx Review — LOW]** select_image_publication treats a non-empty PRIVATE_DOCKER_REGISTRY as proof that publication is configured, but configure_private_registry (choice 2) exports the variable before verifying `docker login` succeeded. On a failed login the guard still passes, -P is appended, and the run fails later at publish time instead of failing fast here.

**Suggestion:** Have configure_private_registry only export PRIVATE_DOCKER_REGISTRY after a confirmed successful login (or unset it on login failure), so the fast-fail guard in select_image_publication is meaningful.

robertomonteromiguel:
**File:** `utils/build/ssi/base/install_script_ssi_installer.sh` (line 3)

🗿 🤖 🔴

**[Sphinx Review — LOW]** install_script_ssi_installer.sh was changed to unconditionally run `bash -c "$(cat install_script_agent7.sh)"` with no existence check and no curl fallback, while the sibling install_script_ssi.sh still guards with `if [ -f install_script_agent7.sh ]` and falls back to curl. The two scripts now handle the identical vendored-installer dependency inconsistently, install_script_ssi.sh's curl branch is now effectively dead code, and a future context-mapping regression would fail loudly in one script but produce a silent no-op `bash -c ""` (RUN exits 0) in the other.

**Suggestion:** Make the two scripts consistent: either add the same `if [ -f install_script_agent7.sh ]` guard (failing hard on absence rather than silently no-op'ing) to install_script_ssi_installer.sh, or drop the now-unreachable curl fallback from install_script_ssi.sh so both assume the vendored file is present.

robertomonteromiguel:
**File:** `utils/docker_ssi/image_builder.py` (line 51)

🗿 🤖 🔴

**[Sphinx Review — LOW]** cached_root/movable_root are field(init=False) with no default, encoding an undocumented, unguarded two-phase lifecycle (construct -> configure -> build_weblog). If build_weblog() is called before configure(), the failure is an opaque AttributeError on 'cached_root' rather than a clear 'configure() must be called first' error.

**Suggestion:** Either default these fields to None with an explicit guard at the top of build_weblog() (raise DockerSSIImageError('configure() must run before build_weblog()')), or fold the base-image resolution/graph loading into build_weblog() so there is no separately-observable half-initialized state.

robertomonteromiguel:
**File:** `tests/test_the_test/test_docker_ssi_image_builder.py` (line 161)

🗿 🤖 🔴

**[Sphinx Review — LOW]** The test asserts against rebuildr's internal Bake target names 'rebuildr0_ssi'/'rebuildr1_weblog', which are derived from rebuildr's own index+name scheme rather than any identifier this repo controls. This couples the test to a pinned library's internal naming and image ordering; a rebuildr version bump or image reorder breaks the assertion even when build behavior is correct.

**Suggestion:** Assert on a stable, repo-owned property instead (e.g. that both movable images are present by their own names/repositories) or derive the expected target set from the loaded graph.

robertomonteromiguel:
**File:** `utils/docker_ssi/image_builder.py` (line 58)

🗿 🤖 🔴

**[Sphinx Review — LOW]** No test in this PR would detect changing `is_absolute()` to `is_relative()` in the ternary for _log_folder initialization, which would cause relative log_folder values to remain relative instead of being rooted at _root_dir. The test always passes an absolute path (str(root / "logs")), masking the bug for relative-path cases.

**Suggestion:** Add a test case (or parameterize an existing test) that passes a relative host_log_folder to DockerSSIImageBuilder and asserts the resulting _log_folder is absolute and rooted at root_dir.

robertomonteromiguel:
**File:** `utils/docker_ssi/image_builder.py` (line 78)

🗿 🤖 🔴

**[Sphinx Review — LOW]** build_weblog passes the raw return value of movable_root.build() straight into get_docker_client().images.get() with no validation that it is a resolvable image reference. This couples correctness to an unpinned assumption about rebuildr's ImageHandle.build() return type, and every test mocks images.get so a rebuildr API change would surface only at runtime.

**Suggestion:** Assert final_image is a non-empty string (or the expected type) before calling images.get, and raise DockerSSIImageError with context otherwise, so an unexpected rebuildr API shape fails with a clear message rather than an opaque docker SDK error.

PR #7316, opened 2026-07-16
FFL-2783: add tests for observeFullEvaluationData PII protection (gated as missing_feature)

vjfridge:
## Motivation

[FFL-2783](https://datadoghq.atlassian.net/browse/FFL-2783) — part of [FFL-2780](https://datadoghq.atlassian.net/browse/FFL-2780) (Protecting PII in flagevaluations track)

See [informal RFC](https://docs.google.com/document/d/19VIf4B9p-zsAL2uWSvdzil59DZlGjQm4yudAMwJBqLs/edit?tab=t.0#heading=h.knfs9oz5axeh) for more context

## Summary
- Adds three test classes to existing `tests/ffe/test_flag_eval_evp.py` covering the new per-env UFC `observeFullEvaluationData` field
- Extends `make_ufc_fixture` with an `observe_full_evaluation_data: bool | None = None` kwarg. `None` omits the field entirely, so existing tests are unaffected
- All three classes are deactivated as `missing_feature (FFL-2783)` on Go — the only SDK where the file-level gate would otherwise activate them. Other SDK manifests already deactivate the whole file via `missing_feature (FFL-2446)` or `irrelevant`.

## Truth-table coverage

| UFC `observeFullEvaluationData` | Expected `targeting_key` | Expected `context.evaluation` |
|---|---|---|
| absent | `sha256_`-prefixed (71 chars, lowercase hex suffix) | omitted |
| `false` | `sha256_`-prefixed | omitted |
| `true` | raw customer value, no prefix | full object with all attributes |

## Canonical PII vector (shared across SDKs)

Every SDK's unit tests should assert against the same input/output. Input: `"jane.doe@datadoghq.com"` → `"sha256_b4698f9b6d186781fa8dc59e533578fa2d8379a46b1cf6db85cda6aa9c99e51b"`. Canonical attributes cover string, integer, dotted-key, and email PII shapes.

## Forward-compat for existing SDKs

Introducing a new field inside the UFC `environment` object could theoretically break SDKs if their UFC parser rejects unknown fields. **That risk is already covered** by [`tests/ffe/test_dynamic_evaluation.py::Test_FFE_Unknown_Fields_Tolerance`](../blob/main/tests/ffe/test_dynamic_evaluation.py), which injects random unknown fields at every UFC level (including inside `environment`) and asserts each SDK continues to evaluate flags correctly. That test runs on every SDK where `test_dynamic_evaluation.py` is active, so no per-fixture retrofit is needed here.

## Out of scope
- Existing kill switch (`DD_FLAGGING_EVALUATION_COUNTS_ENABLED`) — per-SDK unit test territory, no `system-tests` scenario variant needed.

## Test plan
- [x] `./format.sh` clean (mypy, ruff, node linters)
- [ ] Once Go SDK ships FFL-2790, flip Go's three `missing_feature (FFL-2783)` rows to `v2.X.Y-dev` and confirm all three cells pass

[FFL-2783]: https://datadoghq.atlassian.net/browse/FFL-2783?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FFL-2780]: https://datadoghq.atlassian.net/browse/FFL-2780?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/ffe/test_flag_eval_evp.py
```diff
@@ -489,3 +489,158 @@ def test_ffe_evp_flagevaluation_degradation(self) -> None:
...
+    "region": "us-east-1",
+    "account.tier": "gold",
+}
+
+
+def _find_pii_event(flag_key: str) -> tuple[JSON, JSON]:
+    wait_for_evp_flagevaluation_event(flag_key)
+    events = find_evp_flagevaluation_events(flag_key)
+    assert events, f"Expected EVP flagevaluation event for flag {flag_key}"
+    return events[0]
```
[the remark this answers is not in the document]
vjfridge: Addressed in a210c39f0 — removed the `_find_pii_event` helper. All three PII tests now iterate every matching event returned by `find_evp_flagevaluation_events` and apply all assertions to each event, so a clean first event can no longer mask a later leaky one.

tests/ffe/test_flag_eval_evp.py
```diff
@@ -489,3 +489,158 @@ def test_ffe_evp_flagevaluation_degradation(self) -> None:
...
+class Test_FFE_EVP_Flagevaluation_ObserveFullData_Absent_Hashed:
+    """Test that when observeFullEvaluationData is absent from UFC, targeting_key is hashed and context.evaluation is omitted (default PII-protection)."""
+
+    def setup_ffe_evp_flagevaluation_observe_full_data_absent(self) -> None:
+        config_id = "ffe-evp-observe-absent"
+        self.flag_key = "evp-observe-absent-flag"
+        rc.tracer_rc_state.reset().set_config(
+            f"{RC_PATH}/{config_id}/config",
+            make_ufc_fixture(self.flag_key),
+        ).apply()
```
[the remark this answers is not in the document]
vjfridge: Not addressing this one. The other eight test classes in this file (`Test_FFE_EVP_Flagevaluation_Basic`, `_Count`, `_Context_Bounds`, `_Runtime_Default`, `_Load_Aggregation`, `_Burst_Aggregation`, `_High_Cardinality_Aggregation`, `_Degradation`) all place `rc...apply()` and `evaluate_flag(...)` in `setup_*` and only run `assert self.r.status_code == 200` in `test_*`. Diverging in only the three new classes would create an inconsistency reviewers would flag; if the setup-vs-test split needs enforcement here, it belongs in a dedicated refactor across the whole file.

tests/ffe/test_flag_eval_evp.py
```diff
@@ -489,3 +489,158 @@ def test_ffe_evp_flagevaluation_degradation(self) -> None:
...
+        hex_suffix = targeting_key[len("sha256_") :]
+        assert len(hex_suffix) == 64, f"hashed targeting_key suffix must be 64 chars: {targeting_key!r}"
+        assert all(c in "0123456789abcdef" for c in hex_suffix), (
+            f"hashed targeting_key suffix must be lowercase hex: {targeting_key!r}"
+        )
+
+        context = event.get("context")
+        if isinstance(context, dict):
+            assert "evaluation" not in context, (
+                f"context.evaluation must be omitted when observeFullEvaluationData is absent: {context}"
```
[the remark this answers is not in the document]
vjfridge: Addressed in a210c39f0 — added `assert_no_raw_pii_in_event` that walks the serialized event and asserts none of the raw PII values (targeting key + every PII attribute value) appear anywhere in the payload. Now applied to both hashed cells (absent + false).

tests/ffe/test_flag_eval_evp.py
```diff
@@ -489,3 +528,142 @@ def test_ffe_evp_flagevaluation_degradation(self) -> None:
             assert "targeting_key" not in event, f"degraded event must omit targeting_key: {event}"
             assert object_key(event.get("variant"), "variant") == "on"
             assert object_key(event.get("allocation"), "allocation") == "default-allocation"
+
+
+@scenarios.feature_flagging_and_experimentation
+@features.feature_flags_evp_flagevaluation
+@pytest.mark.skip_if_xfail
```
cbeauchesne: There are semantic decorators in utils that allow to achieve the same, and also declare *why* we want to skip them : `@slow` and `@scenario_crash`, could you use them ?

cbeauchesne (dismissed):
From framework usage, just a small request change.

Cold you also get a review from someone familiar with the feature for the test logic ?

tests/ffe/test_flag_eval_evp.py
vjfridge: Thanks for raising this! You're right that using `@pytest.mark.skip_if_xfail` directly is too raw. We looked at `@slow` and `@scenario_crash` but neither quite fits — these tests aren't slow, and they don't crash the scenario. They're gated on SDK implementation status.

So we added a new alias `@not_yet_implemented` in `utils/_decorators.py` (same underlying `skip_if_xfail` mark, just with a clearer name) and used it consistently across all 11 classes in the file.

tests/ffe/test_flag_eval_evp.py
```diff
@@ -371,7 +410,7 @@ def test_ffe_evp_flagevaluation_load_aggregation(self) -> None:
 
 @scenarios.feature_flagging_and_experimentation
 @features.feature_flags_evp_flagevaluation
-@pytest.mark.skip_if_xfail
+@not_yet_implemented
```
leoromanovsky: hehe @cbeauchesne skipped these because they are not set up correctly; I needed to break system test convention to get them to work with PHP, which has its own sidecar that needs an additional await to count received EVP events in the backend. It's on my backlog to address, but a more correct `skip` decorator would be `flaky` or `bug`

tests/ffe/test_flag_eval_evp.py
```diff
@@ -23,6 +23,18 @@
 EVP_FULL_TIER_PER_FLAG_CAP = 10_000
 EVP_DEGRADATION_OVERFLOW_EVALS = 2_000
 
+# Fixed input/output vector for the PII-protection tests. Every SDK's L1 unit tests
```
leoromanovsky:
```suggestion
# Fixed input/output vector for the PII-protection tests. Every SDK's unit tests
```

codegen tool leaked 😄

tests/ffe/test_flag_eval_evp.py
```diff
@@ -23,6 +23,18 @@
...
+# should assert against the same input to prove byte-identical hashing across SDKs.
+PII_TARGETING_KEY = "jane.doe@datadoghq.com"
+PII_TARGETING_KEY_HASHED = "sha256_b4698f9b6d186781fa8dc59e533578fa2d8379a46b1cf6db85cda6aa9c99e51b"
```
leoromanovsky, line 29: I heard you mention this, good idea adding the prefix that will make it look less surprising when customers see it and I can see how this will help the autocomplete design to ignore it.

tests/ffe/test_flag_eval_evp.py
```diff
@@ -489,3 +528,142 @@ def test_ffe_evp_flagevaluation_degradation(self) -> None:
             assert "targeting_key" not in event, f"degraded event must omit targeting_key: {event}"
             assert object_key(event.get("variant"), "variant") == "on"
             assert object_key(event.get("allocation"), "allocation") == "default-allocation"
+
```
leoromanovsky, line 526: Please add a test for `DD_FLAGGING_EVALUATION_COUNTS_ENABLED=false` + `observeFullEvaluationData=true` - sorry I missed adding coverage for the false case.

leoromanovsky (approved):
logic is good, before merging please:

* add a env var=false case
* adjust labels on existing test cases

tests/ffe/test_flag_eval_evp.py
nccatoni: Why is it not enough to mark the tests as xfails in that case?

tests/ffe/test_flag_eval_evp.py
vjfridge: Happy to add coverage for `DD_FLAGGING_EVALUATION_COUNTS_ENABLED =false` + `observeFullEvaluationData=true`, but I think it lives better at L1 unit tests/L2 SDK integration layers. 

The assertion is `"SDK-level killswitch overrides UFC opt-in"` which is purely internal to the SDK and doesn't exercise anything system-level. Adding an L3 scenario here means a new weblog boot per SDK (~minutes of CI × N languages) to verify the boolean killswitch. 

I propose we cover this in each SDK's unit or integration suite instead. Let me know if you have strong feelings otherwise!

tests/ffe/test_flag_eval_evp.py
vjfridge: Ah thanks! Fixed!

tests/ffe/test_flag_eval_evp.py
vjfridge: Thanks for raising Leo! I looked into this more carefully. The existing `@flaky`/`@bug` decorators are all-or-nothing across SDKs. Per-SDK skipping lives in the manifests, so I think it would be best to wait until PHP gets enabled and then add class-level manifest entries in `manifests/php.yml`.

For now I've added a docstring note to `Burst`, `HighCardinality`, and `Degradation` pointing to `FFL-2529` and the manifest pattern, so whenever we drop the file-level `missing_feature` for PHP we will know to add those entries at the same time. 

I left `@not_yet_implemented` on all three for now since that's still the accurate for PHP and most SDKs. Let me know if this works for you or if you were looking for something different!

vjfridge:
## Local system-tests validation — Java SDK ✅

Ran the three new `observeFullEvaluationData` PII-protection tests locally against [dd-trace-java#12042](https://github.com/DataDog/dd-trace-java/pull/12042) (`java@1.65.0-SNAPSHOT+81af153ca7`).

**All three pass:**

| Test | Result |
|---|---|
| `Test_FFE_EVP_Flagevaluation_ObserveFullData_Absent_Hashed` | ✅ PASS |
| `Test_FFE_EVP_Flagevaluation_ObserveFullData_False_Hashed` | ✅ PASS |
| `Test_FFE_EVP_Flagevaluation_ObserveFullData_True_Unhashed` | ✅ PASS |

```
36 passed, 8 skipped, 2704 deselected, 1 xfailed, 3 xpassed in 301.86s
```

**Run command:**
```sh
./run.sh +l java FEATURE_FLAGGING_AND_EXPERIMENTATION \
  -F tests/ffe/test_flag_eval_evp.py::Test_FFE_EVP_Flagevaluation_ObserveFullData_Absent_Hashed \
  -F tests/ffe/test_flag_eval_evp.py::Test_FFE_EVP_Flagevaluation_ObserveFullData_False_Hashed \
  -F tests/ffe/test_flag_eval_evp.py::Test_FFE_EVP_Flagevaluation_ObserveFullData_True_Unhashed
```

Note: an earlier build (`d65c79f266`) failed `True_Unhashed` — the SDK was always hashing regardless of the flag. Root cause was `FlagEvaluationWriterImpl.buildEventList()` reading `observeFullEvaluationData` from `CURRENT_CONFIG` at flush time, which could be overwritten by a subsequent RC update before the 10s flush fired. Fixed in the `dd-trace-java` PR by capturing the value per-bucket at enqueue time with a privacy-preserving fold (any `false` in a bucket's lifetime sinks it to hashed).

tests/ffe/test_flag_eval_evp.py
nccatoni: The `flaky` and `bug` decorators accept arguments that let you filter based on several context elements, such as the library, library version, or weblog. See the ([documentation](https://github.com/DataDog/system-tests/blob/main/docs/edit/skip-tests.md)) for more details.

However, you should prefer using manifests wherever possible. Also, these decorators do not behave the same way as `skip_if_xfail`. The `skip_if_xfail` decorator skips disabled tests and should only be used when running the test would have a significant downside, for example, if it is very slow or could crash the scenario.

The terminology can be a little confusing, so you may also want to review our [glossary](https://github.com/DataDog/system-tests/blob/main/docs/glossary.md) for more details.

I may be missing some context, but from what I can see, you do not need decorators here and should be able to put everything in the manifests.

Please do not hesitate to reach out if you have any questions.

vjfridge: Ahhh thanks for the additional context and docs @nccatoni! This is my first PR in system-tests so those links are great context.

I tried removing the `@pytest.mark.skip_if_xfail` / `@not_yet_implemented` decorators entirely to let the manifest `missing_feature (FFL-2446)` gate do its job (and preserve xpass detection for the easy-win activation script). But CI showed running these tests against the PHP weblog crashes the `/ffe` endpoint mid-run (it starts returning HTTP 500 with empty body), which then breaks sibling tests in `test_flag_eval_metrics.py` that share the scenario. 
- See the failure at [job 90074003763](https://github.com/DataDog/system-tests/actions/runs/30294635869/job/90074003763) (`test_flag_eval_metrics` starts passing and then fails partway through once the weblog has been destabilized). Root cause is a PHP sidecar timing issue tracked in [FFL-2676](https://datadoghq.atlassian.net/browse/FFL-2676).

That's exactly the case `@scenario_crash` is documented for in docs/edit/skip-tests.md: "a failing test that brings down the whole scenario prevents other tests from running." So I've pushed a change (221e146d4) that replaces `@not_yet_implemented` with `@scenario_crash` on all EVP test classes (same force-skip behavior Leo originally added, but with the semantically-correct decorator name from utils).

The tradeoff (as @cbeauchesne noted): once the feature is implemented across SDKs, `@scenario_crash` continues to force-skip xpasses so we won't get automatic easy-win activation on this file. That's a conscious cost given how badly the PHP weblog crashes when we don't force-skip. Once `FFL-2676` is fixed we can drop `@scenario_crash` entirely.

Let me know if I'm missing something! I think this is what Leo intended with his review but I misunderstood what  `@scenario_crash` meant so thanks for helping clarify!

tests/ffe/test_flag_eval_evp.py
vjfridge: See https://github.com/DataDog/system-tests/pull/7316#discussion_r3660871622!

nccatoni (approved):

PR #7320, opened 2026-07-16
[appsec] system tests for agentic onboarding runtime detection

christophe-papazian:
APPSEC-69171

System tests for RFC-1110 (Agentic Onboarding Runtime Detection, feature #563): validate the `appsec.agentic_onboarding` boolean in `app-started` telemetry when `DD_APPSEC_AGENTIC_ONBOARDING` is set — `true` when AppSec is active at startup, `false` otherwise (flag is presence-only).

Reuses existing telemetry scenarios; the two tests are `missing_feature` until each tracer implements the signal.

### DIFF
--- a/manifests/cpp_httpd.yml
+++ b/manifests/cpp_httpd.yml
@@ -29,6 +29,7 @@ manifest:
   tests/appsec/: irrelevant (ASM is not implemented in C++)
   tests/appsec/api_security/test_endpoints.py: irrelevant (language not implementing this feature)
   tests/appsec/iast/source/test_parameter_value.py::TestParameterValue::test_source_reported: irrelevant
+  tests/appsec/test_agentic_onboarding.py::Test_AppsecAgenticOnboarding: irrelevant (ASM is not implemented in C++)
   tests/appsec/test_only_python.py::Test_ImportError: irrelevant (specific tests for python tracer)
   tests/appsec/test_traces.py::Test_AppSecEventSpanTags::test_header_collection: irrelevant (test)
   tests/appsec/waf/test_reports.py::Test_Monitoring::test_waf_monitoring_optional: irrelevant (optional tags)
--- a/manifests/cpp_httpd.yml
+++ b/manifests/cpp_httpd.yml
@@ -29,7 +29,6 @@ manifest:
   tests/appsec/: irrelevant (ASM is not implemented in C++)
   tests/appsec/api_security/test_endpoints.py: irrelevant (language not implementing this feature)
   tests/appsec/iast/source/test_parameter_value.py::TestParameterValue::test_source_reported: irrelevant
-  tests/appsec/test_agentic_onboarding.py::Test_AppsecAgenticOnboarding: irrelevant (ASM is not implemented in C++)
   tests/appsec/test_only_python.py::Test_ImportError: irrelevant (specific tests for python tracer)
   tests/appsec/test_traces.py::Test_AppSecEventSpanTags::test_header_collection: irrelevant (test)
   tests/appsec/waf/test_reports.py::Test_Monitoring::test_waf_monitoring_optional: irrelevant (optional tags)
--- a/manifests/cpp_nginx.yml
+++ b/manifests/cpp_nginx.yml
@@ -76,6 +76,7 @@ manifest:
   tests/appsec/smoke_tests/test_apm_standalone.py::Test_AppSecStandaloneAPMStandalone_Telemetry: missing_feature
   tests/appsec/smoke_tests/test_apm_standalone.py::Test_AppSecStandaloneAPMStandalone_Threats: '>=1.19.1'  # TODO: a lower version might be supported
   tests/appsec/smoke_tests/test_apm_standalone.py::Test_AppSecStandaloneAPMStandalone_UserEvents: missing_feature
+  tests/appsec/test_agentic_onboarding.py::Test_AppsecAgenticOnboarding: irrelevant (ASM is not implemented in C++)
   tests/appsec/test_alpha.py: irrelevant (ASM is not implemented in C++)
   tests/appsec/test_asm_standalone.py::Test_APISecurityStandalone: '>=1.12.0'  # Modified by easy win activation script
   tests/appsec/test_asm_standalone.py::Test_APISecurityStandalone::test_appsec_propagation_does_not_force_schema_collection: missing_feature  # Created by easy win activation script
--- a/manifests/cpp_nginx.yml
+++ b/manifests/cpp_nginx.yml
@@ -76,7 +76,7 @@ manifest:
   tests/appsec/smoke_tests/test_apm_standalone.py::Test_AppSecStandaloneAPMStandalone_Telemetry: missing_feature
   tests/appsec/smoke_tests/test_apm_standalone.py::Test_AppSecStandaloneAPMStandalone_Threats: '>=1.19.1'  # TODO: a lower version might be supported
   tests/appsec/smoke_tests/test_apm_standalone.py::Test_AppSecStandaloneAPMStandalone_UserEvents: missing_feature
-  tests/appsec/test_agentic_onboarding.py::Test_AppsecAgenticOnboarding: irrelevant (ASM is not implemented in C++)
+  tests/appsec/test_agentic_onboarding.py::Test_AppsecAgenticOnboarding: missing_feature
   tests/appsec/test_alpha.py: irrelevant (ASM is not implemented in C++)
   tests/appsec/test_asm_standalone.py::Test_APISecurityStandalone: '>=1.12.0'  # Modified by easy win activation script
   tests/appsec/test_asm_standalone.py::Test_APISecurityStandalone::test_appsec_propagation_does_not_force_schema_collection: missing_feature  # Created by easy win activation script
--- a/tests/appsec/test_agentic_onboarding.py
+++ b/tests/appsec/test_agentic_onboarding.py
@@ -38,11 +38,15 @@ class Test_AppsecAgenticOnboarding:
         """Flag set + AppSec enabled -> appsec.agentic_onboarding=true, origin=env_var."""
         entries = _find_config(CONFIG_KEY)
         assert entries, f"{CONFIG_KEY} missing from telemetry"
-        entry = entries[-1]
-        assert entry.get("origin") == "env_var", f"Expected origin env_var, got: {entry}"
-        # native bool preferred; string fallback for cross-tracer compat. `is True`
-        # rejects numeric 1 (1 == True in Python).
-        assert entry["value"] is True or entry["value"] == "true", f"Expected boolean true, got: {entry}"
+        # Assert every reported entry is consistent rather than counting them: prefork /
+        # multiprocess weblogs (uwsgi, php-fpm) and the app-client-configuration-change
+        # carrier can legitimately emit more than one, so a count check would be flaky;
+        # this still catches a contradictory or wrongly-sourced report.
+        for entry in entries:
+            assert entry.get("origin") == "env_var", f"Expected origin env_var, got: {entry}"
+            # native bool preferred; string fallback for cross-tracer compat. `is True`
+            # rejects numeric 1 (1 == True in Python).
+            assert entry["value"] is True or entry["value"] == "true", f"Expected boolean true, got: {entry}"
         # presence-only: only the derived boolean is sent, never the raw flag
         assert not _find_config(RAW_ENV_VAR), f"{RAW_ENV_VAR} must not be reported in telemetry"
 
--- a/tests/appsec/test_agentic_onboarding.py
+++ b/tests/appsec/test_agentic_onboarding.py
@@ -1,12 +1,18 @@
 from utils import rfc, features, scenarios, interfaces
 
 
-# The runtime signal reported in the app-started telemetry configuration list.
+# The runtime signal reported in the telemetry configuration list.
 CONFIG_KEY = "appsec.agentic_onboarding"
 
 
 def _find_config(name: str) -> list[dict]:
-    """Return app-started configuration entries matching `name` (case-insensitive)."""
+    """Return configuration entries matching `name` (case-insensitive).
+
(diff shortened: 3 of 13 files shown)


christophe-papazian:
@codex review

tests/appsec/test_agentic_onboarding.py
```diff
@@ -0,0 +1,46 @@
...
+# The runtime signal reported in the app-started telemetry configuration list.
+CONFIG_KEY = "appsec.agentic_onboarding"
+
+
+def _find_config(name: str) -> list[dict]:
+    """Return app-started configuration entries matching `name` (case-insensitive)."""
+    return [
+        conf
+        for conf in interfaces.library.get_telemetry_configurations()
+        if conf.get("name", "").lower() == name.lower()
```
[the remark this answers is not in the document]
christophe-papazian: Good catch on the stale wording — but accepting `app-client-configuration-change` is intentional here, not a bug. RFC-1110 (updated 2026-07-16) states the signal normally rides `app-started` but is delivered in the subsequent `app-client-configuration-change` when AppSec finishes starting after the first `app-started`, and that consumers must not assume `app-started` exclusively. So restricting the query would violate the RFC. I've reworded the helper/docstrings to reflect that both carriers are accepted by design (70986e693).

tests/appsec/test_agentic_onboarding.py
```diff
@@ -0,0 +1,46 @@
...
+    @scenarios.telemetry_app_started_products_disabled
+    def test_reported_false(self):
+        """Flag set + AppSec disabled -> appsec.agentic_onboarding=false.
+
+        The flag value in this scenario is "true" while AppSec is off, proving the
+        reported boolean follows AppSec state, not the flag's literal value.
+        """
+        entries = _find_config(CONFIG_KEY)
+        assert entries, f"{CONFIG_KEY} missing from app-started telemetry"
+        assert entries[-1]["value"] in (False, "false"), f"Expected falsy value, got: {entries[-1]}"
```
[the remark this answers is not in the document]
christophe-papazian: Agreed — applied. `test_reported_false` now also asserts `origin == "env_var"` so the `false` entry must be driven by the flag itself rather than a default/product-state configuration that happens to be false (70986e693). This also aligns with RFC-1110 item 3 (asserting the corresponding origin).

tests/appsec/test_agentic_onboarding.py
```diff
@@ -0,0 +1,57 @@
...
+    return [
+        conf
+        for conf in interfaces.library.get_telemetry_configurations()
+        if conf.get("name", "").lower() == name.lower()
+    ]
+
+
+@rfc("https://docs.google.com/document/d/1l09G9w-VQGUy1RPZ41n2uwPDxHDGZNZ9pWWAvHg4f34/edit")
+@features.appsec_agentic_onboarding
+class Test_AppsecAgenticOnboarding:
```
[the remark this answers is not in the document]
christophe-papazian: test file/scenarios are not even run for rust.

avara1986 (approved):

cbeauchesne (approved):

tests/appsec/test_agentic_onboarding.py
```diff
@@ -0,0 +1,57 @@
+from utils import rfc, features, scenarios, interfaces
+
+
+# The runtime signal reported in the telemetry configuration list.
```
e-n-0: **Cleanup:** This comment only restates `CONFIG_KEY` and does not explain a non-obvious constraint. Please remove it; the class and RFC docstrings already carry the useful context.

tests/appsec/test_agentic_onboarding.py
```diff
@@ -0,0 +1,57 @@
...
+CONFIG_KEY = "appsec.agentic_onboarding"
+
+
+def _find_config(name: str) -> list[dict]:
```
e-n-0: `list[dict]` leaves both key and value types unknown and violates the repository typing guidance. Strict LSP diagnostics report missing generic arguments here. Please use a precise telemetry-entry type, preferably a `TypedDict`, or at least `dict[str, object]`.

tests/appsec/test_agentic_onboarding.py
```diff
@@ -0,0 +1,57 @@
...
+
+    Per RFC-1110, the signal normally rides `app-started`, but when AppSec finishes
+    starting after the first `app-started` it is delivered in the subsequent
+    `app-client-configuration-change` event instead. `get_telemetry_configurations()`
+    reads both carriers, so we intentionally accept either.
+    """
+    return [
+        conf
+        for conf in interfaces.library.get_telemetry_configurations()
+        if conf.get("name", "").lower() == name.lower()
```
e-n-0: Please match `CONFIG_KEY` exactly. Lowercasing both sides lets a tracer emit the wrong wire key casing and still pass, while intake normalization and retention mappings rely on the exact configuration name.

tests/appsec/test_agentic_onboarding.py
```diff
@@ -0,0 +1,57 @@
...
+    ]
```
e-n-0: The callers select `entries[-1]`, so duplicate or contradictory reports can pass even though the RFC specifies one-shot reporting. Please preserve runtime/event context and assert one entry per runtime process; a global `len(entries) == 1` could incorrectly fail valid multiprocess weblogs.

tests/appsec/test_agentic_onboarding.py
e-n-0, line 25: Please add a negative control where `DD_APPSEC_AGENTIC_ONBOARDING` is absent from every source and assert that no entry exists. Both current tests set the flag, so an implementation that always emits this configuration can pass.

tests/appsec/test_agentic_onboarding.py
```diff
@@ -0,0 +1,57 @@
...
+@features.appsec_agentic_onboarding
+class Test_AppsecAgenticOnboarding:
+    """RFC-1110: agentic-onboarding runtime signal.
+
+    When `DD_APPSEC_AGENTIC_ONBOARDING` is set, the tracer reports a single boolean
+    configuration entry `appsec.agentic_onboarding` in its telemetry (on `app-started`,
+    or the subsequent `app-client-configuration-change`), with `origin=env_var` when
+    the flag is injected as an environment variable. The reported value reflects
+    whether AppSec is active at startup; the flag is presence-only (value never read).
+    """
```
e-n-0, line 34: The RFC says the flag value is never transmitted. These tests only locate the derived key, so an implementation that also reports the raw `DD_APPSEC_AGENTIC_ONBOARDING` configuration still passes. Please assert that the raw key is absent from telemetry.

tests/appsec/test_agentic_onboarding.py
```diff
@@ -0,0 +1,57 @@
...
+
+    @scenarios.telemetry_enhanced_config_reporting
+    def test_reported_true(self):
+        """Flag set + AppSec enabled -> appsec.agentic_onboarding=true, origin=env_var."""
+        entries = _find_config(CONFIG_KEY)
+        assert entries, f"{CONFIG_KEY} missing from telemetry"
+        entry = entries[-1]
+        assert entry.get("origin") == "env_var", f"Expected origin env_var, got: {entry}"
+        assert entry["value"] in (True, "true"), f"Expected truthy value, got: {entry}"
```
e-n-0: This assertion also accepts numeric `1` because `1 == True` in Python; the false-case equivalent accepts `0`. Please discriminate types explicitly. Ideally require native booleans to match RFC-1110; if string compatibility is intentional, ensure numeric values cannot satisfy the assertion.

tests/appsec/test_agentic_onboarding.py
```diff
@@ -0,0 +1,57 @@
...
+
+    @scenarios.telemetry_app_started_products_disabled
```
e-n-0, line 53: This proves the explicitly-disabled case, but not the RFC case where AppSec is configured on and fails during startup. Without an induced startup failure, an implementation that reports configured enablement instead of actual activation can pass.

utils/_context/_scenarios/__init__.py
```diff
@@ -270,6 +272,9 @@ class _Scenarios:
             "DD_LOGS_INJECTION": "false",
             "CONFIG_CHAINING_TEST": "true",
             "DD_TRACE_CONFIG": "/app/ConfigChaining.properties",
+            # RFC-1110: agentic-onboarding flag set while AppSec is enabled -> appsec.agentic_onboarding=true
+            # value "0" is intentional: the flag is presence-only, its value is never read
+            "DD_APPSEC_AGENTIC_ONBOARDING": "0",
```
e-n-0, line 277: Using `"0"` usefully catches one false-like value, but RFC-1110 explicitly calls out `1`, `true`, `false`, `0`, and the empty string. Empty is the important distinct edge case: an implementation using a non-empty check would pass these tests. Please add parameterized presence-only coverage.

manifests/cpp_httpd.yml
```diff
@@ -29,6 +29,7 @@ manifest:
   tests/appsec/: irrelevant (ASM is not implemented in C++)
   tests/appsec/api_security/test_endpoints.py: irrelevant (language not implementing this feature)
   tests/appsec/iast/source/test_parameter_value.py::TestParameterValue::test_source_reported: irrelevant
+  tests/appsec/test_agentic_onboarding.py::Test_AppsecAgenticOnboarding: irrelevant (ASM is not implemented in C++)
```
e-n-0: This entry is redundant: line 29 already marks the entire `tests/appsec/` directory as irrelevant. Please remove the narrower declaration so the manifest does not accumulate dead configuration.

manifests/cpp_nginx.yml
```diff
@@ -76,6 +76,7 @@ manifest:
   tests/appsec/smoke_tests/test_apm_standalone.py::Test_AppSecStandaloneAPMStandalone_Telemetry: missing_feature
   tests/appsec/smoke_tests/test_apm_standalone.py::Test_AppSecStandaloneAPMStandalone_Threats: '>=1.19.1'  # TODO: a lower version might be supported
   tests/appsec/smoke_tests/test_apm_standalone.py::Test_AppSecStandaloneAPMStandalone_UserEvents: missing_feature
+  tests/appsec/test_agentic_onboarding.py::Test_AppsecAgenticOnboarding: irrelevant (ASM is not implemented in C++)
```
e-n-0: The reason is misleading in this manifest because several AppSec tests are activated for the C++ NGINX integration. The actual reason is that RFC-1110 targets the seven application tracers and excludes this integration; please state that explicitly.

e-n-0 (approved):

tests/appsec/test_agentic_onboarding.py
christophe-papazian: Done — removed the redundant comment.

tests/appsec/test_agentic_onboarding.py
christophe-papazian: Done — the helper now returns `list[dict[str, object]]`.

tests/appsec/test_agentic_onboarding.py
christophe-papazian: Done — matching is now case-sensitive (exact `CONFIG_KEY`), so a wrong wire casing fails instead of passing.

tests/appsec/test_agentic_onboarding.py
christophe-papazian: Addressed. A count check would be flaky here — prefork/multiprocess weblogs (uwsgi, php-fpm) and the RFC-1110 `app-client-configuration-change` carrier can legitimately emit more than one entry, and even a per-runtime `len==1` can be tripped by app-started re-sends. Instead of `entries[-1]`, the tests now iterate **all** matching entries and assert each is consistent (origin=env_var + expected boolean), which catches a contradictory/duplicate report without any count-based flakiness.

tests/appsec/test_agentic_onboarding.py
christophe-papazian: Won't fix, by design: system-tests avoids always-green / easy-win tests, and a flag-absent negative control passes trivially (and can never fail before implementation). The presence-only contract is instead exercised by the value-agnostic assertions in the true/false cases.

tests/appsec/test_agentic_onboarding.py
christophe-papazian: Done — both tests now assert the raw `DD_APPSEC_AGENTIC_ONBOARDING` key is absent from telemetry, so only the derived boolean is accepted.

tests/appsec/test_agentic_onboarding.py
christophe-papazian: Done — now `value is True or value == "true"` (and `is False`/`"false"`), so numeric 1/0 no longer satisfy the assertion.

tests/appsec/test_agentic_onboarding.py
christophe-papazian: Won't fix here: reliably inducing an AppSec startup failure while `DD_APPSEC_ENABLED=true` would require a dedicated broken-init scenario, and the behavior varies per tracer. We're intentionally not adding a new scenario for this; tracking it as a follow-up rather than blocking this PR.

utils/_context/_scenarios/__init__.py
christophe-papazian: Won't fix here: a full presence-only value matrix (esp. the empty-string edge) means multiplying end-to-end scenarios, which we're deliberately avoiding — this is better suited to parametric coverage. The current `"0"` value already exercises a false-like value while AppSec is on, and the true/false assertions never read the value. Tracking the exhaustive matrix as a parametric follow-up.

manifests/cpp_httpd.yml
christophe-papazian: Done — removed; the blanket `tests/appsec/: irrelevant` already covers it.

manifests/cpp_nginx.yml
christophe-papazian: Done — replaced with a bare `missing_feature` (no misleading reason). RFC-1110 targets the 7 application tracers; the C++ integrations are out of scope.

PR #7329, opened 2026-07-16
test(ffe): frame mocked CDN responses for fetch clients

leoromanovsky:
## Motivation

Node.js built-in `fetch` rejects close-delimited HTTP response bodies when the connection terminates without an explicit framing signal. The agentless UFC mock merged in #7315 returned JSON bodies without `Content-Length`, so otherwise valid Node system tests failed while decoding the response.

## Changes

- Add `Content-Length` to mocked UFC configuration responses.
- Add `Content-Length` to the mock's JSON status and error responses.
- Verify the advertised configuration length in the fixture self-test.

## Decisions

- Keep this follow-up limited to standards-compliant HTTP response framing; the JSON:API payload, `dd_env` query contract, authentication checks, and request tracking remain unchanged.
- Validate framing in the fixture's own `TEST_THE_TEST` scenario so SDK system tests can rely on the controlled CDN surface.

## Validation

```text
uv run --python 3.12 pytest -q -S TEST_THE_TEST tests/test_the_test/test_mock_ffe_agentless_backend.py
3 passed in 2.34s
```

Required by DataDog/dd-trace-js#9397.

### DIFF
--- a/tests/test_the_test/test_mock_ffe_agentless_backend.py
+++ b/tests/test_the_test/test_mock_ffe_agentless_backend.py
@@ -34,6 +34,7 @@ def test_mock_ffe_agentless_backend_serves_fixture_and_tracks_metadata(worker_id
             timeout=5,
         )
         response.raise_for_status()
+        assert response.headers["Content-Length"] == str(len(response.content))
 
         payload = response.json()
         assert payload["data"]["type"] == UFC_RESPONSE_TYPE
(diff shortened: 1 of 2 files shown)


leoromanovsky:
Closing because the response framing fix is already present on main. Both response paths now set Content-Length in the relocated mocked backend: https://github.com/DataDog/system-tests/blob/8770fe3d914d8f7028c1d4177292acc5b5498a34/utils/mocked_backend/ffe.py#L184-L235. Leaving the branch intact.

PR #7355, opened 2026-07-21
test(ffe): enable Node.js agentless flag evaluation

leoromanovsky:
## Motivation

Node.js now supports the complete agentless configuration-source behavior. The system-test manifest still disables this contract before its 6.8.0 and 5.119.0 releases.

## Changes and Decisions

- Enable agentless configuration and flag evaluation for `>=6.8.0 || ^5.119.0`.
- Exercise the default agentless source without the legacy Remote Configuration opt-in.
- Run all configuration-source cases without per-test or weblog exclusions.
- Expose the evaluation endpoint in the TypeScript, Fastify, and Next.js weblogs.
- Call OpenFeature `setProviderAndWait()` on the first evaluation request, after weblog startup.
- Keep custom-endpoint authentication in the parametric configuration-source contract.
- Refresh changed application code and Node 18-compatible OpenFeature dependencies that existing base images otherwise hide.
- Skip the Agent-backed weblog flush when the scenario does not start an Agent.
- Persist the mocked CDN request status so replay mode can verify the agentless contract.
- Keep exposure events, evaluation events, and evaluation metrics disabled.

### DIFF
--- a/tests/parametric/capabilities.yml
+++ b/tests/parametric/capabilities.yml
@@ -72,7 +72,7 @@ capabilities:
       - ASM_ACTIVATION
       - ASM_AUTO_USER_INSTRUM_MODE
 
-    '>=5.72.0 <5.119.0 || >=6.0.0-pre <6.8.0':
+    '>=5.72.0':
       - FFE_FLAG_CONFIGURATION_RULES
 
     '>=5.83.0':
--- a/tests/parametric/capabilities.yml
+++ b/tests/parametric/capabilities.yml
@@ -72,7 +72,7 @@ capabilities:
       - ASM_ACTIVATION
       - ASM_AUTO_USER_INSTRUM_MODE
 
-    '>=5.72.0':
+    '>=5.72.0 <5.119.0 || >=6.0.0-pre <6.8.0':
       - FFE_FLAG_CONFIGURATION_RULES
 
     '>=5.83.0':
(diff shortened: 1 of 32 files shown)


leoromanovsky:
@codex

tests/parametric/capabilities.yml
```diff
@@ -72,7 +72,7 @@ capabilities:
       - ASM_ACTIVATION
       - ASM_AUTO_USER_INSTRUM_MODE
 
-    '>=5.72.0':
+    '>=5.72.0 <5.119.0 || >=6.0.0-pre <6.8.0':
```
sameerank: Codex had a thought about this

> This range removes FFE_FLAG_CONFIGURATION_RULES exactly at Node.js 5.119.0/6.8.0, but test_default_capability_completeness explicitly sets DD_FEATURE_FLAGS_CONFIGURATION_SOURCE=remote_config, and the enabled source-selection contract requires that capability in remote_config mode. Pre-releases still satisfy the `<5.119.0`/`<6.8.0` bounds, explaining why CI is green; once either stable release is tested, strict capability equality will reject the still-advertised capability. Keep this capability expected for explicit remote_config runs.

sameerank (approved):
One thing that might be worth a commit before merging

nccatoni (approved):

tests/parametric/capabilities.yml
leoromanovsky: Thank you!

leoromanovsky: Fixed in 0977bb31c300ad94ccf7c1f4db4f96a09155943b. Restored the >=5.72.0 expectation because this capability check selects Remote Config explicitly.

PR #7372, opened 2026-07-23
feat(otel): scaffold OTel consistent probability sampling (ot.th/ot.rv) conformance tests

genesor:
## Motivation

RFC for cross-tracer support of OTel consistent probability sampling (`ot.th`/`ot.rv`) tracestate members across all 9 dd-trace-* SDKs (APMAPI-2171 distributed tracing, APMAPI-2172 trace export).

## Changes

Scaffolding only — every new test class is landed **disabled** for all 9 tracers via manifest entries (`missing_feature`). No tracer implements this yet.

- `tests/test_otel_tracestate_sampling.py` (new): A1-A6 from APMAPI-2171
  - `Test_EmitOtOnProbabilityDecision_Rate0_{01,1,2,5,99}` — golden-vector rv/th derivation at 5 fixed rates, crossed with the 23 trace IDs from `tests/fixtures/sampling_rates.csv`, verified against the Go Knuth-hash sampler formula
  - `Test_ForwardInboundOtUnchanged` — inbound th/rv forwarded unchanged, never re-derived
  - `Test_PreserveDdAndOtherVendors` — ot= handling doesn't disturb dd= or unrelated vendor tracestate members
  - `Test_ForceKeepClearsTh` — a non-probability (force-keep, via ASM `/waf`) decision erases th but forwards an inherited rv
  - `Test_SampledWithoutOtNotFabricated` — an inherited sampling decision with no ot= must not fabricate th/rv
  - `Test_MalformedOtHandling` — malformed th/rv treated as absent, dd=/vendors preserved, trace never rejected
- `tests/otel/test_tracing_otlp.py`: B1/B2 from APMAPI-2172 — `Test_Otlp_Carries_Ot`, `Test_Otlp_Forwards_Inherited_Ot`, asserting on the exported OTLP span's `traceState`
- `utils/_context/_scenarios/__init__.py`: 4 new fixed-rate scenarios (`otel_sampling_rate_0_{01,1,2,99}`); rate 0.5 reuses the existing `sampling` scenario
- `manifests/*.yml` (all 9 languages): `missing_feature (APMAPI-2171)` / `(APMAPI-2172)` entries for every new class

## Workflow

    * Framework is modified, or non obvious usage of it -> get a review from R&P team

## Reviewer checklist

* [ ] Anything but `tests/` or `manifests/` is modified ? I have the approval from R&P team (note: `utils/_context/_scenarios/__init__.py` is modified to add the 4 new scenarios)
* [ ] A docker base image is modified? N/A
* [ ] A scenario is added, removed or renamed? Yes — 4 new scenarios added, needs R&P team review


tests/test_otel_tracestate_sampling.py
```diff
@@ -0,0 +1,540 @@
...
+            tracestate = _outbound_tracestate(req)
+            assert "ot" in tracestate, f"trace_id={trace_id}: no ot= tracestate member emitted on a probability decision"
+            ot = _parse_ot(tracestate)
+
+            assert ot.get("rv") == expected_rv, f"trace_id={trace_id}: rv={ot.get('rv')!r}, expected {expected_rv!r}"
+            assert ot.get("th") == expected_th, f"trace_id={trace_id}: th={ot.get('th')!r}, expected {expected_th!r}"
+
+            # the configured sample rate must be recoverable from th alone: rate = 1 - th / 2**56
+            recovered_rate = 1 - int(ot["th"], 16) / (1 << 56)
+            assert abs(recovered_rate - self.RATE) < 1e-6, (
```
vpellan: What is the point of this assertion? How can this assertion fails if the one above don't? It actually fails with 0.01 rate on my side, which is weird because we checked that we have the correct `th` on line 241, but then we seem to check that the difference between the recovered rate and self.RATE is small. However the recovered rate is something calculated in the test, not something exported by our tracers. Checking that we have the exact correct `th` is more than enough, and this assertion failing while the one above passed only means that the formula at line 244 is incorrect (probably due to floating point imprecision)

tests/test_otel_tracestate_sampling.py
```diff
@@ -0,0 +1,537 @@
...
+        self.malformed_th_request = weblog.get(
+            "/make_distant_call",
+            params={"url": "http://weblog:7777"},
+            headers={
+                "traceparent": _traceparent(self.TRACE_ID, sampled=True),
+                "tracestate": f"ot=rv:{self.MALFORMED_TH_RV};th:not-hex-either",
+            },
+        )
+
+    def test_malformed_th_only_treated_as_absent(self):
```
vpellan, line 612: Maybe we could also add a test with malformed rv only, and what we do in this situation (ensure that we fallback on the same behaviour than ot.th without rv described above)

tests/test_otel_tracestate_sampling.py
```diff
@@ -0,0 +1,537 @@
...
+                "User-Agent": "Arachni/v1",
+            },
+        )
+
+    def test_force_keep_overrides_inherited_drop_decision(self):
+        """Upstream already decided to drop (a full th/rv pair); the local force-keep still clears th but forwards the inherited rv unchanged."""
+        assert self.dropped_request.status_code == 200
+        ot = _parse_ot(_outbound_tracestate(self.dropped_request))
+
+        assert ot.get("rv") == self.DROPPED_RV, "inherited rv was not forwarded when overriding an inherited drop"
```
vpellan, line 537: I wouldn't expect `rv` to be included, as without a `th` it has no meaning anyway
```suggestion
        assert "rv" not in ot, "inherited rv was not forwarded when overriding an inherited drop"
```

tests/test_otel_tracestate_sampling.py
```diff
@@ -0,0 +1,537 @@
...
+                "User-Agent": "Arachni/v1",
+            },
+        )
+
+    def test_force_keep_forwards_inherited_rv(self):
+        """An inherited rv (no th, so no upstream probability decision) is still forwarded on a local force-keep."""
+        assert self.rv_only_request.status_code == 200
+        ot = _parse_ot(_outbound_tracestate(self.rv_only_request))
+
+        assert ot.get("rv") == self.INHERITED_RV, "inherited rv was not forwarded on a force-keep decision"
```
vpellan, line 514: I wouldn't expect `rv` to be included, as without a `th` it has no meaning anyway
```suggestion
        assert "rv" not in ot, "inherited rv was not forwarded on a force-keep decision"
```

vpellan: I'd rename the test `test_force_keep_overrides_inherited_keep_decision`

tests/test_otel_tracestate_sampling.py
```diff
@@ -0,0 +1,583 @@
...
+
+    def test_sampled_without_ot_not_fabricated(self):
+        assert self.r.status_code == 200
+        tracestate = _outbound_tracestate(self.r)
+        assert "ot" not in tracestate, "th/rv were fabricated for an inherited sampling decision with no inbound ot="
+
+
+@scenarios.otel_sampling_rate_0_1
+@features.w3c_headers_injection_and_extraction
+class Test_MalformedOtHandling:
```
vpellan, line 569: For these cases, as we generate new th and rv, should `th` be derived from the sampling rate, and `rv` from the trace id last 56 bits?

tests/test_otel_tracestate_sampling.py
```diff
@@ -0,0 +1,583 @@
...
+            },
+        )
+
+    def test_malformed_th_only_treated_as_absent(self):
+        """Th malformed, rv otherwise well-formed: th is cleared, but the well-formed rv is still forwarded."""
+        req = self.malformed_th_request
+        assert req.status_code == 200, "a malformed ot.th must not cause the trace to be rejected"
+
+        ot = _parse_ot(_outbound_tracestate(req))
+        assert ot.get("rv") == self.MALFORMED_TH_RV, "the well-formed inbound rv was not forwarded"
```
vpellan, line 618: Again, I wouldn't expect `rv` to be included, as without a `th` it has no meaning.
```suggestion
        assert "rv" not in ot, "the well-formed inbound rv was forwarded"
```

vpellan (commented):
I also believe that we should add tests for imprecision workaround while going from 64 to 56 bit (see the annex of the RFC about 64->56 bit imprecision)

tests/test_otel_tracestate_sampling.py
genesor: No, we do not do anything. We have no way of knowing how the decision was made upstream. We just to cleanup

tests/test_otel_tracestate_sampling.py
genesor: rv is intentionally forwarded even without th. Per the RFC: "DD makes a non-probability decision (force-keep, rate limiter) -> no th; forward an inherited rv if present, otherwise emit nothing." rv is randomness metadata a later hop can pair with its own th, so it's not meaningless on its own. Keeping the assertions as-is.

tests/test_otel_tracestate_sampling.py
genesor: rv is intentionally forwarded even without th. Per the RFC: "DD makes a non-probability decision (force-keep, rate limiter) -> no th; forward an inherited rv if present, otherwise emit nothing." rv is randomness metadata a later hop can pair with its own th, so it's not meaningless on its own. Keeping the assertions as-is.

tests/test_otel_tracestate_sampling.py
genesor: rv is intentionally forwarded even without th. Per the RFC: "DD makes a non-probability decision (force-keep, rate limiter) -> no th; forward an inherited rv if present, otherwise emit nothing." rv is randomness metadata a later hop can pair with its own th, so it's not meaningless on its own. Keeping the assertions as-is.

tests/test_otel_tracestate_sampling.py
genesor: This test covers inbound rv-only with no th. Without th we have no signal that a probability decision happened upstream, so this isn't a case of overriding an inherited keep decision. The suggested name fits the next test instead (test_force_keep_overrides_inherited_drop_decision, which has a full th/rv pair). Keeping the current name, happy to rename to something clearer if test_force_keep_forwards_inherited_rv still reads ambiguous, but not to the suggested name.

tests/test_otel_tracestate_sampling.py
genesor: Same policy as the malformed th case: we just cleanup, we don't derive anything. No plans to add a dedicated malformed-rv-only test for now.

vpellan:
There are a bunch of tests in`tests/parametric/test_headers_precedence.py` that needs to be fixed, because they were checking that we had exactly 1 vendor in the tracestate which is now not true anymore, and I believe we should do it on this PR as well.
Basically in all these tests we need to change
```python
assert len(tracestate_1_arr) == 1
assert tracestate_1_arr[0].startswith("dd=")
```
into
```python
assert any(item.startswith("dd=") for item in tracestate_1_arr)
```
The test are:
- `test_headers_precedence_propagationstyle_tracecontext_datadog`: Only part 1, with no headers
- `test_headers_precedence_propagationstyle_tracecontext`: Part 1, with no headers, but also part 5, with  Only Datadog headers, as the datadog headers should be ignored because `DD_TRACE_PROPAGATION_STYLE` is set to `tracecontext` only. I believe for 6 too, with an invalid tracecontext, we should generate a new trace id, which means we're the root span, and we should add `ot` to the tracestate
- `test_headers_precedence_propagationstyle_datadog_tracecontext`: Only part 1, with no headers as well

genesor:
Good catch, fixed in 52e0d3a01. Updated the 5 assertions across the 3 tests to `assert any(item.startswith("dd=") for item in tracestate_arr)`.

Part 6 already falls into the "new Datadog span context" path, so the same fix covers it.

vpellan:
Taking over this PR as Ben is on PTO

tests/parametric/test_headers_precedence.py
```diff
@@ -265,8 +265,7 @@ def test_headers_precedence_propagationstyle_tracecontext_datadog(self, test_lib
         assert int(traceparent1.trace_id[-16:], base=16) == int(headers1["x-datadog-trace-id"])
         assert int(traceparent1.parent_id, base=16) == int(headers1["x-datadog-parent-id"])
         assert "tracestate" in headers1
-        assert len(tracestate_1_arr) == 1
-        assert tracestate_1_arr[0].startswith("dd=")
+        assert any(item.startswith("dd=") for item in tracestate_1_arr)
```
MilanGarnier, line 268: nit: As far as I understand, after the change `dd` should still be emitted first and `ot` second. Why not keep 
`assert tracestate_1_arr[0].startswith("dd=")` ?

MilanGarnier (approved):

tests/parametric/test_headers_precedence.py
vpellan: That's true, this was done in an attempt to not make this test fail in case of wrong implementation as it is not tied to ot.th and ot.rv tests, and to be future proof, but it was probably not necessary

PR #7385, opened 2026-07-27
AI Guard: end-to-end tests for SDS redaction

avara1986:
## Motivation

**This is a proof of concept.** No tracer implements the AI Guard redaction contract yet, so every new test is marked `missing_feature` in all manifests. The goal is to agree on the *shape* of the cross-language assertions before libraries start implementing.

Jira: [APPSEC-68867](https://datadoghq.atlassian.net/browse/APPSEC-68867)
RFC: [SDS for AI Guard](https://datadoghq.atlassian.net/wiki/spaces/SDS/pages/6426329400/SDS+for+AI+Guard)
Python Implementation: https://github.com/DataDog/dd-trace-py/pull/19360
## What this PR does

Adds three test classes to `tests/ai_guard/test_ai_guard_sdk.py`:

- **`Test_Redaction`** — the backend returns a top-level `redaction_replacements` array of `{path, replacement}` pairs; the tracer must overwrite each path verbatim and store the redacted messages in the `ai_guard` meta struct. Assertions run against the meta struct because that is the cross-language, cross-provider surface. Covered shapes: single value, multiple messages with one finding, one message with several findings, mixed, tool result content, tool call arguments (must stay valid JSON), and system prompt.
- **`Test_NoRedaction`** — benign payloads (plain message, benign tool-call conversation) come back with no `redaction_replacements` and unmodified messages.
- **`Test_RedactionInSDKResponse`** — the SDK `evaluate()` response exposes the `redaction_replacements` contract, independent of the `sds_findings` detection metadata, which must still be present.

## Keeping fixtures in sync

Scenarios (messages, expected replacements, raw sensitive values) live in `tests/ai_guard/redaction_scenarios.json`. The matching VCR cassettes under `utils/build/docker/vcr/cassettes/aiguard/` are generated from that same file by `utils/scripts/gen_redaction_cassettes.py`, which computes the cassette filename hash from the request body — so the payloads the tests send and the responses the mock backend replays cannot drift. Regenerate both after editing a scenario.

## Notes for reviewers

- All new tests are `missing_feature` for cpp_httpd, cpp_nginx, dotnet, golang, java, nodejs, php, python, ruby.
- The `redaction_replacements` field name and the path grammar (`messages[0].content`, `messages[1].tool_calls[0].function.arguments`) are the main things worth reviewing — they're taken from the RFC and encoded in `_resolve_path`.
- Also includes a one-line `.gitignore` addition for a local tooling directory (`.omc/`).

[APPSEC-68867]: https://datadoghq.atlassian.net/browse/APPSEC-68867?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/ai_guard/test_ai_guard_sdk.py
```diff
@@ -775,3 +1326,135 @@ def _truncated_has_required_tags(self, series: list[dict]) -> None:
             tag_prefixes = {t.split(":")[0] for t in s["tags"]}
             missing = required_prefixes - tag_prefixes
             assert not missing, f"Missing required tag prefixes {missing} in {s['tags']}"
+
+
+@rfc("https://docs.google.com/document/d/1PYVAi9p8YzPSlmZDIwUuM0DZeRlyj5dNszOteRaUtH8/edit")
+@features.ai_guard
+@scenarios.ai_guard_telemetry
```
christophe-papazian: Blocking: this scenario sets `DD_AI_GUARD_MAX_CONTENT_SIZE=5`, so the tracer sends a *truncated* body — and cassettes are addressed by the hash of the exact body. None of the 4 scenarios posted here have a matching cassette once truncated (I computed them), so no `redaction_replacements` ever come back and `redacted:true` can never be emitted. Needs a dedicated scenario without the truncation knobs.

tests/ai_guard/test_ai_guard_sdk.py
```diff
@@ -775,3 +1326,135 @@ def _truncated_has_required_tags(self, series: list[dict]) -> None:
...
+class Test_AIGuardTelemetryRedacted:
+    """The ai_guard.requests telemetry metric reports whether an evaluation redacted anything.
+
+    The redacted tag sits alongside the existing action, block and error tags. It is true when at
+    least one replacement was applied and false when the response carried no replacements or
+    every entry was skipped fail-safe, which makes redaction measurable without meta struct.
+    """
+
+    def setup_telemetry_redacted(self):
```
christophe-papazian, line 1366: All `setup_*` run before any `test_*`, and `_find_telemetry_series` isn't request-filtered, so the `>= 2` count can be satisfied by traffic from `Test_AIGuardTelemetryRequests` in the same scenario. Isolate and assert exact counts.

tests/ai_guard/test_ai_guard_sdk.py
```diff
@@ -775,3 +1326,135 @@ def _truncated_has_required_tags(self, series: list[dict]) -> None:
...
+        """redaction_replacements in the response is ignored: meta struct keeps the original messages."""
+        assert self.r.status_code == 200
+        interfaces.library.validate_one_span(
+            self.r, validator=self._assert_unredacted_span(self.scenario), full_trace=True
+        )
+
+    def setup_findings_still_reported(self):
+        self.scenario, self.r = _post_redaction_scenario("REDACT_MIXED")
+
+    def test_findings_still_reported(self):
```
christophe-papazian, line 1467: The kill-switch is only asserted on meta_struct. A tracer that records originals in the span but returns redacted messages to the caller passes both tests. Worth also asserting `body["messages"] == scenario["messages"]`.

tests/ai_guard/test_ai_guard_sdk.py
```diff
@@ -775,3 +1326,135 @@ def _truncated_has_required_tags(self, series: list[dict]) -> None:
...
+            ai_guard = _assert_key(span["meta_struct"], "ai_guard")
+            messages = _assert_key(ai_guard, "messages")
+            assert messages == scenario["messages"], (
+                f"Messages must be untouched when redaction is disabled: {messages} != {scenario['messages']}"
+            )
+
+            # Absence of the tag means "redaction is off", which the RFC keeps distinct from
+            # false, meaning "redaction is on and nothing was redacted".
+            assert _span_tag(span, REDACTED_TAG) is None, (
+                f"'{REDACTED_TAG}' must not be set when redaction is disabled, got '{_span_tag(span, REDACTED_TAG)}'"
```
christophe-papazian, line 1448: Tri-state contract (tag absent = off, `false` = on-but-nothing, `true` = redacted) imposed on 9 tracers. Worth spelling out in `docs/`.

tests/ai_guard/test_ai_guard_sdk.py
```diff
@@ -494,6 +497,545 @@ def test_sds_in_response(self):
...
+        assert self.r.status_code == 403
+        interfaces.library.validate_one_span(
+            self.r, validator=self._assert_blocked_and_redacted(self.scenario), full_trace=True
+        )
+
+
+@rfc("https://docs.google.com/document/d/1PYVAi9p8YzPSlmZDIwUuM0DZeRlyj5dNszOteRaUtH8/edit")
+@features.ai_guard
+@scenarios.ai_guard
+class Test_RedactedMessagesInSDKResponse:
```
christophe-papazian, line 973: `messages` / `redaction_replacements` in the response is a new weblog contract. Python works by accident (`jsonify(evaluation)`), but the Rails controllers build an explicit field whitelist and drop both — those will fail for weblog reasons, not tracer ones. `docs/understand/weblogs/end-to-end_weblog.md` (endpoint spec) isn't updated either.

tests/ai_guard/test_ai_guard_sdk.py
```diff
@@ -494,6 +497,545 @@ def test_sds_in_response(self):
...
+        ai_guard = _assert_key(span["meta_struct"], "ai_guard")
+        messages = _assert_key(ai_guard, "messages")
+        assert messages == scenario["expected_messages"], (
+            f"Messages in meta struct do not match the expected redaction outcome: "
+            f"{messages} != {scenario['expected_messages']}"
+        )
+
+        serialized = json.dumps(messages)
+        for sensitive_value in scenario["sensitive_values"]:
+            assert sensitive_value not in serialized, (
```
christophe-papazian, line 583: `sensitive_values` is generated as "what the reference implementation removed", so anything it leaves behind is excluded from this check by construction — e.g. `paco@gmail.com` in `MIXED_APPLIED_AND_SKIPPED`. This asserts tracer == generator, not tracer == RFC.

tests/ai_guard/test_ai_guard_sdk.py
```diff
@@ -494,6 +497,545 @@ def test_sds_in_response(self):
...
+    return None
+
+
+def _assert_redacted_tag(span: DataDogLibrarySpan, *, expected: bool) -> None:
+    """The ai_guard.redacted span tag must state whether the evaluation redacted anything."""
+    raw = _span_tag(span, REDACTED_TAG)
+    assert raw is not None, f"'{REDACTED_TAG}' not set on the ai_guard span"
+    if isinstance(raw, (int, float)) and not isinstance(raw, bool):
+        # A tracer reporting the tag as a metric uses 1/0 rather than true/false.
+        actual: bool | str = bool(raw)
```
christophe-papazian: `bool(raw)` accepts `2`, `-1`, `0.5` as true, despite the comment saying 1/0. Require exactly 0 or 1.

tests/ai_guard/test_ai_guard_sdk.py
```diff
@@ -595,8 +1137,20 @@ def test_ai_guard_event(self):
...
+def _find_telemetry_series(metric: str) -> list[dict]:
+    """Extract the telemetry series of one AI Guard metric, under either accepted identity.
+
+    AI Guard metrics currently live in the appsec namespace as `ai_guard.<metric>`, and the
+    telemetry design moves them to a dedicated `ai_guard` namespace where they are named
+    `<metric>` (see the AI Guard telemetry metrics doc). The metric is unambiguous either way, so
+    both are accepted: this keeps the tests passing across the migration rather than pinning them
+    to a namespace no tracer emits yet.
+    """
+    accepted = {("appsec", f"ai_guard.{metric}"), (TELEMETRY_NAMESPACE, metric)}
```
christophe-papazian: Accepting two namespaces loosens two existing telemetry tests and is unrelated to redaction (same for the manifest flip `missing_feature` -> `bug` and the new heartbeat env vars). Worth splitting into its own PR.

utils/scripts/gen_redaction_cassettes.py
```diff
@@ -0,0 +1,733 @@
...
+        write_cassette(request_body, response_body)
+        cassettes[cassette] = name
+        sidecar[name] = entry
+        print(f"{name:32s} {'redacted' if entry['redacted'] else '        '} -> {cassette}")
+
+    # A renamed or reworded scenario leaves its old cassette behind, which then never matches
+    # a request again. Drop anything in the directory that no test claims.
+    for existing in sorted(CASSETTES_DIR.glob("*.json")):
+        if existing.name not in cassettes and existing.name not in FOREIGN_CASSETTES:
+            existing.unlink()
```
christophe-papazian: Blocking: this deletes every cassette in the shared dir that isn't in `SCENARIOS` or the hand-maintained `FOREIGN_CASSETTES`. The next AI Guard test that adds a cassette gets silently removed by anyone running the documented regeneration command. Safer: delete only names owned by the previous `redaction_scenarios.json`.

utils/scripts/gen_redaction_cassettes.py
```diff
@@ -0,0 +1,733 @@
...
+    removed = [value for _, sensitive in targets for value, _ in sensitive if value not in serialized]
+
+    request_body = _wrap_request_body(messages)
+    response_body = _response_body(
+        action=action,
+        reason=reason,
+        redaction_replacements=replacements,
+        sds_findings=sds_findings,
+        tags=scenario.get("tags", []),
+        tag_probs=dict(_BENIGN_TAG_PROBS),
```
christophe-papazian: All scenarios get benign probs, so `REDACT_ON_DENY` returns tag `data-exfiltration` with probability `0.0` — `Test_Tag_Probabilities` requires every returned tag to have a positive probability. Suggest asserting tags ⊆ tag_probs with prob > 0 in the generator.

utils/scripts/gen_redaction_cassettes.py
```diff
@@ -0,0 +1,733 @@
...
+}
+
+
+# ---------------------------------------------------------------------------
+# Reference implementation of the RFC redaction algorithm
+# ---------------------------------------------------------------------------
+
+# One segment of a location path: a field name plus an optional non-negative list index.
+# The whole segment must match, a partial match is rejected.
+SEGMENT_RE = re.compile(r"^(?P<name>[A-Za-z0-9_]+)(?:\[(?P<index>[0-9]+)\])?$")
```
christophe-papazian: `$` also matches before a trailing newline; the test-side resolver uses `\Z`. Use `fullmatch` here for symmetry.

docs/understand/scenarios/ai_guard.md
```diff
@@ -112,6 +112,41 @@ The scenario sets the following environment variables on the weblog:
...
+python3 utils/scripts/gen_redaction_cassettes.py
+```
+
+Cassettes are addressed by a hash of the request body, so every scenario must send a distinct
+message list. The script fails rather than overwriting a cassette owned by another AI Guard test,
+and deletes cassettes no test claims any more.
+
+The blocked path is covered too: `Test_RedactionOnBlock` sends a redacting scenario with
+`X-AI-Guard-Block: true`, so the SDK aborts instead of returning an evaluation. The weblog then
+answers 403 with the messages carried by the abort error, which must be the redacted ones.
```
christophe-papazian: Contradicts the `Test_RedactionOnBlock` docstring and the Flask abort handler: the 403 body carries no messages.

docs/understand/scenarios/ai_guard.md
```diff
@@ -112,6 +112,41 @@ The scenario sets the following environment variables on the weblog:
...
+`tests/ai_guard/redaction_scenarios.json` describing the expected outcome (the messages after
+redaction, the sensitive values that must be gone, and whether the evaluation must report itself
+as redacted). The expected outcome is computed by a reference implementation of the RFC redaction
+algorithm inside the script, cross-checked against the intent each scenario declares.
+
+To add or change a redaction scenario, edit `SCENARIOS` in that script and re-run it from the
+repository root:
+
+```bash
+python3 utils/scripts/gen_redaction_cassettes.py
```
christophe-papazian, line 152: Nothing in CI verifies the committed fixtures still match a regeneration — a drift check in `format.sh` would make this guarantee real.

christophe-papazian (commented):
Reviewed the fixture machinery first: regeneration is deterministic (no drift), and I re-derived all 31 cassette filenames from their request bodies, including the 6 previously recorded ones — the hash formula is correct. mypy/ruff clean, all 9 manifests complete, CI wiring matches `AI_GUARD_TELEMETRY`.

The issues below are mostly about the test oracles. Blocking ones IMO: the truncation/cassette mismatch on `Test_AIGuardTelemetryRedacted`, and the cassette deletion in the generator.

(Reviewed with Claude Code + codex.)

cbeauchesne (approved):
From framework usage, AGTM

tests/ai_guard/test_ai_guard_sdk.py
avara1986: Split the three parts of this up, and pushed two of them out of the PR in b18d80b:

- **Two-namespace `_find_telemetry_series`** — reverted. The helper takes `(namespace, metric)` again and the tests target only the `ai_guard` namespace, exactly as on `main`. Agreed it doesn't belong here.
- **Heartbeat env vars** — removed, and they turned out to be redundant anyway: `WeblogContainer` already sets both `DD_TELEMETRY_HEARTBEAT_INTERVAL` and `DD_TELEMETRY_METRICS_INTERVAL_SECONDS` from `telemetry_heartbeat_interval` (= 2) for every scenario, and `weblog_env` merely overrode 2s with 1s. No AI Guard scenario sets either knob now.
- **The manifest declaration** — this one I'd like to keep, because `missing_feature` is the wrong classification. These tests don't fail because the tracer hasn't implemented the metric; they fail because nothing arrives on the telemetry agentless side at all. In the AI_GUARD* scenarios the dd-trace-py telemetry writer sends no payload to the agent — every series that shows up is tagged `src_library:libdatadog` (native telemetry) and there's no app-started at all, so no `ai_guard.*` metric can be observed regardless of what the SDK emits. That's a bug, and I opened APPSEC-69627 to track it. Declaring it `missing_feature` would hide a real defect behind "not implemented yet".

Happy to move the manifest line to the follow-up PR too if you'd rather have all four classes change declaration in one place.

tests/ai_guard/test_ai_guard_sdk.py
avara1986: Good catch on both halves.

**Endpoint spec** — updated in b18d80b. `docs/understand/weblogs/end-to-end_weblog.md` now documents the full evaluate response contract (`sds`, `messages`, `redaction_replacements`), with an explicit note that a weblog building the JSON field by field has to add them, and that the 403 deliberately carries none of them.

**Rails controllers** — leaving the weblog change to the Ruby integration work rather than doing it here. The tracer-side contract is what this PR is trying to pin down, and patching another team's weblog for an SDK surface that doesn't exist yet would only add reviewers without making anything pass.

Recording the specifics so it isn't a surprise later. Three classes read those response fields: `Test_RedactedMessagesInSDKResponse`, `Test_RedactionInSDKResponse`, and `Test_RedactionDisabled` (that last one reads `body["messages"]` as of the fix for your kill-switch comment). All three are `missing_feature` in `manifests/ruby.yml`, so nothing fails today. When Ruby implements redaction and those entries flip, `AiGuardController#evaluate` needs the two fields added to its `response_data` whitelist — otherwise the failure will look like a tracer bug when it's the controller dropping the fields, which is precisely what you were flagging.

docs/understand/scenarios/ai_guard.md
avara1986: Fixed

tests/ai_guard/test_ai_guard_sdk.py
avara1986: Fixed

tests/ai_guard/test_ai_guard_sdk.py
avara1986: Fixed

tests/ai_guard/test_ai_guard_sdk.py
avara1986: Fixed

tests/ai_guard/test_ai_guard_sdk.py
avara1986: Fixed

PR #7391, opened 2026-07-27
test(rust): enable passing ot.th/ot.rv tracestate sampling scenarios

MilanGarnier:
### What does this PR do?

Flips all `tests/test_otel_tracestate_sampling.py` Rust entries from `missing_feature` to `>=0.5.0-dev` in `manifests/rust.yml` except the ones involving force keep which is not honored by the tracer currently.


paullegranddc (approved):
🥳

PR #7407, opened 2026-07-29
feat: add /trace/manual_keep_drop endpoint to all weblogs

genesor:
## Motivation

Nothing in the end-to-end suite can force a trace's sampling decision. The manual keep/drop API was only reachable from the parametric apps, so the "a manual decision overrides everything else" contract had no end-to-end coverage, and no weblog exposed an entry point for it.

### New weblog endpoint

`GET /trace/manual_keep_drop?decision=keep|drop` applies the tracer's manual keep/drop API to the current trace and returns `200`. A missing or invalid `decision` returns `400`. Documented in `docs/understand/weblogs/end-to-end_weblog.md`.

Implemented in every end-to-end weblog, one commit per variant, using each library's own manual keep/drop API:

| library | mechanism | weblogs |
|---|---|---|
| java | `GlobalTracer.get().activeSpan()` + `DDTags.MANUAL_KEEP` / `MANUAL_DROP` | spring-boot, spring-boot-3-native, vertx3, vertx4, akka-http, ratpack, play, jersey-grizzly2, resteasy-netty3 |
| python | `tracer.current_span().set_tag(MANUAL_KEEP_KEY / MANUAL_DROP_KEY)` | flask, django, fastapi, tornado |
| nodejs | `tracer.scope().active().setTag(MANUAL_KEEP / MANUAL_DROP, true)`, constants from `ext.tags` | express4, express5, express4-typescript, fastify, nextjs |
| dotnet | `Tracer.Instance.ActiveScope?.Span.SetTag(Tags.ManualKeep / ManualDrop, "true")` | weblog, via `Endpoints/ManualKeepDropEndpoint.cs` |
| ruby | `Datadog::Tracing.active_trace.keep!` / `.reject!` | rails42, rails52, rails61, rails72, rails80, sinatra14, sinatra22, sinatra32, sinatra41, rack |
| golang | shared `common.ManualKeepDrop` handler + `ext.ManualKeep` / `ext.ManualDrop` | net-http, gin, echo, chi, net-http-orchestrion, gqlgen, graphql-go, graph-gophers |
| php | `\DDTrace\set_priority_sampling(...)` | plain, symfony7x, laravel11x |

cpp and rust are skipped: they have no end-to-end weblog, only parametric.

Notes on three of those choices:

* **golang**: the handler lives once in `app/_shared/common/common.go` as an `http.HandlerFunc`, matching how `rasp.LFI`, `common.Requestdownstream` and `dbm.StubDbmHandler` are already shared. gin and echo reuse the existing `ginHandleFunc` / `echoHandleFunc` adapters.
* **php**: `set_priority_sampling` is what `\DDTrace\Span::setTag` itself calls for the `manual.keep` / `manual.drop` tags. Writing the tag into span meta was not viable: in symfony and laravel `\DDTrace\active_span()` is the `symfony.controller` / `laravel.action` child span rather than the root, and the meta path is skipped entirely when a keep decision was propagated from upstream.
* **nodejs**: express4 and express5 share `utils/build/docker/nodejs/express/app.js`, so a single commit covers both. nextjs uses `global._ddtrace` and the literal tag names, matching its neighbouring routes, which avoid direct `dd-trace` imports because of webpack bundling.

### Parametric fix

`utils/build/docker/nodejs/parametric/server.js` destructured `MANUAL_KEEP` / `MANUAL_DROP` off the top level of `dd-trace/ext`, where they do not exist (they live under `ext.tags`). It was therefore calling `span.setTag(undefined, true)`, so the existing nodejs parametric manual-keep coverage was asserting nothing. 

### Tests

`tests/test_sampling_manual.py::Test_Manual_Sampling`, SAMPLING scenario, `@features.ensure_that_sampling_is_consistent_across_languages`. Both directions are driven by an upstream distributed trace context, so each case asserts that the manual decision wins over a propagated one:

| test | upstream `x-datadog-sampling-priority` | `decision` | expected `_sampling_priority_v1` |
|---|---|---|---|
| `test_manual_keep_overrides_upstream_drop` | `0` (auto reject) | `keep` | `2` (user keep) |
| `test_manual_drop_overrides_upstream_keep` | `2` (user keep) | `drop` | `-1` (user reject) |

The SAMPLING scenario is used because it sets `DD_TRACE_STATS_COMPUTATION_ENABLED=false`, so dropped P-1 traces still reach the agent and the drop case is observable. A manual decision takes precedence over that scenario's 0.5 sample rate, so the rate does not affect either assertion.

## Impact

The drop direction is new ground: the parametric suite only ever covered manual keep overriding an upstream drop. Asserting the reverse surfaced two library gaps, both declared in the manifests so they run as xfail and turn into an xpass when fixed.

| library | declaration | reason |
|---|---|---|
| java | `bug (APMAPI-2194)` on `test_manual_drop_overrides_upstream_keep` | [APMAPI-2194](https://datadoghq.atlassian.net/browse/APMAPI-2194): `manual.keep` goes through `DDSpanContext.forceKeep()`, an unconditional update that deliberately wins over an already propagated decision, while `manual.drop` goes through `setSamplingPriority()`, a compare-and-set from `UNSET` that refuses when the priority is locked. Keep can override, drop cannot. |
| nodejs | `bug (APMAPI-1720)` on the whole class | [APMAPI-1720](https://datadoghq.atlassian.net/browse/APMAPI-1720): `PrioritySampler.sample()` returns early whenever a priority already exists, before it reads the manual tags, so neither direction is honoured. |

APMAPI-2194 is filed separately rather than folded into APMAPI-1720 because 1720's analysis only covers the keep direction and lists java as working.

dotnet, golang, php, python and ruby honour both directions.

Weblogs with no SDK to call also get declarations, since they run the SAMPLING scenario:

| manifest | declaration |
|---|---|
| `c.yml` (perl-mojolicious) | `missing_feature`, dd-trace-c does not implement manual keep/drop |
| `cpp_httpd.yml` | `missing_feature`, endpoint not implemented |
| `cpp_kong.yml`, `cpp_nginx.yml` | `irrelevant`, proxy rather than an application |
| `java_otel.yml`, `nodejs_otel.yml`, `python_otel.yml` | `irrelevant`, the OpenTelemetry test library has no Datadog manual keep/drop |

## Workflow

## Reviewer checklist

[APMAPI-2194]: https://datadoghq.atlassian.net/browse/APMAPI-2194?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/understand/weblogs/end-to-end_weblog.md
+++ b/docs/understand/weblogs/end-to-end_weblog.md
@@ -575,8 +575,21 @@ Examples:
 - `GET /trace/manual_keep_drop?decision=keep`
 - `GET /trace/manual_keep_drop?decision=drop`
 
-The endpoint returns a `200` status code with a plain text body once the decision has been applied.
-Any other value for `decision`, or a missing `decision` parameter, returns a `400` status code.
+Once the decision has been applied, the endpoint makes an outgoing HTTP call to `http://localhost:7777/`,
+so that tests can assert on the sampling decision propagated downstream. Unlike `/make_distant_call`
+the target is not configurable. The response mirrors `/make_distant_call`, with a `200` status code:
+
+```js
+{
+    "url": "http://localhost:7777/",  // the hardcoded downstream url
+    "status_code": 200, // status code of the downstream response
+    "request_headers": {}, // headers sent downstream, including the propagated sampling decision
+    "response_headers": {} // headers of the downstream response
+}
+```
+
+Any other value for `decision`, or a missing `decision` parameter, returns a `400` status code and
+makes no downstream call.
 
 ### GET /dbm
 
--- a/docs/understand/weblogs/end-to-end_weblog.md
+++ b/docs/understand/weblogs/end-to-end_weblog.md
@@ -559,6 +559,25 @@ This endpoint accept a mandatory parameter `url`. It'll make a call to these url
 }
 ```
 
+### GET /trace/manual_keep_drop
+
+This endpoint forces the sampling decision of the current trace, using the tracer's manual keep/drop
+API (the `manual.keep` / `manual.drop` tags, or the equivalent idiomatic call in the library) applied
+to the currently active span.
+
+It accepts a mandatory query parameter `decision`, whose value must be either:
+
+- `keep`: force the trace to be kept (`manual.keep`)
+- `drop`: force the trace to be dropped (`manual.drop`)
+
+Examples:
+
+- `GET /trace/manual_keep_drop?decision=keep`
+- `GET /trace/manual_keep_drop?decision=drop`
+
+The endpoint returns a `200` status code with a plain text body once the decision has been applied.
+Any other value for `decision`, or a missing `decision` parameter, returns a `400` status code.
+
 ### GET /dbm
 
 This endpoint executes database queries for [DBM supported libraries](https://docs.datadoghq.com/database_monitoring/guide/connect_dbm_and_apm/?tab=go#before-you-begin). A 200 response is returned if the query
--- a/utils/build/docker/dotnet/weblog/Endpoints/ManualKeepDropEndpoint.cs
+++ b/utils/build/docker/dotnet/weblog/Endpoints/ManualKeepDropEndpoint.cs
@@ -1,11 +1,28 @@
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Http;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.Json.Serialization;
 using Datadog.Trace;
 
 namespace weblog
 {
     public class ManualKeepDropEndpoint : ISystemTestEndpoint
     {
+        private const string DownstreamUrl = "http://localhost:7777/";
+
+        private class EndpointResponse
+        {
+            [JsonPropertyName("url")]
+            public string? Url { get; set; }
+            [JsonPropertyName("status_code")]
+            public int StatusCode { get; set; }
+            [JsonPropertyName("request_headers")]
+            public Dictionary<string, string>? RequestHeaders { get; set; }
+            [JsonPropertyName("response_headers")]
+            public Dictionary<string, string>? ResponseHeaders { get; set; }
+        }
+
         public void Register(Microsoft.AspNetCore.Routing.IEndpointRouteBuilder routeBuilder)
         {
             routeBuilder.MapGet("/trace/manual_keep_drop", async context =>
@@ -21,7 +38,17 @@ namespace weblog
                 var span = Tracer.Instance.ActiveScope?.Span;
                 span?.SetTag(decision == "keep" ? Tags.ManualKeep : Tags.ManualDrop, "true");
 
-                await context.Response.WriteAsync("OK\\n");
+                // Call downstream so that tests can assert on the sampling decision that gets propagated
+                var response = await HttpClientWrapper.LocalGetRequest(DownstreamUrl);
+                var endpointResponse = new EndpointResponse()
+                {
+                    Url = DownstreamUrl,
+                    StatusCode = (int)response.StatusCode,
+                    RequestHeaders = response.RequestMessage?.Headers.Select(kvp => new KeyValuePair<string, string>(kvp.Key, kvp.Value.First())).ToDictionary(),
+                    ResponseHeaders = response.Headers.Select(kvp => new KeyValuePair<string, string>(kvp.Key, kvp.Value.First())).ToDictionary(),
+                };
+
+                await context.Response.WriteAsJsonAsync(endpointResponse);
             });
         }
     }
--- a/utils/build/docker/dotnet/weblog/Endpoints/ManualKeepDropEndpoint.cs
+++ b/utils/build/docker/dotnet/weblog/Endpoints/ManualKeepDropEndpoint.cs
@@ -1,0 +1,28 @@
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Http;
+using Datadog.Trace;
+
+namespace weblog
+{
+    public class ManualKeepDropEndpoint : ISystemTestEndpoint
+    {
+        public void Register(Microsoft.AspNetCore.Routing.IEndpointRouteBuilder routeBuilder)
+        {
+            routeBuilder.MapGet("/trace/manual_keep_drop", async context =>
+            {
+                string? decision = context.Request.Query["decision"];
+                if (decision != "keep" && decision != "drop")
+                {
+                    context.Response.StatusCode = 400;
+                    await context.Response.WriteAsync("decision must be keep or drop\\n");
+                    return;
+                }
+
+                var span = Tracer.Instance.ActiveScope?.Span;
+                span?.SetTag(decision == "keep" ? Tags.ManualKeep : Tags.ManualDrop, "true");
+
+                await context.Response.WriteAsync("OK\\n");
+            });
+        }
+    }
+}
--- a/utils/build/docker/golang/app/_shared/common/common.go
+++ b/utils/build/docker/golang/app/_shared/common/common.go
@@ -9,6 +9,7 @@ import (
 	"net/url"
 	"os"
 
+	"github.com/DataDog/dd-trace-go/v2/ddtrace/ext"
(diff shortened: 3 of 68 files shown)


zacharycmontoya (approved):
LGTM for .NET

utils/build/docker/nodejs/express/app.js
```diff
@@ -214,6 +215,18 @@ app.get('/status', (req, res) => {
   res.status(parseInt(req.query.code) || 400).send('OK')
 })
 
+app.get('/trace/manual_keep_drop', (req, res) => {
```
bm1549, line 218: [P1] Please add the `build-nodejs-base-images` label before merging. These Node.js Dockerfiles copy the app into the shared `nodejs/base-v3` image, and this PR currently has no label that rebuilds it. The normal build will reuse the old Express, Express 4 TypeScript, Fastify, and Next.js apps, so this endpoint returns 404 in those weblogs.

utils/build/docker/dotnet/weblog/Endpoints/ManualKeepDropEndpoint.cs
```diff
@@ -0,0 +1,28 @@
...
+            {
+                string? decision = context.Request.Query["decision"];
+                if (decision != "keep" && decision != "drop")
+                {
+                    context.Response.StatusCode = 400;
+                    await context.Response.WriteAsync("decision must be keep or drop\\n");
+                    return;
+                }
+
+                var span = Tracer.Instance.ActiveScope?.Span;
```
bm1549, line 38: [P1] Please gate `tests/test_sampling_manual.py::Test_Manual_Sampling` to `v3.0.0` in `manifests/dotnet.yml`. With supported pre-3.0 tracer versions, the weblog compatibility path can leave `Tracer.Instance.ActiveScope` null for ASP.NET Core requests. This call then does nothing while returning 200, and the tests fail because the upstream decision was not overridden.

utils/build/docker/golang/app/_shared/common/common.go
```diff
@@ -28,6 +29,30 @@ func InitDatadog() {
...
+	span, ok := tracer.SpanFromContext(r.Context())
+	if !ok {
+		w.WriteHeader(http.StatusInternalServerError)
+		w.Write([]byte("no active span"))
+		return
+	}
+
+	switch r.URL.Query().Get("decision") {
+	case "keep":
+		span.SetTag(ext.ManualKeep, true)
```
bm1549, line 46: [P1] Please gate `tests/test_sampling_manual.py::Test_Manual_Sampling` to `v2.4.0` in `manifests/golang.yml`. Through v2.3.1, extracted sampling priorities are locked and these manual tags use the non-forcing setter, so they cannot override the upstream decision. v2.4.0 includes the fix that force-sets manual sampling priorities; without the gate, both tests fail on older supported versions.

utils/build/docker/php/weblogs/plain/trace_manual_keep_drop.php
```diff
@@ -0,0 +1,18 @@
...
+if ($decision !== 'keep' && $decision !== 'drop') {
+    http_response_code(400);
+    header('Content-Type: text/plain');
+    echo 'decision must be keep or drop';
+    exit;
+}
+
+// set_priority_sampling is what \DDTrace\Span::setTag does for the manual.keep / manual.drop tags,
+// and unlike writing the tag into the root span's meta it also overrides a decision propagated upstream.
+\DDTrace\set_priority_sampling(
```
bm1549, line 13: [P2] Please gate `tests/test_sampling_manual.py::Test_Manual_Sampling` to `v0.68.0` in `manifests/php.yml`. `DDTrace\set_priority_sampling` and the `DD_TRACE_PRIORITY_SAMPLING_USER_*` constants were introduced in 0.68.0. Older supported tracer versions hit an undefined function or constant here and return 500 instead of 200.

utils/build/docker/ruby/rack/config.ru
```diff
@@ -152,6 +152,24 @@ module Status
...
+
+  def run(request)
+    decision = request.params['decision']
+
+    unless %w[keep drop].include?(decision)
+      return [400, { 'Content-Type' => 'text/plain' }, ['decision must be keep or drop']]
+    end
+
+    trace = Datadog::Tracing.active_trace
+    decision == 'keep' ? trace.keep! : trace.reject!
```
bm1549, line 167: [P2] Please gate `tests/test_sampling_manual.py::Test_Manual_Sampling` to `v1.0.0` in `manifests/ruby.yml`. `Datadog::Tracing.active_trace` and trace-level `keep!`/`reject!` are not available in pre-1.0 tracer versions, which system-tests still supports. Those runs return 500 here; pre-1.0 would need the older active-span manual-tag API if the test is meant to run there.

bm1549 (commented):
### 💡 Codex Review

Here are some automated review suggestions for this pull request.

bm1549 (approved):
Overall Codex has some suggestions about manifest versions and a PR label, which I agree with

Outside of that, this is all good to go

mabdinur (approved):

MilanGarnier (approved):
At time of reviewing the endpoints don't match the doc (distant call missing) but I it's WIP.

Can you also add the endpoint for the rust axum weblog which was recently merged ?

docs/understand/weblogs/end-to-end_weblog.md
```diff
@@ -559,6 +559,38 @@ This endpoint accept a mandatory parameter `url`. It'll make a call to these url
...
+
+- `keep`: force the trace to be kept (`manual.keep`)
+- `drop`: force the trace to be dropped (`manual.drop`)
+
+Examples:
+
+- `GET /trace/manual_keep_drop?decision=keep`
+- `GET /trace/manual_keep_drop?decision=drop`
+
+Once the decision has been applied, the endpoint makes an outgoing HTTP call to `http://localhost:7777/`,
```
anna-git, line 578: Is this still true? I dont see the `manual_keep_drop` making outgoing http calls? they seem to just return 200

utils/build/docker/python/flask/app.py
```diff
@@ -609,6 +611,18 @@ def stats_unique():
...
 
 
+@app.route("/trace/manual_keep_drop")
+def trace_manual_keep_drop():
+    decision = flask_request.args.get("decision")
+    if decision not in ("keep", "drop"):
+        return Response("decision must be keep or drop", status=400)
+
+    span = tracer.current_span()
+    span.set_tag(MANUAL_KEEP_KEY if decision == "keep" else MANUAL_DROP_KEY)
```
anna-git, line 621:
```suggestion
    span.set_tag(MANUAL_KEEP_KEY if decision == "keep" else MANUAL_DROP_KEY, True)
```
maybe we should set a value? in case some asserts test a value and to keep it consistent with other samples
(valid for all python frameworks)

docs/understand/weblogs/end-to-end_weblog.md
genesor: It's ongoing, agent is running 😅 We discussed this improvement this morning and it shipped the documentation first

anna-git: or maybe just java does?

anna-git: oh ok! thanks

utils/build/docker/python/flask/app.py
genesor: `set_tag` short-circuits on these two keys ([span.py#L206-L211](https://github.com/DataDog/dd-trace-py/blob/main/ddtrace/_trace/span.py)): it calls `_override_sampling_decision` and returns before storing anything, so `manual.keep`/`manual.drop` never appear in the span's meta and no assert can read a value off them.

Kept the single-arg form to match the python parametric app (`utils/build/docker/python/parametric/apm_test_client/server.py:353`), which is the existing convention in this repo. Happy to add `True` if you'd rather be explicit, it's a no-op either way.

anna-git: yep it was more consistency / convention-wise but certainly a nit

vpellan (approved):
LGTM for shared and ruby part

anna-git (approved):

nccatoni (approved):

PR #7427, opened 2026-07-31
docker fixtures: unique container names, ownership-scoped cleanup, unconditional removal

bm1549:
## Motivation

> [View the color-coded bug and fix report](https://chonk.us1.prod.dog/v2/get/chonk-uploads-prod/brian.marks/073ee4cf-084e-4ea5-b719-1ce399d8201b) (AppGate required)

In CI Visibility over the last 30 days, this bug caused **5 of 145 final-status parametric
failures (3.45%)** on default branches (`main` and `master`).

Two parametric tests failed in a dd-trace-php run: one in setup with
`409 Conflict. The container name "/php-test-client-<id>" is already in use`, the other in
teardown on that same container. Neither is a sampling bug; any two tests in the scenario can
hit it. Three defects combine, and the third is what actually killed the container:

1. `test_id` was `str(uuid.uuid4())[0:6]`. That 24-bit value is the only variable part of every
   container and network name built from it, and those names are shared by all xdist workers
   against one docker daemon. At ~800 container-creating tests per session that is a ~1.8%
   collision chance per session. Two node ids drew the same value 2.1s apart, on gw5 and gw3.

2. `tests/integration_frameworks/conftest.py` carried a byte-identical copy of the same fixture,
   so the scenario had the same exposure.

3. On a failed create, `docker_run` force-removed every container matching the name without
   checking that it had created them. When the name was already taken, that destroyed the owning
   worker's *live* container. The losing test's cleanup removed the winner's container while the
   winner was inside `stop(timeout=5)`, so the winner's `logs()` then returned
   `409 ... dead or marked for removal`.

Separately, `docker_run`'s teardown gated `container.remove(force=True)` on `container.logs()`,
so any log-capture error skipped removal and leaked the container. Its name stayed taken for the
rest of the session, and the log file needed to explain the death was discarded.

## Changes

- `new_test_id()` is now exported by `utils.docker_fixtures` and used by both the parametric and
  integration-frameworks fixtures, so the two cannot drift apart again. It uses eight random
  bytes. The 16 hex characters keep the longest generated name
  (`{library}-test-library-{framework}-{framework_version}-{test_id}`) inside the 63-character DNS
  label limit.
- Containers are created with a per-invocation label, and failed-create cleanup selects on that
  label instead of the name, so another worker's container is never removed. Cleanup attempts all
  matching containers and reports any cleanup failures without losing the original create error.
- `container.remove(force=True)` moved into its own `finally`. A body, stop, or log-capture error
  remains the primary exception if removal also fails; the removal error is attached as a note.
  A removal-only error still propagates, while `NotFound` is suppressed when the container is
  already gone.
- The test helper now restores its patched Docker client after every invocation.
- `tests/test_the_test/test_docker_run_cleanup.py` has 15 cases covering ID entropy, ownership,
  create-failure cleanup, teardown precedence, removal paths, log capture, and helper isolation.

Verified: `ruff format --check`, `ruff check`, and `mypy` pass; full `TEST_THE_TEST` has 379 passed
and 1 xfailed.

### Note for reviewers

This changes `docker_run`'s create-failure and teardown-failure contracts, which every tracer's
parametric run depends on. Cleanup now preserves the original operation error and reports any
secondary cleanup error. A removal-only failure still fails the test.

## Workflow

## Reviewer checklist

    * `utils/docker_fixtures/_core.py` and `utils/docker_fixtures/__init__.py` are modified - R&P approval still needed, not yet obtained.
    * No base image changes.
    * No scenario changes.

### DIFF
--- a/tests/integration_frameworks/conftest.py
+++ b/tests/integration_frameworks/conftest.py
@@ -5,8 +5,8 @@ import pytest
 from utils.docker_fixtures import (
     FrameworkTestClientApi,
     TestAgentAPI,
+    new_test_id,
 )
-from utils.docker_fixtures._core import new_test_id
 from utils import context, scenarios, logger
 
 
--- a/tests/integration_frameworks/conftest.py
+++ b/tests/integration_frameworks/conftest.py
@@ -1,5 +1,4 @@
 from collections.abc import Generator
-import uuid
 
 import pytest
 
@@ -7,6 +6,7 @@ from utils.docker_fixtures import (
     FrameworkTestClientApi,
     TestAgentAPI,
 )
+from utils.docker_fixtures._core import new_test_id
 from utils import context, scenarios, logger
 
 
@@ -24,7 +24,7 @@ def pytest_collection_modifyitems(config: pytest.Config, items: list[pytest.Item
 
 @pytest.fixture
 def test_id(request: pytest.FixtureRequest) -> str:
-    result = str(uuid.uuid4())[0:6]
+    result = new_test_id()
     logger.info(f"Test {request.node.nodeid} ID: {result}")
     return result
 
--- a/tests/parametric/conftest.py
+++ b/tests/parametric/conftest.py
@@ -9,8 +9,7 @@ import pytest
 import yaml
 
 from utils import scenarios, logger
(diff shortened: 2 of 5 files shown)


cbeauchesne (approved):
So nice !

PR #7455, opened 2026-08-04
test(otel): OTel HTTP semantics coverage against the cross-tracer RFC

link04:
### Summary

Adds tests for HTTP OpenTelemetry semantic conventions enabled by `DD_TRACE_OTEL_SEMANTICS_ENABLED` and exported over OTLP.

Coverage includes HTTP server and client attributes, span naming and status (and custom error-status configuration), and trace metrics. 

### Scenarios

| Scenario | Test class | Tests | Coverage |
| --- | --- | ---: | --- |
| `OTEL_SEMANTICS_OTLP` | `Test_OtelSemantics_Spans_Http_Server` | 19 | Server attributes, naming, status semantics, URL handling, addressing, span kind, and endpoint retention |
| `OTEL_SEMANTICS_OTLP` | `Test_OtelSemantics_Spans_Http_Client` | 11 | Client attributes, naming, status semantics, peer-service suppression, span kind, and URL redaction |
| `OTEL_SEMANTICS_OTLP_CUSTOM_ERROR_STATUSES` | `Test_OtelSemantics_OTLP_Spans_Http_ErrorStatusConfiguration` | 4 | Configured client and server error-status ranges replace the defaults |
| `OTEL_SEMANTICS_OTLP_TRACE_METRICS` | `Test_OtelSemantics_OTLP_TraceMetrics` | 4 | OTLP trace metrics agree with their corresponding HTTP spans |
| `OTEL_SEMANTICS_OTLP_SAMPLING_RULES` | `Test_OtelSemantics_SamplingRules` | 2 | Semantic resource names are available when sampling decisions occur |

### `Test_OtelSemantics_Spans_Http_Server`

| Test | Description |
| --- | --- |
| `test_otel_attributes_present` | Verifies canonical method, URL, and typed response-status attributes are present. |
| `test_datadog_attributes_absent` | Verifies Datadog HTTP attribute aliases are omitted. |
| `test_span_name_with_route` | Verifies a resolved low-cardinality route contributes to the span name. |
| `test_span_name_route_invariance` | Verifies path-parameter values do not change the route or span name. |
| `test_span_name_without_route` | Verifies an unresolved route never places the concrete URI path in the span name. |
| `test_span_name_unknown_method` | Verifies unknown methods use `_OTHER` attributes and `HTTP` in the span name. |
| `test_status_500_no_exception_is_status_code_error` | Verifies an HTTP 500 response marks the server span as an error. |
| `test_status_400_is_not_error` | Verifies an HTTP 400 response does not mark a server span as an error. |
| `test_status_3xx_is_not_error` | Verifies redirects do not mark server spans as errors. |
| `test_url_query_present_with_query_string` | Verifies `url.query` is emitted when a query string is present. |
| `test_url_query_absent_without_query_string` | Verifies `url.query` is omitted when no query string exists. |
| `test_url_query_obfuscation` | Verifies sensitive query values are obfuscated. |
| `test_user_agent` | Verifies the canonical original user-agent attribute. |
| `test_client_address` | Verifies the canonical client address and removal of its Datadog alias. |
| `test_network_peer_address` | Verifies the network peer address is available. |
| `test_server_address` | Verifies the canonical server address and removal of its Datadog alias. |
| `test_server_port` | Verifies `server.port` is emitted as an OTLP integer. |
| `test_http_endpoint_retained` | Verifies the Datadog endpoint fallback remains available for route-less requests. |
| `test_span_kind_is_server` | Verifies the OTLP span kind is `SERVER`. |

### `Test_OtelSemantics_Spans_Http_Client`

| Test | Description |
| --- | --- |
| `test_otel_attributes_present` | Verifies canonical method, URL, server, and typed response-status attributes are present. |
| `test_datadog_attributes_absent` | Verifies Datadog client HTTP aliases are omitted. |
| `test_peer_service_suppressed` | Verifies peer-service defaults remain disabled in semantic-conventions mode. |
| `test_span_name_is_method` | Verifies the client span name uses the normalized HTTP method without the URI path. |
| `test_span_name_unknown_method` | Verifies unknown methods use `_OTHER` attributes and `HTTP` in the span name. |
| `test_span_kind_is_client` | Verifies the OTLP span kind is `CLIENT`. |
| `test_status_400_is_error` | Verifies an HTTP 400 response marks a client span as an error. |
| `test_status_500_is_error` | Verifies an HTTP 500 response marks a client span as an error. |
| `test_status_3xx_not_error` | Verifies redirects do not mark client spans as errors. |
| `test_url_full_credential_redaction` | Verifies credentials embedded in `url.full` are removed or redacted. |
| `test_url_full_query_obfuscation` | Verifies sensitive query values in `url.full` are obfuscated. |

### `Test_OtelSemantics_OTLP_Spans_Http_ErrorStatusConfiguration`

| Test | Description |
| --- | --- |
| `test_server_error_statuses_config_overrides` | Verifies the configured server range can mark HTTP 200 as an error. |
| `test_server_error_statuses_config_excludes_500` | Verifies replacing the server range can exclude HTTP 500 from errors. |
| `test_client_error_statuses_config_overrides` | Verifies the configured client range can mark HTTP 200 as an error. |
| `test_client_error_statuses_config_excludes_500` | Verifies replacing the client range can exclude HTTP 500 from errors. |

### `Test_OtelSemantics_OTLP_TraceMetrics`

| Test | Description |
| --- | --- |
| `test_trace_metric_agrees_with_client_error` | Verifies an errored client metric matches the span method, status, name, kind, and error decision. |
| `test_trace_metric_agrees_with_client_success` | Verifies a successful client metric matches the span and leaves `status.code` unset. |
| `test_trace_metric_agrees_with_server_error` | Verifies an errored server metric matches the span method, status, name, kind, and error decision. |
| `test_trace_metric_agrees_with_server_success` | Verifies a successful server metric matches the span and leaves `status.code` unset. |

### `Test_OtelSemantics_SamplingRules`

| Test | Description |
| --- | --- |
| `test_otel_span_name_is_available_to_sampling_rules` | Verifies the normalized unknown-method resource matches its keep rule and every exported span is sampled. |
| `test_server_span_name_is_available_before_client_sampling` | Verifies outbound propagation can sample the server trace and keep both server and client spans. |


tests/test_otel_http_semantics.py
```diff
@@ -0,0 +1,859 @@
...
+Transport-specific helpers below normalize those payloads so every semantic assertion is
+identical on both paths. OTLP-only tests additionally validate attribute value types and
+custom error-status configuration.
+
+Two conventions in here are deliberate and load-bearing:
+
+* Span-name expectations are derived from the attributes the span itself reports, never
+  hardcoded to a literal, because the RFC excludes the server span's ``http.route`` value and
+  name value from the cross-tracer contract as framework-specific.
+* Attributes whose specified value type is an int go through ``_assert_int_attribute``, which
```
zacharycmontoya, line 12: We're now only testing the OTLP protocol and not DD protocol, so some of these comments can be revised now

marcotc: TODO: 
1. Remove msgpack transport testing. OTEL semantics sets the transport to OTLP.

tests/test_otel_http_semantics.py
```diff
@@ -0,0 +1,859 @@
...
+    def setup_legacy_attributes_absent(self) -> None:
+        self.response = weblog.get("/", headers={"X-Forwarded-For": _CLIENT_ADDRESS})
+
+    def test_legacy_attributes_absent(self) -> None:
+        span = _server_span(self.response)
+        attrs = _attributes(span)
+        for legacy in _LEGACY_SERVER_KEYS:
+            assert legacy not in attrs, f"legacy Datadog key {legacy!r} must be absent under the flag"
+
+        if isinstance(span, DataDogLibrarySpan):
```
zacharycmontoya: We can remove this now 🎉

tests/test_otel_http_semantics.py
```diff
@@ -0,0 +1,859 @@
...
+
+        The RFC's system-tests section excludes the server span's ``http.route`` value and the
+        server span's name value from the cross-tracer contract because they are
+        framework-specific, so the expectation is derived from the span rather than hardcoded
+        to a literal such as ``GET /users``.
+        """
+        span = _server_span(self.response)
+        route = _attributes(span).get("http.route")
+        assert route, "http.route must be published when the framework resolved a route"
+        assert route != "/sample_rate_route/1", "http.route must be a low-cardinality route template"
```
zacharycmontoya, line 358: Let's find a different endpoint to test here. `/sample_rate_route/1` is not a low-cardinality route template for us to test against

tests/test_otel_http_semantics.py
```diff
@@ -0,0 +1,859 @@
...
+        assert _span_name(span) == _expected_span_name(span, "http.route")
+        _assert_name_has_no_uri_path(span)
```
zacharycmontoya: In this case, we don't need an additional check. If the span name resolves to `{http.request.method} {http.route}` then by definition it is not resolving to `{http.request.method} {uri path}`. We may also not need this in other test methods if we already assert that span name is exactly a certain value

tests/test_otel_http_semantics.py
zacharycmontoya: Sorry, I misinterpreted the assertion so ignore my previous comment. Perhaps what we can do is create a list of valid results and just assert `route in valid_results`. That way we can see from reading the test code what a valid http.route looks like

tests/test_otel_http_semantics.py
```diff
@@ -0,0 +1,859 @@
...
+        """Bare ``{method}`` when no route resolved, and never the URI path.
+
+        Conditioned on the route the span itself reports rather than assuming a 404 means no
+        route: frameworks with a catch-all (Spring resolves ``/**``, a Go mux resolves ``/``)
+        publish a real low-cardinality route for an unmatched path, and ``{method} {route}`` is
+        the correct name there. Hardcoding ``"GET"`` tests the weblog's route table, not the
+        tracer. The URI-path negative below is what actually guards the MUST NOT.
+        """
+        span = _server_span(self.response)
+        assert _span_name(span) == _expected_span_name(span, "http.route")
```
zacharycmontoya, line 395: In the cases where we know what the route should be, can we avoid calling `_expected_span_name` and simply put a string literal here?


```suggestion
        assert _span_name(span) == "GET"
```

tests/test_otel_http_semantics.py
```diff
@@ -0,0 +1,859 @@
...
+        """
+        span = _server_span(self.response)
+        attrs = _attributes(span)
+        assert attrs.get("http.request.method") == "_OTHER"
+        assert attrs.get("http.request.method_original") == "PROPFIND"
+        assert _span_name(span) == _expected_span_name(span, "http.route")
+        assert "PROPFIND" not in _span_name(span), "the raw method must never appear in the span name"
+        _assert_name_has_no_uri_path(span)
+
+    def setup_span_name_route_invariance(self) -> None:
```
zacharycmontoya: I don't think we need this test if you take my example above to assert specific allowable values for `http.route`

tests/test_otel_http_semantics.py
```diff
@@ -0,0 +1,859 @@
...
+            f"span name must be invariant across parameter values: {_span_name(low)!r} vs {_span_name(high)!r}"
+        )
+        assert "99999" not in str(route_high), f"the path parameter leaked into http.route {route_high!r}"
+        assert "99999" not in _span_name(high), f"the path parameter leaked into the span name {_span_name(high)!r}"
+
+    def setup_status_500_no_exception_is_status_code_error(self) -> None:
+        self.response = weblog.get("/status", params={"code": "500"})
+
+    def test_status_500_no_exception_is_status_code_error(self) -> None:
+        _assert_status_error(_server_span(self.response), 500)
```
zacharycmontoya: nit: Can you remove the helper method `_assert_status_error` and inline its instructions here? Even though this will add lines of code, it's much easier for me to see the test name, the test input, and the test assertions all right next to each other

tests/test_otel_http_semantics.py
```diff
@@ -0,0 +1,859 @@
...
+
+    def setup_status_400_is_not_error(self) -> None:
+        self.response = weblog.get("/status", params={"code": "400"})
+
+    def test_status_400_is_not_error(self) -> None:
+        _assert_status_not_error(_server_span(self.response), 400)
```
zacharycmontoya: nit: same comment for `_assert_status_not_error`

tests/test_otel_http_semantics.py
```diff
@@ -0,0 +1,859 @@
...
+    def setup_server_port(self) -> None:
+        self.response = weblog.get("/")
+
+    def test_server_port(self) -> None:
+        """``server.port`` is Conditionally Required on server spans and its value type is int.
+
+        The RFC's own test tables only check ``server.port`` on client spans, but the server
+        attribute table requires it too, and it is subject to the same two-transport typing rule
+        as the status code.
+        """
```
zacharycmontoya:
```suggestion
```

tests/test_otel_http_semantics.py
```diff
@@ -0,0 +1,859 @@
...
+        span = _server_span(self.response)
+        if isinstance(span, DataDogLibrarySpan):
+            assert span.meta.get("span.kind") == "server", f"got span.kind {span.meta.get('span.kind')!r}"
+        else:
+            assert span.get("kind") == SpanKind.SERVER.value, f"got OTLP kind {span.get('kind')!r}"
+
+    def setup_http_route_retained(self) -> None:
+        self.response = weblog.get("/sample_rate_route/1")
+
+    def test_http_route_retained(self) -> None:
```
zacharycmontoya: This test looks redundant since we already have `test_span_name_with_route`. I think we can remove it

tests/test_otel_http_semantics.py
```diff
@@ -0,0 +1,859 @@
...
+            assert legacy not in attrs, f"legacy Datadog key {legacy!r} must be absent under the flag"
+
+        if isinstance(span, DataDogLibrarySpan):
+            for legacy in _LEGACY_CLIENT_KEYS:
+                assert legacy not in span.metrics, f"legacy Datadog key {legacy!r} must be absent from metrics"
+
+    def setup_url_path_absent_on_client(self) -> None:
+        self.response = _distant_call("http://weblog:7777/sample_rate_route/1?otel=visible")
+
+    def test_url_path_absent_on_client(self) -> None:
```
zacharycmontoya: We already have a positive test that url.full is specified. I'm not sure if we need to go out of our way to assert that `url.path` is absent

tests/test_otel_http_semantics.py
```diff
@@ -0,0 +1,859 @@
...
+        # while normalizing request options. If userinfo remains, both fields must be redacted.
+        if "@" in full_url:
+            assert "REDACTED:REDACTED@" in full_url
+
+    def setup_url_full_query_obfuscation(self) -> None:
+        self.response = _distant_call(f"http://weblog:7777/?token={_SENSITIVE_QUERY_VALUE}")
+
+    def test_url_full_query_obfuscation(self) -> None:
+        full_url = str(_attributes(_client_span(self.response)).get("url.full", ""))
+        assert full_url.startswith("http://weblog:7777/?token=")
```
zacharycmontoya, line 647: Can we assert `full_url == "http://weblog:7777/?token=<redacted>"`?

zacharycmontoya: Or are you trying to future-proof this test case by allowing the OTel `REDACTED` to be inserted here instead of `<redacted>` and still pass?

tests/test_otel_http_semantics.py
```diff
@@ -0,0 +1,859 @@
...
+        assert _SENSITIVE_QUERY_VALUE not in full_url
+        assert "redacted" in full_url.lower()
+
+
+@rfc("https://docs.google.com/document/d/1SONUGEa38eLumE5b6gnNhykFhzZL9uQpsnMFq06uDMY/edit")
+@features.semantic_core_validations
+@scenarios.otel_semantics_otlp
+class Test_OtelSemantics_OTLP_Server:
```
zacharycmontoya: I actually don't think we need this class any more because you're already making the same assertions in the `Test_OtelSemantics_Spans_Http_Server` test class

tests/test_otel_http_semantics.py
```diff
@@ -0,0 +1,859 @@
...
+    def test_legacy_attributes_absent_otlp(self) -> None:
+        attrs = _attributes(_server_span(self.response))
+        for legacy in _LEGACY_SERVER_KEYS:
+            assert legacy not in attrs, f"legacy Datadog key {legacy!r} must be absent under the flag"
+
+
+@rfc("https://docs.google.com/document/d/1SONUGEa38eLumE5b6gnNhykFhzZL9uQpsnMFq06uDMY/edit")
+@features.semantic_core_validations
+@scenarios.otel_semantics_otlp
+class Test_OtelSemantics_OTLP_Client:
```
zacharycmontoya: Same thing here, I think we can remove this class since this is duplicating the assertions done in Test_OtelSemantics_Spans_Http_Client

tests/test_otel_http_semantics.py
```diff
@@ -0,0 +1,859 @@
...
+            assert legacy not in attrs, f"legacy Datadog key {legacy!r} must be absent under the flag"
+
+
+@rfc("https://docs.google.com/document/d/1SONUGEa38eLumE5b6gnNhykFhzZL9uQpsnMFq06uDMY/edit")
+@features.semantic_core_validations
+@scenarios.otel_semantics_otlp_custom_error_statuses
+class Test_OtelSemantics_OTLP_Spans_Http_ErrorStatusConfiguration:
+    """A user-configured error status range takes precedence over the OTel default.
+
+    Server spans only. ``DD_TRACE_HTTP_CLIENT_ERROR_STATUSES`` exists in Java, .NET, Go, Ruby
```
zacharycmontoya: Please add test cases for DD_TRACE_HTTP_CLIENT_ERROR_STATUSES. If it doesn't exist in a language we can skip it with an `irrelevant` tag

tests/test_otel_http_semantics.py
```diff
@@ -0,0 +1,859 @@
...
+        assert not _span_is_error(span), (
+            "HTTP 500 is outside the configured server error range, so the span must not be an error"
+        )
+        assert "error.type" not in _attributes(span)
+
+
+@rfc("https://docs.google.com/document/d/1SONUGEa38eLumE5b6gnNhykFhzZL9uQpsnMFq06uDMY/edit")
+@features.semantic_core_validations
+@scenarios.otel_semantics_stats
+class Test_OtelSemantics_Stats_Consistency:
```
zacharycmontoya: Since we're no longer emitting DD MsgPack, we're no longer doing our traditional client-side stats. We'll need to instead test against Munir's OTLP Trace Metrics to make sure the errors and whatnot line up

tests/test_otel_http_semantics.py
```diff
@@ -0,0 +1,682 @@
...
+
+@rfc("https://docs.google.com/document/d/1SONUGEa38eLumE5b6gnNhykFhzZL9uQpsnMFq06uDMY/edit")
+@features.semantic_core_validations
+@scenarios.otel_semantics_otlp_sampling_rules
+class Test_OtelSemantics_SamplingRules:
+    """Sampling rules consume the OTel span name before the trace is exported."""
+
+    def setup_otel_span_name_is_available_to_sampling_rules(self) -> None:
+        self.control_response = weblog.get("/")
+        self.matched_response = weblog.request("PROPFIND", "/")
```
zacharycmontoya, line 769: Let's also do a call to `/make_distant_call` because when that outbound http client request is made, we have to immediately do sampling on the server span. Maybe you can test by making a call to `/make_distant_call` and have a sampling rule `{"resource":"GET *make_distant_call","sample_rate":1.0}` so we can match resource names of `GET /make_distant_call` or `GET make_distant_call` and make sure we keep the HTTP server AND HTTP client span

tests/test_otel_http_semantics.py
link04: I addressed this with framework-independent assertions instead of a fixed list. The RFC excludes the exact server http.route value from the cross-tracer contract. The tests now verify that the route is low-cardinality, differs from the concrete URI, and remains invariant across different path-parameter values.

tests/test_otel_http_semantics.py
link04: I kept this framework-independent because an unmatched request may still resolve to a framework catch-all route. The expected name is derived from the route reported by the span, and the test separately verifies that the concrete URI path never appears in the name.

tests/test_otel_http_semantics.py
link04: Yes, this is intentionally future-proofed. Implementations may represent the value as <redacted> or REDACTED. The test verifies the expected URL prefix, confirms the sensitive value is absent, and requires a redaction marker without enforcing one tracer-specific representation.

tests/test_otel_http_semantics.py
link04: Added both client configuration tests: one verifies that the configured range can mark HTTP 200 as an error, and the other verifies that replacing the range can exclude HTTP 500. Unsupported implementations use missing_feature rather than irrelevant, because the behavior is applicable but not yet implemented.

tests/test_otel_http_semantics.py
link04: Added this coverage. The test calls /make_distant_call, correlates the server and outbound client spans to the same trace, verifies the matching rule through _dd.rule_psr, and checks that both spans carry the sampled flag. The rule uses GET* because sampling can occur before or after route resolution, covering both GET and GET <route> consistently.

docs/understand/weblogs/end-to-end_weblog.md
marcotc, line 561: I think it's helpful to add an example of the `method` parameter in this snippet.

Also, let's define `method` more strictly: I recommend requiring it to be all uppercase, so weblog implementations can confidently match against `== "GET"` and not `case_insensitive("get")`.

tests/test_otel_http_semantics.py
```diff
@@ -0,0 +1,772 @@
...
+    return attributes if isinstance(attributes, dict) else {}
+
+
+def _assert_int_attribute(span: HttpSpan, key: str, expected: int) -> None:
+    """Assert that an OTLP attribute retained its required integer type."""
+    value = _attributes(span).get(key)
+    assert value is not None, f"{key} is absent from the OTLP export"
+    assert not isinstance(value, bool), f"{key} must be an OTLP intValue, got a boolValue ({value!r})"
+    assert isinstance(value, int), f"{key} must be exported as an OTLP intValue, got {type(value).__name__} ({value!r})"
+    assert value == expected, f"expected {key} == {expected}, got {value}"
```
marcotc: I think this could be just:
```suggestion
    assert type(value) is int, f"expected {key} to be int, got {type(value).__name__} ({value!r})"
    assert value == expected
```

tests/test_otel_http_semantics.py
```diff
@@ -0,0 +1,772 @@
...
+    def setup_otel_attributes_present(self) -> None:
+        self.response = weblog.get("/")
+
+    def test_otel_attributes_present(self) -> None:
+        span = _server_span(self.response)
+        attrs = _attributes(span)
+        assert attrs.get("http.request.method") == "GET"
+        assert attrs.get("url.path") == "/"
+        assert attrs.get("url.scheme") in ("http", "https")
+        _assert_int_attribute(span, "http.response.status_code", 200)
```
marcotc, line 334: What's the concern with just doing?
```suggestion
        assert attrs.get("http.response.status_code") == 200
```

tests/test_otel_http_semantics.py
```diff
@@ -0,0 +1,772 @@
...
+
+type HttpSpan = dict[str, Any]
+
+_DISTANT_CALL_TARGET = "http://weblog:7777/"
+_SENSITIVE_QUERY_VALUE = "otel-sensitive-value"
+_CLIENT_ADDRESS = "203.0.113.42"
+
+# Datadog keys that have an OTel equivalent and so must be replaced, not duplicated, under the
+# flag. http.endpoint is deliberately not here: it has no OTel equivalent and is retained.
+_LEGACY_SERVER_KEYS = (
```
marcotc: Let's use "dd" or "datadog" instead of legacy, for the whole PR.

marcotc: Change this in the PR description too (search for `legacy`).

marcotc: Addressed.

marcotc:
@codex review

marcotc:
@codex review

manifests/python.yml
```diff
@@ -2440,6 +2440,15 @@ manifest:
   tests/test_library_logs.py::Test_NoExceptions::test_dotnet: irrelevant (only for .NET)
   tests/test_library_logs.py::Test_NoExceptions::test_java_logs: irrelevant (only for Java)
   tests/test_library_logs.py::Test_NoExceptions::test_java_telemetry_logs: irrelevant (only for Java)
+  tests/test_otel_http_semantics.py::Test_OtelSemantics_OTLP_Spans_Http_ErrorStatusConfiguration: 4.15.0-dev
```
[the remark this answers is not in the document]
marcotc: This exact activation is intentional. `v4.15.0-dev` also matches `4.15.0-rc1`, but that RC does not contain this feature and caused the current CI failures. The manifest should move back to a minimum-version declaration once a release containing the feature is available.

utils/_context/_scenarios/__init__.py
```diff
@@ -261,6 +261,89 @@ class _Scenarios:
...
+        "client and server error status configuration takes precedence over the OTel defaults",
+        scenario_groups=[scenario_groups.open_telemetry],
+    )
+
+    otel_semantics_otlp_trace_metrics = DdTraceEndToEndScenario(
+        "OTEL_SEMANTICS_OTLP_TRACE_METRICS",
+        weblog_env={
+            "DD_TRACE_OTEL_SEMANTICS_ENABLED": "true",
+            "DD_TRACE_OTEL_ENABLED": "true",
+            "OTEL_EXPORTER_OTLP_PROTOCOL": "http/protobuf",
```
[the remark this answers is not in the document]
marcotc: Valid. Fixed in 99cfa518b by selecting `http/json` specifically for the metrics exporter while retaining `http/protobuf` for traces.

manifests/python.yml
```diff
@@ -2440,6 +2440,15 @@ manifest:
...
+  ? tests/test_otel_http_semantics.py::Test_OtelSemantics_OTLP_Spans_Http_ErrorStatusConfiguration::test_client_error_statuses_config_excludes_500
+  : missing_feature (DD_TRACE_HTTP_CLIENT_ERROR_STATUSES support is outside the current Python OTel semantics implementation scope)
+  ? tests/test_otel_http_semantics.py::Test_OtelSemantics_OTLP_Spans_Http_ErrorStatusConfiguration::test_client_error_statuses_config_overrides
+  : missing_feature (DD_TRACE_HTTP_CLIENT_ERROR_STATUSES support is outside the current Python OTel semantics implementation scope)
+  tests/test_otel_http_semantics.py::Test_OtelSemantics_OTLP_TraceMetrics: 4.15.0-dev
+  tests/test_otel_http_semantics.py::Test_OtelSemantics_SamplingRules: 4.15.0-dev
+  tests/test_otel_http_semantics.py::Test_OtelSemantics_Spans_Http_Client: 4.15.0-dev
+  tests/test_otel_http_semantics.py::Test_OtelSemantics_Spans_Http_Server: 4.15.0-dev
```
zacharycmontoya, line 2448: Are the Python changes merged yet? If not, we should leave this as missing_feature

manifests/nodejs.yml
```diff
@@ -2852,6 +2852,27 @@ manifest:
   tests/test_library_logs.py::Test_NoExceptions::test_dotnet: irrelevant (only for .NET)
   tests/test_library_logs.py::Test_NoExceptions::test_java_logs: irrelevant (only for Java)
   tests/test_library_logs.py::Test_NoExceptions::test_java_telemetry_logs: irrelevant (only for Java)
+  tests/test_otel_http_semantics.py::Test_OtelSemantics_OTLP_Spans_Http_ErrorStatusConfiguration: missing_feature (dd-trace-js has no HTTP status environment configuration for these client or server override tests)
+  tests/test_otel_http_semantics.py::Test_OtelSemantics_OTLP_TraceMetrics:
+    - weblog_declaration:
+        "*": irrelevant (the shared HTTP instrumentation is covered by express4)
```
zacharycmontoya: We shouldn't be marking the other ones (`*`) as irrelevant, we should only mark them as `missing_feature`

tests/otel/test_tracing_otlp.py
```diff
@@ -152,9 +164,21 @@ def test_128bit_trace_id_consistent_across_spans(self):
         """
         data = list(interfaces.open_telemetry.get_otel_spans(self.req))
 
-        # `get_otel_spans` yields the server span, identified by the user-agent header
-        assert len(data) >= 1, f"Expected at least one matching OTLP span, got {data}"
-        _, content, server_span = data[0]
+        request_id = self.req.get_rid()
+        server_spans = []
```
zacharycmontoya: nit: Can we introduce a helper method to get a server span by request_id?

tests/test_otel_http_semantics.py
```diff
@@ -0,0 +1,795 @@
...
+
+    def test_span_name_route_invariance(self) -> None:
+        """Resolved routes and route-derived names remain stable across path parameters."""
+        first_span = _server_span(self.first_response)
+        second_span = _server_span(self.second_response)
+        first_route = _attributes(first_span).get("http.route")
+        second_route = _attributes(second_span).get("http.route")
+        assert first_route, "the first request must report a resolved http.route"
+        assert second_route, "the second request must report a resolved http.route"
+        assert first_route == second_route, "http.route must not vary with path-parameter values"
```
zacharycmontoya, line 377: nit: I think it would be good to assert that the first url_path == /sample_rate_route/` and the second url_path == /sample_rate_route/2 to make sure we're comparing two different spans

tests/test_otel_http_semantics.py
```diff
@@ -0,0 +1,795 @@
...
+        algorithm never looks at. The ``HTTP `` prefix negative is the one that earns its keep:
+        upstream otelhttp names client spans ``HTTP GET``, which is neither the method nor
+        ``{method} {url.template}``.
+        """
+        span = _client_span(self.response)
+        name = _span_name(span)
+        assert name == _expected_span_name(span, "url.template")
+        assert not name.startswith("HTTP "), (
+            f"the client span name must be {{method}} or {{method}} {{url.template}}, got {name!r}"
+        )
```
zacharycmontoya: I personally think we can remove this. We're already asserting an exact value for the name, so we don't gain much from making another assertion

```suggestion
```

zacharycmontoya (approved):
I've left several small comments to clean up some span collection and assertions, but the overall logic looks fine to me. And .NET is passing the HTTP client testing so 👍🏼

tests/test_otel_http_semantics.py
```diff
@@ -0,0 +1,795 @@
...
+        self.response = _distant_call("http://weblog:7777/status?code=201")
+
+    def test_trace_metric_agrees_with_client_success(self) -> None:
+        span = _client_span(self.response)
+        _assert_int_attribute(span, "http.response.status_code", 201)
+        assert not _span_is_error(span), "HTTP 201 client span must not be an error"
+        attrs = _metric_attributes(_client_trace_metric(span))
+        assert attrs.get("http.request.method") == "GET"
+        assert attrs.get("http.response.status_code") == 201
+        assert "status.code" not in attrs, "successful HTTP trace metrics must leave OTel status unset"
```
zacharycmontoya, line 735: This is failing for .NET, because successful trace metrics are currently setting the `status.code=STATUS_CODE_OK`

tests/test_otel_http_semantics.py
```diff
@@ -0,0 +1,795 @@
...
+        self.response = weblog.get("/status?code=202")
+
+    def test_trace_metric_agrees_with_server_success(self) -> None:
+        span = _server_span(self.response)
+        _assert_int_attribute(span, "http.response.status_code", 202)
+        assert not _span_is_error(span), "HTTP 202 server span must not be an error"
+        attrs = _metric_attributes(_server_trace_metric(span))
+        assert attrs.get("http.request.method") == "GET"
+        assert attrs.get("http.response.status_code") == 202
+        assert "status.code" not in attrs, "successful HTTP trace metrics must leave OTel status unset"
```
zacharycmontoya, line 759: This is failing for .NET, because successful trace metrics are currently setting the `status.code=STATUS_CODE_OK`

xlamorlette-datadog (approved):
OK for the C++ part.

cbeauchesne (approved):

manifests/python.yml
marcotc: Yes. The Python changes are merged: dd-trace-py #18495 added the opt-in semantics support and #17718 added the HTTP resource behavior. The exact `4.15.0-dev` activation is intentional because `4.15.0-rc1` predates these changes.

manifests/nodejs.yml
marcotc: Fixed in e3ec976cc. Non-Express weblog declarations now use `missing_feature`, not `irrelevant`, for all affected classes.

tests/otel/test_tracing_otlp.py
marcotc: Fixed in e3ec976cc. `_server_span_by_request_id` now owns the correlation and uniqueness check used by both tests.

tests/test_otel_http_semantics.py
marcotc: Fixed in e3ec976cc. The test now verifies `/sample_rate_route/1` and `/sample_rate_route/2` before comparing their route templates and names.

tests/test_otel_http_semantics.py
marcotc: Fixed in e3ec976cc. Removed the redundant prefix assertion and its supporting explanation; exact expected-name equality already covers it.

tests/test_otel_http_semantics.py
marcotc: Fixed in e3ec976cc. Added a focused .NET `missing_feature` declaration for the successful client trace-metric status gap.

tests/test_otel_http_semantics.py
marcotc: Fixed in e3ec976cc. Added the corresponding .NET `missing_feature` declaration for successful server trace metrics.

marcotc:
I rewrote commit history (!) to fix previous unsigned commits.

PR #7488, opened 2026-08-10
perf(ffe): fix FEATURE_FLAGGING_AND_EXPERIMENTATION CI regression on Node.js

rochdev:
## Summary
- `FEATURE_FLAGGING_AND_EXPERIMENTATION` regressed from <3 minutes to 7+ minutes in dd-trace-js CI (nodejs express4 shard).
- Two independent costs were identified from the real CI job log/timing breakdown, and both are fixed here:

### 1. Exposures: waiting on a periodic flush timer instead of flushing immediately
- `tests/ffe/test_exposures.py`'s `wait_for_exposure_event`/`wait_for_min_exposure_count` helpers were purely polling `interfaces.agent.wait_for(..., timeout=30)` for exposure events, at the mercy of the exposures writer's periodic flush interval.
- This adds a `weblog.get("/flush")` call before polling, so the weblog forces its exposures writer to flush immediately via the existing `/flush` endpoint pattern, instead of waiting on the timer.
- Wired the Node.js side: the 4 Node weblogs (`express`, `express4-typescript`, `fastify`, `nextjs`) now publish the `ffe:writers:flush` diagnostics channel (already consumed inside dd-trace-js) from their `/flush` handlers.
- `weblog.get("/flush")` is a no-op for weblogs/languages that don't implement the route, so this is safe across all libraries.
- Confirmed in CI: `test_exposures.py` now completes in ~30s instead of stalling on the writer's timer.

### 2. EVP flagevaluation: `missing_feature` tests were still executing under `xfail`, burning their full timeout
- `missing_feature` (and `bug`, etc.) map to `pytest.mark.xfail`, not `skip` (`utils/_decorators.py`) — the test body still runs. dd-trace-js has no implementation of `/api/v2/flagevaluation` at all, so `tests/ffe/test_flag_eval_evp.py`'s 8 non-`@scenario_crash` classes each ran to completion and burned their full 30s `wait_for` timeout before failing as expected. That's ~240s, the actual dominant cost of the scenario.
- Adds `@slow` (existing `skip_if_xfail` mechanism, already used on 3 sibling classes in this file via `@scenario_crash`) to the remaining 8 classes. `@slow` only converts a test to a hard skip when it's paired with the manifest's `declaration` marker for the *current* library — so this only skips for libraries where the feature is currently declared missing (today: Node.js for all 8, plus golang for the 3 `ObserveFullData_*` classes it doesn't yet implement).
- Self-healing: once a library implements the feature and its manifest entry is removed, `@slow` becomes a no-op and the test runs for real. Confirmed against golang's actual CI timing (the one library that implements this today) that real event delivery is on the order of 1-2s, nowhere near the 30s ceiling, so no added flakiness risk anywhere.

## Test plan
- [ ] Re-run the `FEATURE_FLAGGING_AND_EXPERIMENTATION` scenario for nodejs (express4/express4-typescript/fastify/nextjs) in CI and confirm it completes well under 3 minutes.
- [ ] Confirm `tests/ffe/test_exposures.py` still passes for nodejs.
- [ ] Confirm `tests/ffe/test_flag_eval_evp.py` still passes for golang (only the 3 `ObserveFullData_*` classes should skip there).
- [ ] Confirm other languages are unaffected.

### DIFF
--- a/utils/build/docker/nodejs/express4-typescript/app.ts
+++ b/utils/build/docker/nodejs/express4-typescript/app.ts
@@ -388,6 +388,8 @@ app.get('/flush', (req: Request, res: Response) => {
   tracer.dogstatsd?.flush?.()
   tracer._pluginManager?._pluginsByName?.openai?.metrics?.flush?.()
   tracer._tracer?._processor?._stats?.onInterval()
+  // force FFE exposure events out immediately instead of waiting for the writer's periodic flush
+  require('diagnostics_channel').channel('ffe:writers:flush').publish()
 
   // does have a callback :)
   const promises = []
--- a/tests/ffe/test_exposures_datadog_agent.py
+++ b/tests/ffe/test_exposures_datadog_agent.py
@@ -31,6 +31,10 @@ def find_exposure_events(flag_key: str, subject_id: str | None = None) -> list[d
 
 def wait_for_exposure_event(flag_keys: set[str], subject_id: str | None = None) -> None:
     """Wait until the agent receives an exposure event for one of the given flags."""
+    # ask the weblog to flush its exposure writer instead of waiting on its periodic timer.
+    # weblogs that don't support this route (or don't buffer exposures) just no-op here, and the
+    # wait_for below still covers them.
+    weblog.get("/flush")
     assert interfaces.agent.wait_for(
         lambda data: bool(exposure_events_from_data(data, flag_keys, subject_id)),
         timeout=EXPOSURE_WAIT_TIMEOUT_SECONDS,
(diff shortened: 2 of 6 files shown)


greghuels (approved):

utils/build/docker/nodejs/express4-typescript/app.ts
```diff
@@ -388,6 +388,8 @@ app.get('/flush', (req: Request, res: Response) => {
   tracer.dogstatsd?.flush?.()
   tracer._pluginManager?._pluginsByName?.openai?.metrics?.flush?.()
   tracer._tracer?._processor?._stats?.onInterval()
+  // force FFE exposure events out immediately instead of waiting for the writer's periodic flush
+  require('diagnostics_channel').channel('ffe:writers:flush').publish()
```
leoromanovsky, line 392: This is cool thanks

leoromanovsky (approved):

cbeauchesne (approved):

PR #7516, opened 2026-08-14
[golang] Add apim weblog variant for azure/apim-callout

eliottness:
## Motivation

`dd-trace-go` ships an Azure API Management integration (`contrib/azure/apim-callout`) that had no system-tests coverage. The other two Go proxy integrations, Envoy and HAProxy, each have a weblog variant; APIM did not, so the callout's behaviour was only covered by that repo's own unit tests.

The obstacle is that Azure APIM has no runnable container — it is a managed Azure service. There is nothing to `docker run` as the gateway.

## Changes

A new Go weblog variant `apim`, at activation parity with `envoy`.

Since APIM itself cannot run locally, the gateway role is filled by a small system-tests-owned Go shim that reimplements APIM `send-request` **behaviour**. It deliberately does not parse policy XML — it reproduces what the policy *does*:

```
test runner --> apim-gateway (shim, :80 -> host 127.0.0.1:7777)
                  |-> apim-callout (POST :8080, health :8081)   <- library under test
                  `-> http-app (:8080)
```

The shim is stdlib-only, has no `require` block, and adds no configuration surface beyond a single control header.

**Body delivery modes.** The callout supports two, and testing only one would leave half its body-handling code dark. Default is the deferred 4-call flow. `X-Datadog-Apim-Body-Mode: inline` switches both sides to inline delivery, which suppresses `allowed-body-size` and collapses the exchange to 2 calls. The header is stripped before the callout payload and before the upstream forward, so it never reaches the WAF or the app, and the shared suite never sends it.

**Fail-closed, deliberately unlike APIM.** The real policy uses `ignore-error="true"`. The shim instead returns 502 plus a stderr diagnostic on detectable failure, so system-tests can assert on it. One case is genuinely undetectable and documented as such: the callout returns HTTP 200 `{}` both for processor errors and as a legitimate phase-2/4 success.

**Manifest.** All 89 `envoy:` rows are widened to `envoy, apim:`, transcribing enabled rows as enabled. Widening all 89 rather than only the non-redundant ones is necessary, not merely tidy: 29 rows have a `"*"` sibling and in every one the `"*"` value differs from envoy's, so for 13 of them the fallback would have handed `apim` an outright wrong declaration.

One row is declared `missing_feature` for `apim` rather than enabled — `test_span_links_from_conflicting_contexts`. This is a real library gap, not a shim defect: `apim-callout` defaults `DD_TRACE_PROPAGATION_STYLE` to `datadog`, so a conflicting W3C `traceparent` is never extracted and no terminated-context span link is created. `haproxy-spoa` ships the identical default and is already `missing_feature` on that exact row; envoy's processor leaves the variable unset and gets the tracer default, which is why it passes. Recording the gap was preferred over overriding the variable in the harness, which would test a configuration no default deployment runs.

`containers.py:998` also gains `apim`. Without it `get_image_list` tries to open a nonexistent `apim.Dockerfile` and breaks `get-image-list.py` **while every test still passes** — a silent failure.

### CI

`apim` is picked up automatically by `compute-workflow-parameters.py` (verified: it appears in `endtoend_defs_parallel_jobs` with `weblog_build_required: false` and scenarios `[APPSEC_BLOCKING, DEFAULT]`), so no workflow changes were needed. Like `envoy` and `haproxy`, it uses the `:latest` callout image in normal PR CI and `:dev` in `_system_tests_dev_mode`, where `load-binary.sh` writes the pointer.

### Validation

Both scenarios were run locally against `apim-callout:dev` (`golang@2.11.0-dev.1`): DEFAULT `128 passed` and APPSEC_BLOCKING `50 passed`, with zero enabled-row failures. Also verified by hand: a live 403 WAF block through the shim, inline yielding exactly 2 callout calls and deferred exactly 4 (each sharing one `request-id`), and the fail-closed 502 path.

Note a bodiless GET correctly produces **3** calls, not 4 — the callout only returns `allowed-body-size` when it actually wants a request body.

One caveat: `DummyServerContainer` hardcodes an amd64-only image, so local validation on arm64 required substituting a native `http-app`. This affects `envoy` and `haproxy` identically and is pre-existing; amd64 CI exercises the stock topology.

## Workflow

## Reviewer checklist

* [x] Anything but `tests/` or `manifests/` is modified ? I have the approval from [R&P team](https://dd.enterprise.slack.com/archives/C025TJ4RZ8X)
    * **Framework changes are involved and need R&P review**: a new weblog variant and two new container classes in `utils/_context/containers.py`, the `GoProxyWeblogs` union and dispatch in `utils/_context/weblog_infrastructure.py`, the `containers.py:998` image-list guard, plus `utils/scripts/load-binary.sh` and the new shim under `utils/build/docker/golang/apim/`.
    * No base image is modified. The variant is `build_mode: none` and runs `golang:1.25-alpine` (already mirrored) with the shim bind-mounted read-only.
    * No scenario is added, removed or renamed. `apim` is restricted to the existing `DEFAULT` and `APPSEC_BLOCKING`.

### DIFF
--- a/manifests/golang.yml
+++ b/manifests/golang.yml
@@ -1155,6 +1155,22 @@ manifest:
   tests/debugger/test_debugger_symdb.py::Test_Debugger_SymDb::test_event_metadata: missing_feature (extended event schema not yet shipped)
   tests/debugger/test_debugger_telemetry.py::Test_Debugger_Telemetry: missing_feature
   tests/docker_ssi/test_docker_ssi_appsec.py::TestDockerSSIAppsecFeatures::test_telemetry_source_ssi: v2.0.0
+  tests/external_processing/test_apim_callout.py::Test_ApimCallout::test_default_body_mode_uses_four_callouts:
+    - weblog_declaration:
+        "*": irrelevant (only applies to apim)
+        apim: v2.7.1
+  tests/external_processing/test_apim_callout.py::Test_ApimCallout::test_inline_body_mode_closes_request_state:
+    - weblog_declaration:
+        "*": irrelevant (only applies to apim)
+        apim: v2.7.1
+  tests/external_processing/test_apim_callout.py::Test_ApimCallout::test_inline_body_mode_preserves_trace_structure:
+    - weblog_declaration:
+        "*": irrelevant (only applies to apim)
+        apim: v2.7.1
+  tests/external_processing/test_apim_callout.py::Test_ApimCallout::test_inline_body_mode_uses_two_callouts:
+    - weblog_declaration:
+        "*": irrelevant (only applies to apim)
+        apim: v2.7.1
   tests/ffe/test_agentless_configuration.py: missing_feature (FFL-2695 tracks Go agentless configuration-source implementation; FFL-2731 tracks the system-tests contract)
   tests/ffe/test_dynamic_evaluation.py::Test_FFE_Flag_Parse_Error_Isolation::test_valid_flag_unaffected: bug (FFL-2184)
   tests/ffe/test_dynamic_evaluation.py::Test_FFE_RC_Down_From_Start: v2.4.0
--- a/manifests/golang.yml
+++ b/manifests/golang.yml
@@ -1155,22 +1155,6 @@ manifest:
   tests/debugger/test_debugger_symdb.py::Test_Debugger_SymDb::test_event_metadata: missing_feature (extended event schema not yet shipped)
   tests/debugger/test_debugger_telemetry.py::Test_Debugger_Telemetry: missing_feature
   tests/docker_ssi/test_docker_ssi_appsec.py::TestDockerSSIAppsecFeatures::test_telemetry_source_ssi: v2.0.0
-  tests/external_processing/test_apim_callout.py::Test_ApimCallout::test_default_body_mode_uses_four_callouts:
-    - weblog_declaration:
-        "*": irrelevant (only applies to apim)
-        apim: v2.7.1
-  tests/external_processing/test_apim_callout.py::Test_ApimCallout::test_inline_body_mode_closes_request_state:
-    - weblog_declaration:
-        "*": irrelevant (only applies to apim)
-        apim: v2.7.1
-  tests/external_processing/test_apim_callout.py::Test_ApimCallout::test_inline_body_mode_preserves_trace_structure:
-    - weblog_declaration:
-        "*": irrelevant (only applies to apim)
-        apim: v2.7.1
-  tests/external_processing/test_apim_callout.py::Test_ApimCallout::test_inline_body_mode_uses_two_callouts:
-    - weblog_declaration:
-        "*": irrelevant (only applies to apim)
-        apim: v2.7.1
   tests/ffe/test_agentless_configuration.py: missing_feature (FFL-2695 tracks Go agentless configuration-source implementation; FFL-2731 tracks the system-tests contract)
   tests/ffe/test_dynamic_evaluation.py::Test_FFE_Flag_Parse_Error_Isolation::test_valid_flag_unaffected: bug (FFL-2184)
   tests/ffe/test_dynamic_evaluation.py::Test_FFE_RC_Down_From_Start: v2.4.0
@@ -1775,6 +1759,7 @@ manifest:
   tests/test_distributed.py::Test_Span_Links_From_Conflicting_Contexts: v2.2.3
   tests/test_distributed.py::Test_Span_Links_From_Conflicting_Contexts::test_span_links_from_conflicting_contexts:
     - weblog_declaration:
+        apim: missing_feature (apim-callout defaults DD_TRACE_PROPAGATION_STYLE to datadog, so the conflicting W3C context is not extracted)
         haproxy: missing_feature
   tests/test_distributed.py::Test_Span_Links_Omit_Tracestate_From_Conflicting_Contexts: missing_feature (implementation specs have not been determined)
   tests/test_distributed.py::Test_Synthetics_APM_Datadog:  # Modified by easy win activation script
--- a/manifests/golang.yml
+++ b/manifests/golang.yml
@@ -399,7 +399,8 @@ manifest:
         haproxy: irrelevant (not applicable to proxies)
   tests/appsec/test_alpha.py:
     - weblog_declaration:
-        envoy, apim: v1.72.0
+        apim: v2.7.1
+        envoy: v1.72.0
         haproxy: v2.4.0
   tests/appsec/test_alpha.py::Test_Basic:
     - weblog_declaration:
@@ -526,7 +527,8 @@ manifest:
   tests/appsec/test_automated_user_and_session_tracking.py::Test_Automated_User_Tracking: irrelevant (Go only supports SDK events)
   tests/appsec/test_blocking_addresses.py:
     - weblog_declaration:
-        envoy, apim: v1.72.0
+        apim: v2.7.1
+        envoy: v1.72.0
         haproxy: v2.4.0
   tests/appsec/test_blocking_addresses.py::Test_BlockingGraphqlResolvers:
     - weblog_declaration:
@@ -644,7 +646,8 @@ manifest:
         net-http-orchestrion: missing_feature (path params are not supported)
   tests/appsec/test_client_ip.py:
     - weblog_declaration:
-        envoy, apim: v1.72.0
+        apim: v2.7.1
+        envoy: v1.72.0
         haproxy: v2.4.0
   tests/appsec/test_client_ip.py::Test_StandardTagsClientIp: v1.44.1
   tests/appsec/test_conf.py::Test_ConfigurationVariables: v2.0.0
@@ -771,7 +774,8 @@ manifest:
   tests/appsec/test_ip_blocking_full_denylist.py::Test_AppSecIPBlockingFullDenylist:
     - weblog_declaration:
         "*": v1.47.0
-        envoy, apim: v1.72.0
+        apim: v2.7.1
+        envoy: v1.72.0
         haproxy: v2.4.0
   tests/appsec/test_logs.py::Test_Standardization: missing_feature
   tests/appsec/test_logs.py::Test_StandardizationBlockMode: missing_feature
@@ -789,7 +793,8 @@ manifest:
   tests/appsec/test_remote_config_rule_changes.py::Test_UpdateRuleFileWithRemoteConfig: v2.0.0
   tests/appsec/test_reports.py:
     - weblog_declaration:
-        envoy, apim: v1.72.0
+        apim: v2.7.1
+        envoy: v1.72.0
         haproxy: v2.4.0
   tests/appsec/test_reports.py::Test_ExtraTagsFromRule:
     - weblog_declaration:
@@ -817,7 +822,8 @@ manifest:
   tests/appsec/test_request_blocking.py::Test_AppSecRequestBlocking:
     - weblog_declaration:
         "*": v1.50.0-rc.1
-        envoy, apim: v1.72.0
+        apim: v2.7.1
+        envoy: v1.72.0
         haproxy: v2.4.0
   tests/appsec/test_runtime_activation.py::Test_RuntimeActivation: v1.69.0
   tests/appsec/test_runtime_activation.py::Test_RuntimeActivationCapabilities: v1.69.0
@@ -856,7 +862,8 @@ manifest:
   tests/appsec/test_trace_tagging.py::Test_TraceTaggingRulesRcCapability: v2.1.0-dev
   tests/appsec/test_traces.py:
     - weblog_declaration:
-        envoy, apim: v1.72.0
+        apim: v2.7.1
+        envoy: v1.72.0
         haproxy: v2.4.0
   tests/appsec/test_traces.py::Test_AppSecEventSpanTags:
     - weblog_declaration:
@@ -889,7 +896,8 @@ manifest:
     - weblog_declaration:
         "*": v1.36.0
         gin: v1.37.0
-        envoy, apim: v1.72.0
+        apim: v2.7.1
+        envoy: v1.72.0
         haproxy: v2.4.0
   tests/appsec/waf/test_addresses.py::Test_BodyJson: v1.37.0
   tests/appsec/waf/test_addresses.py::Test_BodyJson::test_json_array:
@@ -1738,7 +1746,8 @@ manifest:
   tests/test_config_consistency.py::Test_Config_UnifiedServiceTagging_CustomService:
     - weblog_declaration:
         "*": v1.67.0
-        envoy, apim: v1.72.0
+        apim: v2.7.1
+        envoy: v1.72.0
         haproxy: v2.4.0
   tests/test_config_consistency.py::Test_Config_UnifiedServiceTagging_CustomService::test_specified_service_name:
     - weblog_declaration:
@@ -1907,7 +1916,8 @@ manifest:
   tests/test_sampling_rates.py::Test_SamplingRates: v1.36.2
   tests/test_scrubbing.py:
     - weblog_declaration:
-        envoy, apim: v1.72.0
+        apim: v2.7.1
+        envoy: v1.72.0
         haproxy: v2.4.0
   tests/test_scrubbing.py::Test_UrlField::test_main:
     - weblog_declaration:
@@ -1916,12 +1926,14 @@ manifest:
   tests/test_scrubbing.py::Test_UrlQuery: v1.40.0
   tests/test_semantic_conventions.py:
     - weblog_declaration:
-        envoy, apim: v1.72.0
+        apim: v2.7.1
+        envoy: v1.72.0
         haproxy: v2.4.0
   tests/test_semantic_conventions.py::Test_Meta: v1.45.0
   tests/test_semantic_conventions.py::Test_Meta::test_meta_component_tag:
     - weblog_declaration:
-        envoy, apim: v2.3.0
+        apim: v2.7.1
+        envoy: v2.3.0
   tests/test_semantic_conventions.py::Test_Meta::test_meta_http_url:
     - component_version: <1.69.0-dev
       declaration: bug (APMRP-360)
@@ -1945,7 +1957,8 @@ manifest:
       declaration: missing_feature (span got an event property set as an empty list)
   tests/test_standard_tags.py:
     - weblog_declaration:
-        envoy, apim: v1.72.0
+        apim: v2.7.1
+        envoy: v1.72.0
         haproxy: v2.4.0
   tests/test_standard_tags.py::Test_StandardTagsClientIp: v1.46.0
   tests/test_standard_tags.py::Test_StandardTagsClientIp::test_client_ip_vendor:
--- a/manifests/golang.yml
+++ b/manifests/golang.yml
@@ -94,22 +94,18 @@ manifest:
   tests/appsec/api_security_testing/test_headers_collection.py::Test_SecurityTestingHeaders:
     - weblog_declaration:
         "*": v2.10.0-dev
-        envoy, apim: irrelevant (not applicable to proxies)
-        haproxy: irrelevant (not applicable to proxies)
+        envoy, apim, haproxy: irrelevant (not applicable to proxies)
   tests/appsec/api_security_testing/test_normalized_route.py::Test_NormalizedRoute:
     - weblog_declaration:
-        envoy, apim: irrelevant (not applicable to proxies)
-        haproxy: irrelevant (not applicable to proxies)
+        envoy, apim, haproxy: irrelevant (not applicable to proxies)
         "*": missing_feature
   tests/appsec/api_security_testing/test_normalized_route.py::Test_NormalizedRouteMultiParamsInSegment:
     - weblog_declaration:
-        envoy, apim: irrelevant (not applicable to proxies)
-        haproxy: irrelevant (not applicable to proxies)
+        envoy, apim, haproxy: irrelevant (not applicable to proxies)
         "*": missing_feature
   tests/appsec/api_security_testing/test_normalized_route.py::Test_NormalizedRouteOptionalParams:
     - weblog_declaration:
-        envoy, apim: irrelevant (not applicable to proxies)
-        haproxy: irrelevant (not applicable to proxies)
+        envoy, apim, haproxy: irrelevant (not applicable to proxies)
         "*": missing_feature
   tests/appsec/iast/sink/test_code_injection.py::TestCodeInjection: missing_feature
   tests/appsec/iast/sink/test_code_injection.py::TestCodeInjection_ExtendedLocation: missing_feature
@@ -395,8 +391,7 @@ manifest:
   tests/appsec/test_agentic_onboarding.py::Test_AppsecAgenticOnboarding:
     - weblog_declaration:
         "*": v2.11.0-dev
-        envoy, apim: irrelevant (not applicable to proxies)
-        haproxy: irrelevant (not applicable to proxies)
+        envoy, apim, haproxy: irrelevant (not applicable to proxies)
   tests/appsec/test_alpha.py:
     - weblog_declaration:
         apim: v2.7.1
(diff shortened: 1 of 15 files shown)


tests/external_processing/test_apim_callout.py
```diff
@@ -0,0 +1,166 @@
...
+            f"inline body delivery changed the trace: deferred spans {default_spans}, inline spans {inline_spans}"
+        )
+
+    def setup_inline_body_mode_closes_request_state(self) -> None:
+        self.r = weblog.post(
+            STATE_CLOSURE_PROBE_PATH,
+            json={"body": "inline state"},
+            headers={"X-Datadog-Apim-Body-Mode": "inline"},
+        )
+        time.sleep(31)
```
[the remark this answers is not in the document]
eliottness: Fixed in `ec42fca26`.

`pr-review.mdc:34` is explicit and this violated it. The wait turned out to be unnecessary rather than needing a timeout parameter: `conftest.py:422-489` runs every `setup_` in a loop and only calls `post_setup(session)` afterwards, which stops the containers and collects their logs. So by the time the assertion reads the log, far more than the callout's 30s state TTL has already elapsed.

Also scoped the assertion to the probe's own request-id, which `_assert_inline_probe_phases` was already computing and discarding. Matching the bare warning text would have failed on any unrelated request that legitimately orphaned state and blamed inline mode for it.

Side benefit: the apim jobs went from 4m05s/4m07s to 2m52s/3m49s.

tests/external_processing/test_apim_callout.py
```diff
@@ -0,0 +1,166 @@
...
+def _trace_structure(request: HttpResponse) -> list[tuple[str, str, str]]:
+    interfaces.library.assert_trace_exists(request=request)
+    assert _span_structure(interfaces.library.get_root_span(request=request)) == APIM_SPAN
+    interfaces.library.validate_all_spans(request=request, validator=_assert_apim_span)
+    return sorted(_span_structure(span) for _, _, span in interfaces.library.get_spans(request=request))
+
+
+@irrelevant(context.weblog_variant != "apim")
+@features.go_proxies
+class Test_ApimCallout:
```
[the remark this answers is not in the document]
eliottness: Not applying this, to stay consistent with the surrounding directory.

`tests/external_processing/test_apm.py::Test_GoProxies_Tracing` covers envoy/haproxy/apim in the same directory with the same `@features.go_proxies`, and carries `@irrelevant` with no scenario decorator. No test class in `tests/external_processing/` declares one, so adding it here alone would make this the only outlier.

Happy to add it if the convention should apply directory-wide, but that seems like a separate change covering both files rather than only the new one.

tests/external_processing/test_apim_callout.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+def _trace_structure(request: HttpResponse) -> list[tuple[str, str, str]]:
+    interfaces.library.assert_trace_exists(request=request)
+    assert _span_structure(interfaces.library.get_root_span(request=request)) == APIM_SPAN
+    interfaces.library.validate_all_spans(request=request, validator=_assert_apim_span)
+    return sorted(_span_structure(span) for _, _, span in interfaces.library.get_spans(request=request))
+
+
+@irrelevant(context.weblog_variant != "apim")
```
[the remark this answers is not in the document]
eliottness: Keeping the decorator, but flagging the tension rather than leaving it implicit.

`test-activation.mdc:8` does say to prefer manifests, and this condition is weblog-only so a manifest could express it. Two things pointed the other way:

1. The sibling `test_apm.py::Test_GoProxies_Tracing` gates the same way with no manifest rows, so a manifest-only form here would be inconsistent with the class this was modelled on.
2. This PR did originally carry `"*": irrelevant (only applies to apim)` + `apim: v2.7.1` rows for all four tests. They were removed during review as duplicating what the decorator already produces, and because `apim: v2.7.1` is unreachable — the reported library version comes from the apim-callout image and the oldest published tag is `v2.7.1`.

@DataDog/system-tests-core owns the activation tooling, so happy to restore the manifest rows and drop the decorator if you prefer manifest-only. The 16 lines are a one-line revert away.

tests/external_processing/test_apim_callout.py
```diff
@@ -0,0 +1,166 @@
...
+DEFAULT_PROBE_PATH = "/?probe=default-deferred-body"
+INLINE_PROBE_PATH = "/?probe=inline-two-call"
+TRACE_DEFAULT_PROBE_PATH = "/?probe=trace-default"
+TRACE_INLINE_PROBE_PATH = "/?probe=trace-inline"
+STATE_CLOSURE_PROBE_PATH = "/?probe=inline-state-closure"
+
+
+def _container_stderr(container_name: str) -> str:
+    log_path = Path(context.scenario.host_log_folder) / "docker" / container_name / "stderr.log"
+    return log_path.read_text(encoding="utf-8")
```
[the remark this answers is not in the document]
eliottness: This one does not reproduce — `stderr.log` does exist when these test methods run.

The premise is right that `collect_logs()` is the writer and `remove()` its only caller, but the ordering is the other way round: removal happens between the setups and the tests, not after the scenario.

`conftest.py:422` `pytest_collection_finish` iterates every selected item calling its `setup_` method (`:467-475`), and only after that loop calls `context.scenario.post_setup(session)` (`:489`). `post_setup` runs `_wait_and_stop_containers()` then `close_targets()` in a `finally` (`endtoend.py:428-435`), `close_targets()` removes every container (`endtoend.py:185-187`), and `remove()` calls `collect_logs()` first (`containers.py:455-462`), which writes `stderr.log` (`containers.py:432-443`). `pytest_runtest_call` is defined after that hook, so assertions run last.

The CI output shows this directly — the file appears twice, once per phase:

```
--------------------------------- tests setup ----------------------------------
tests/external_processing/test_apim_callout.py ....
----------------------- Wait for library interface (10s) -----------------------
tests/external_processing/test_apim_callout.py .... [ 97%]
```

Four dots in each phase, matching the four `test_` methods. That is from run `31791309961` on a GitHub-hosted runner, so no stale `logs_*` directory can be involved — the point your own caveat raises. Both complete job logs contain zero `FileNotFoundError`.

Worth noting the suggested remedy would not work here: at assertion time `post_setup` has already removed the containers, so `Container.logs()` would raise `APIError`. Reading the collected file is the only option available at that point.

I did check the adjacent risk you may have been circling — whether the helper reads the same directory the container writes. `DockerScenario.configure()` passes the scenario `host_log_folder` to every container (`endtoend.py:108-109`), and both resolve to `logs/docker/<name>/stderr.log` for DEFAULT. They could only diverge under an external `SYSTEM_TESTS_HOST_PROJECT_DIR` override, which nothing in the repo sets.

eliottness:
@cbeauchesne can you review this PR please :pray:?

tests/external_processing/test_apim_callout.py
```diff
@@ -159,18 +160,13 @@ def setup_inline_body_mode_closes_request_state(self) -> None:
             json={"body": "inline state"},
             headers={"X-Datadog-Apim-Body-Mode": "inline"},
         )
+        # Keep the callout alive beyond its configured 2s request-state TTL. If inline response
+        # handling leaves the state cached, the live expiry sweep logs the orphan before teardown.
+        time.sleep(3)
```
cbeauchesne: By experience, this kind of stuff is very brittle, and hard to debug. Would it be possible to wait for an explicit signal with a longer timeout ? not only, it'll be faster to run (I presume that you've chose  3s to be a wide window), but you'll be able to set a higher timeout that will make any timeout hit being a real error, with an explicit error message on what was expected.

cbeauchesne (changes requested):

utils/build/docker/golang/apim/main.go
```diff
@@ -0,0 +1,361 @@
...
+	}
+	defer response.Body.Close()
+	if response.StatusCode < http.StatusOK || response.StatusCode >= http.StatusMultipleChoices {
+		err := fmt.Errorf("callout returned %s", response.Status)
+		g.logCallout(phase, requestID, requestPath, err)
+		return calloutResult{}, err
+	}
+
+	var result calloutResult
+	if err := json.NewDecoder(response.Body).Decode(&result); err != nil {
```
christophe-papazian: Fails **open**: `Decode` doesn't check for trailing data, so `{"request-id":"rid"} GARBAGE {"block":{"status":403}}` decodes "fine", the block is dropped, and the request reaches the upstream. Verified: `status=200 upstream_reached=true`.

`dec := json.NewDecoder(...)` then assert `dec.Decode(&result) == nil && dec.More() == false`.

utils/build/docker/golang/apim/main.go
```diff
@@ -0,0 +1,361 @@
...
+		headers[http.CanonicalHeaderKey(name)] = append([]string(nil), values...)
+	}
+}
+
+func writeBlock(w http.ResponseWriter, block *blockResult) error {
+	content, err := base64.StdEncoding.DecodeString(block.Content)
+	if err != nil {
+		return err
+	}
+	if block.Status < http.StatusContinue || block.Status > 999 {
```
christophe-papazian: `http.StatusContinue` admits 100–199. Go's `WriteHeader(103)` sends an informational response without committing, then the `Write` below commits `200 OK` — client sees 200 with the block body, upstream never called. Verified.

Use `http.StatusOK` as the lower bound. `main_test.go:476` can't catch this: `ResponseRecorder` doesn't model 1xx.

utils/build/docker/golang/apim/main.go
```diff
@@ -0,0 +1,361 @@
...
+		fmt.Fprintln(os.Stderr, err)
+		os.Exit(1)
+	}
+}
+
+func newGateway() *gateway {
+	return &gateway{
+		calloutURL:     calloutEndpoint,
+		upstreamURL:    upstreamEndpoint,
+		calloutClient:  &http.Client{Timeout: calloutTimeout},
```
christophe-papazian: Both clients follow redirects. A gateway must hand the 3xx back: verified a POST to a redirecting upstream comes out as `200 followed! method=GET`. Same on the callout client — a `307` lands as an accepted 200, bypassing the non-2xx check below.

Add `CheckRedirect: func(*http.Request, []*http.Request) error { return http.ErrUseLastResponse }` to both. Latent today (`http-app` never 3xxes), but it's a load-bearing proxy invariant.

utils/build/docker/golang/apim/main.go
```diff
@@ -0,0 +1,361 @@
...
+		g.writeFailClosed(w)
+		return
+	}
+	if phase3.Block != nil {
+		if err := writeBlock(w, phase3.Block); err != nil {
+			g.failClosed(w, "write-block-response-headers", err)
+		}
+		return
+	}
+	if bodyMode == bodyModeVerify {
```
christophe-papazian: The verify-mode repeat discards its result, so a `Block` returned here is silently dropped. Worth asserting it's nil rather than `_`.

manifests/golang.yml
```diff
@@ -395,11 +395,11 @@ manifest:
   tests/appsec/test_agentic_onboarding.py::Test_AppsecAgenticOnboarding:
     - weblog_declaration:
         "*": v2.11.0-dev
-        envoy: irrelevant (not applicable to proxies)
+        envoy, apim: irrelevant (not applicable to proxies)
         haproxy: irrelevant (not applicable to proxies)
   tests/appsec/test_alpha.py:
     - weblog_declaration:
-        envoy: v1.72.0
+        envoy, apim: v1.72.0
```
christophe-papazian: These floors are Envoy's, not APIM's. From ghcr the earliest published images are:

| image | earliest tag | declared |
|---|---|---|
| `service-extensions-callout` | `v1.72.0` | `envoy: v1.72.0` ✅ |
| `haproxy-spoa` | `v2.4.0` | `haproxy: v2.4.0` ✅ |
| `apim-callout` | **`v2.7.1`** | `apim: v1.72.0` ❌ |

So the convention is "version the integration first shipped". 13 rows are affected (12× `v1.72.0`, plus `v2.3.0` at line 1924). Against a v2.5.0 library these enable tests for an integration that doesn't exist, so they error instead of reporting `missing_feature`.

manifests/golang.yml
```diff
@@ -498,7 +498,7 @@ manifest:
   tests/appsec/test_asm_standalone.py::Test_AppSecStandalone_APMDisabledMarker: missing_feature
   tests/appsec/test_asm_standalone.py::Test_AppSecStandalone_NotEnabled:
     - weblog_declaration:
-        envoy: missing_feature
+        envoy, apim: missing_feature
```
christophe-papazian: Most widened rows now carry the same value twice like this — `envoy, apim, haproxy: missing_feature` collapses them. For the version-gated ones, `docs/edit/manifest.md:200` documents `refs:` anchors as the idiom (the comma form has no prior inline use in `manifests/`). Would cut the 179-line diff a lot. Purely cosmetic.

(Checked `manifest_editor.py:150` for index-arithmetic breakage from comma keys — safe, one raw key still yields one condition.)

utils/_context/containers.py
```diff
@@ -1776,3 +1776,98 @@ def __init__(
...
+            "DD_AGENT_HOST": "proxy",
+            "DD_TRACE_AGENT_PORT": str(ProxyPorts.weblog),
+            # not inherited: this lives in ExternalProcessingContainer, not in GoProcessorContainer
+            "DD_APPSEC_WAF_TIMEOUT": "1s",
+            # required, do not remove: the callout's own getDefaultEnvVars() defaults this to
+            # "false" whenever the variable is empty, which rate-limits ordinary traces to one per
+            # minute and makes APM assertions nondeterministic
+            "DD_APM_TRACING_ENABLED": "true",
+            # Expose the explicit cache-miss signal used by the state-closure test.
+            "DD_TRACE_DEBUG": "true",
```
christophe-papazian: This turns on debug logging for **every** test running against `apim`, to expose one log line for one assertion. Combined with `DD_APM_TRACING_ENABLED` above, `apim` isn't at library-config parity with `envoy`, which weakens the "transcribe envoy's rows verbatim" argument for the manifest.

Could it be scoped to just that test's container env instead?

tests/external_processing/test_apim_callout.py
```diff
@@ -0,0 +1,187 @@
...
+    def test_inline_body_mode_closes_request_state(self) -> None:
+        assert self.r.status_code == 200, f"inline probe returned {self.r.status_code}, expected 200"
+        request_id = _assert_state_verification_probe_phases(STATE_CLOSURE_PROBE_PATH)
+        # The verification mode repeats <ResponseHeaders> with the same request-id immediately
+        # after the real response phase. The callout logs this positive cache-miss signal only when
+        # the first response phase already deleted the request state.
+        cache_misses = [
+            line
+            for line in _container_stderr("apim-callout").splitlines()
+            if "no request state found" in line and request_id in line
```
christophe-papazian: Greps a non-contractual dd-trace-go debug string with no version gate — silently breaks if upstream rewords it. Is there a contractual signal (tag, metric, span) that state was closed?

tests/test_the_test/test_get_image_list.py
```diff
@@ -0,0 +1,114 @@
...
+
+@contextmanager
+def _pointer_file(path: Path, *, present: bool) -> Iterator[None]:
+    """Force the presence or the absence of a binaries/ image pointer file.
+
+    A pointer file already sitting there is a local build artifact: move it aside and put it back,
+    so both states can be tested whatever the checkout looks like.
+    """
+
+    backup = path.parent / f"{path.name}.test_the_test_backup"
```
christophe-papazian: Single fixed backup name for both parametrized states. Safe today (`run.sh:423` forces `-n 1` outside PARAMETRIC), but a manual `pytest -n` could lose a real `binaries/` pointer. `monkeypatch` on the path, or a `tmp_path` pointer, would avoid touching the worktree at all.

christophe-papazian (commented):
Reviewed the shim, the Python wiring and the manifest widening. Nice work — the topology is coherent, the shim unit tests are thorough (74% stmt coverage), `format.sh --check` and `TEST_THE_TEST` (398/398) pass, and I verified the widening leaves `envoy`/`haproxy` parsed declarations byte-identical.

Non-blocking, but three shim defects are worth fixing before merge — I confirmed all three empirically against the real `gateway` type:

```
1xx block   -> final status=200 body="blocked"        (block asked for 103)
redirect    -> status=200 body="followed! method=GET" (should be 302)
bad JSON    -> status=200 upstream_reached=true       (should be fail-closed 502)
```

The third is the important one: it fails *open*, which contradicts the PR's central design claim.

Also: the 13 version-gated `apim` rows inherit Envoy's floors, and the branch is 14 commits behind `main` (no conflict yet, but `manifests/golang.yml` is hot on both sides).

utils/build/docker/golang/apim/main.go
eliottness: Fixed in `[golang] Fail closed on trailing data in apim callout response` (cbc1e14a8). The decode now uses a `json.Decoder` and rejects trailing data via `decoder.More()`, so a valid head followed by a trailing block value fails closed instead of dropping the block. Pinned by `TestFailClosedOnTrailingCalloutData`.

https://github.com/DataDog/system-tests/commit/cbc1e14a8

utils/build/docker/golang/apim/main.go
eliottness: Fixed in `[golang] Reject 1xx block status so it can't commit an implicit 200` (fc29ff20c). The lower bound is now `http.StatusOK`, so a 1xx block is rejected at the guard before any `WriteHeader`. Added an `informational-status` (103) case to `TestFailClosedOnMalformedBlock` — since the fix short-circuits before any write, the ResponseRecorder assertion is sufficient for the fixed path.

https://github.com/DataDog/system-tests/commit/fc29ff20c

utils/build/docker/golang/apim/main.go
eliottness: Fixed in `[golang] Stop apim clients following redirects, forward 3xx verbatim` (0a7c5f8c5). Both clients set `CheckRedirect` returning `http.ErrUseLastResponse`. Two real-server tests: `TestUpstreamRedirectIsForwardedNotFollowed` (a 302 from the upstream is forwarded verbatim, upstream hit exactly once) and `TestCalloutRedirectFailsClosed` (a 302 from the callout hits the non-2xx guard → 502, not an accepted 200).

https://github.com/DataDog/system-tests/commit/0a7c5f8c5

utils/build/docker/golang/apim/main.go
eliottness: Resolved by removing the branch in `[golang] Drop the debug-log-based apim state-closure test` (12ee02766). The discarded `_` was in the `inline-verify-state` repeat callout, which existed only to observe request-state closure — a property with no contractual signal in dd-trace-go (see the containers.py and test_apim_callout.py threads). The whole mechanism was dropped rather than patched.

https://github.com/DataDog/system-tests/commit/12ee02766

manifests/golang.yml
eliottness: Fixed in `[golang] Floor apim manifest rows at v2.7.1, the first shipped version` (37e3dcde9). apim is split onto its own key at `v2.7.1` (the earliest published `apim-callout` tag) across all 13 rows — the 12 that were `v1.72.0` plus the `v2.3.0` at `Test_Meta::test_meta_component_tag`. envoy and haproxy floors are unchanged.

https://github.com/DataDog/system-tests/commit/37e3dcde9

manifests/golang.yml
eliottness: Done in `[golang] Collapse duplicated proxy manifest rows into a single key` (7d77fa5de). I used the comma form `envoy, apim, haproxy: <value>` rather than `refs:` anchors — golang.yml has no `refs:` section and the test-activation guidance says not to introduce one without sign-off. Collapsed 68 same-value adjacent pairs (−136/+68 lines); left the `.../param ... (from the test_blocking test)` value-mismatch row and the rows followed by a non-haproxy key untouched. The version rows from the other thread no longer share a key, so they're excluded.

https://github.com/DataDog/system-tests/commit/7d77fa5de

utils/_context/containers.py
eliottness: Fixed in `[golang] Drop the debug-log-based apim state-closure test` (12ee02766). `DD_TRACE_DEBUG` existed only to surface the cache-miss log for the state-closure assertion. That property has no contractual signal (verified against dd-trace-go: the callout response contract, span tags and metrics carry nothing about request-state lifecycle — it's only a DEBUG log line), and one container per scenario means the flag can't be scoped per-test, so the whole state-closure test is removed along with the flag.

`DD_APM_TRACING_ENABLED` stays, and I don't think it's a real parity break: it compensates for the apim-callout image's own `getDefaultEnvVars()`, which defaults it to `false` when unset and rate-limits traces to 1/min (envoy's service-extensions-callout has no such default). It's there to make the observable trace behaviour match envoy's, so the "transcribe envoy's rows" argument still holds.

https://github.com/DataDog/system-tests/commit/12ee02766

tests/external_processing/test_apim_callout.py
eliottness: Removed in `[golang] Drop the debug-log-based apim state-closure test` (12ee02766). You're right that grepping a non-contractual debug string is fragile. I checked dd-trace-go for a contractual alternative and there isn't one: state closure is observable only via the DEBUG `no request state found` log; the callout response contract (`request-id`/`propagate-headers`/`allowed-body-size`/`block`), the span tags, and the metrics carry nothing about request-state lifecycle. Since it can't be made contractual, and asserting an internal via a debug log runs against the black-box model, the whole state-closure test (plus its `inline-verify-state` gateway mode and `DD_TRACE_DEBUG`) is removed.

https://github.com/DataDog/system-tests/commit/12ee02766

tests/test_the_test/test_get_image_list.py
eliottness: Fixed in `[golang] Run get-image-list pointer test against a tmp binaries dir` (a28ba2c87). The test now runs `get-image-list.py` with `cwd` set to a `tmp_path` holding the fake `binaries/` pointer and `PYTHONPATH` pointing at the repo root, so the container constructor resolves the pointer from tmp and the worktree is never touched — both states can run under `pytest -n` with no shared backup to race on. This let me drop the `_pointer_file` backup/restore fixture and its recovery test entirely.

https://github.com/DataDog/system-tests/commit/a28ba2c87

tests/external_processing/test_apim_callout.py
eliottness: Following up on the brittleness: rather than harden the explicit-signal wait, the whole state-closure check is removed in `[golang] Drop the debug-log-based apim state-closure test` (12ee02766). The signal it waited on (`no request state found`) is a non-contractual DEBUG log with no contractual equivalent in dd-trace-go, and asserting an internal cache lifecycle via a log line runs against system-tests' black-box model — so the brittle wait is gone entirely rather than made more robust. Thanks for pushing on this.

https://github.com/DataDog/system-tests/commit/12ee02766

christophe-papazian (approved):

utils/_context/weblog_infrastructure.py
```diff
@@ -44,12 +46,13 @@ class EndToEndWeblogInfra(WeblogInfra):
...
-    or an HAProxy SPOA agent. It intercepts HTTP traffic from the proxy runtime to apply
-    AppSec rules and emit traces. This is the "weblog" from the library's point of view."""
+    """the Datadog library under test, running as a proxy-side callout: an Envoy external
+    processor, an HAProxy SPOA agent or an Azure APIM callout. It intercepts HTTP traffic from the
+    proxy runtime to apply AppSec rules and emit traces. This is the "weblog" from the library's
+    point of view."""
 
-    _proxy_runtime_container: EnvoyContainer | HAProxyContainer
-    """the reverse proxy (Envoy or HAProxy) that sits in front of the
+    _proxy_runtime_container: TestedContainer
```
cbeauchesne: Is it possible to keep `EnvoyContainer | HAProxyContainer |  ApimGatewayContainer` ?

cbeauchesne (approved):
Just a small non-blocking comment

utils/_context/weblog_infrastructure.py
eliottness: Done in `[golang] Narrow _proxy_runtime_container to the concrete proxy types` (fcb66d0c6) — the annotation is back to `EnvoyContainer | HAProxyContainer | ApimGatewayContainer`. The three classes are already imported and the only assignments are exactly those types, so mypy stays happy.

https://github.com/DataDog/system-tests/commit/fcb66d0c6

PR #7518, opened 2026-08-14
test(otel): port otel tracestate sampling rate scenarios to parametric

MilanGarnier:
## Summary

- Move the multi-rate tracestate golden-vector matrix and precision-boundary coverage to `tests/parametric/test_otel_tracestate_sampling.py`; retain the rate-0.5 end-to-end sampling regression.
- Keep the direct end-to-end forwarding, force-keep, and malformed-input coverage; run malformed-input checks under the general sampling scenario.
- Share tracestate thresholds, golden vectors, and forwarding values through `tests/otel/utils.py`; the OTLP end-to-end cases consume those values without changing their assertions.
- Remove the dedicated `otel_sampling_rate_*` scenarios.
- Add matching parametric manifest declarations: Go and Ruby remain version-gated, while unsupported tracer targets remain `missing_feature`.

## Validation

- `./format.sh`: mypy, ruff, YAML, manifest parsing, and AI Guard checks pass. ShellCheck is blocked by the unrelated directory error at `utils/build/docker/internal_server/app.sh`.
- `git diff --check`
- Targeted Python syntax compilation

No broad test suite was run.


MilanGarnier:
@codex review

tests/parametric/test_otel_tracestate_sampling.py
```diff
@@ -0,0 +1,618 @@
...
+            )
+
+        traceparent, tracestate = get_tracecontext(headers)
+        assert _parse_ot(tracestate) == {"rv": expected_rv, "th": expected_th}
+        assert (traceparent.trace_flags == "01") is expected_sampled
+
+
+@features.otel_api
+@scenarios.parametric
+class Test_OtlpTracestateSampling:
```
vpellan: I think this should be in a separate file (or if parametric tests for OTLP already exists it should be there) to keep the structure we had previously similar

tests/parametric/test_otel_tracestate_sampling.py
```diff
@@ -0,0 +1,618 @@
...
+        assert _parse_ot(rv_only_tracestate) == {"rv": "1234567890abcd"}
+
+        dropped_traceparent, dropped_tracestate = get_tracecontext(dropped_headers)
+        assert dropped_traceparent.trace_flags == "01"
+        assert _parse_ot(dropped_tracestate) == {"rv": "65cd67504a538e"}
+
+        traces = test_agent.wait_for_num_traces(3)
+        for trace in traces:
+            (span,) = trace
+            assert span["metrics"].get(SAMPLING_PRIORITY_KEY) == SamplingPriority.USER_KEEP
```
vpellan: Minor: This works but I think it would be clearer to wait for a single trace each time, which means it'd be one test per input. It's up to your preference

tests/parametric/test_otel_tracestate_sampling.py
```diff
@@ -0,0 +1,618 @@
...
+
+        traces = test_agent.wait_for_num_traces(len(vectors))
+        for trace_id, _, expected_sampled in vectors:
+            (span,) = find_trace(traces, trace_id)
+            sampling_priority = span["metrics"].get(SAMPLING_PRIORITY_KEY)
+            assert sampling_priority is not None
+            assert (sampling_priority in (SamplingPriority.AUTO_KEEP, SamplingPriority.USER_KEEP)) is expected_sampled
+
+    @pytest.mark.parametrize("library_env", [_library_env(0.1)])
+    def test_forwards_and_sanitizes_inbound_ot(self, test_library: APMLibrary) -> None:
```
vpellan: This test merges several ones from the previous implem, I think it was clearer before (1 test for A2, 1 for A6, 1 for A2b, 1 for A3).
The test that was merged in A2 (unknown fields in ot) is already tested in `test_forwards_unknown_inbound_ot_subkeys` , so I'd remove it too.

vpellan (approved):
LGTM

tests/parametric/test_otel_tracestate_sampling.py
MilanGarnier: I moved it in a new `tests/parametric/test_otlp_trace_export.py` file e168e354f

cbeauchesne (approved):

tests/otel/test_tracing_otlp.py
```diff
@@ -211,9 +216,9 @@ def _parse_ot_from_trace_state(trace_state: str) -> dict[str, str]:
 
 # Trace ID, rv and th match the RFC's own verified worked example (rate 0.1, trace ID 0xfff972474538efff),
 # also used in tests/test_otel_tracestate_sampling.py. rv only depends on the trace ID, not the sample rate.
-OTLP_TRACE_ID = 18444899399302180863
-OTLP_RV = "ef284ace7a91e1"
-OTLP_TH = "e6666666666668"
+OTLP_TRACE_ID = _FORWARD_TRACE_ID
+OTLP_RV = _FORWARD_RV
+OTLP_TH = _FORWARD_TH
```
vpellan: We're doing 2 renaming, I think we could use `FORWARD_*` directly, same in tests/test_otel_tracestate_sampling.py

MilanGarnier: Done in 53116d845: both modules now use the shared FORWARD_* values directly.

PR #7522, opened 2026-08-14
test(cpp): enable otel tracestate sampling

MilanGarnier:
Enables C++ OpenTelemetry tracestate sampling coverage. Stacked on #7518.

### DIFF
--- a/manifests/cpp.yml
+++ b/manifests/cpp.yml
@@ -237,7 +237,7 @@ manifest:
   tests/parametric/test_otel_span_with_baggage.py::Test_Otel_Span_With_Baggage: missing_feature
   tests/parametric/test_otel_tracer.py::Test_Otel_Tracer::test_otel_force_flush: irrelevant (library does not implement OpenTelemetry)
   tests/parametric/test_otel_tracer.py::Test_Otel_Tracer::test_otel_simple_trace: irrelevant (library does not implement OpenTelemetry)
-  tests/parametric/test_otel_tracestate_sampling.py::Test_OtelTracestateSampling: missing_feature (APMAPI-2171)
+  tests/parametric/test_otel_tracestate_sampling.py::Test_OtelTracestateSampling: ">2.2.0"
   tests/parametric/test_otlp_trace_metrics.py: '>=2.1.1'  # TODO: a lower version might be supported
   tests/parametric/test_otlp_trace_metrics.py::Test_FR01_Enablement_Configuration::test_fr01_1_enabled_explicit: missing_feature
   tests/parametric/test_otlp_trace_metrics.py::Test_FR01_Enablement_Configuration::test_fr01_3_enabled_by_default: missing_feature


manifests/cpp.yml
```diff
@@ -235,6 +235,8 @@ manifest:
   tests/parametric/test_otel_span_with_baggage.py::Test_Otel_Span_With_Baggage: missing_feature
   tests/parametric/test_otel_tracer.py::Test_Otel_Tracer::test_otel_force_flush: irrelevant (library does not implement OpenTelemetry)
   tests/parametric/test_otel_tracer.py::Test_Otel_Tracer::test_otel_simple_trace: irrelevant (library does not implement OpenTelemetry)
+  tests/parametric/test_otel_tracestate_sampling.py::Test_OtelTracestateSampling: ">2.1.2"
```
xlamorlette-datadog: This should rather be `>=2.2.0` since [dd-trace-cpp v2.2.0 was delivered 2 days ago](https://github.com/DataDog/dd-trace-cpp/releases/tag/v2.2.0).
See https://github.com/DataDog/system-tests/pull/7517.

xlamorlette-datadog (approved):

manifests/cpp.yml
MilanGarnier: oh thanks I hadn't seen that yet !

dougqh:
🤖 *Comment from Claude (via Claude Code), reviewing this PR at a user's request.*

I think `test_precision_boundary_decisions` in `tests/parametric/test_otel_tracestate_sampling.py` has incorrect hardcoded expected values for both parametrized cases.

I reimplemented the sampling formula documented in `tests/test_otel_tracestate_sampling.py`'s module docstring:
```
h = (trace_id * 1111111111111111111) mod 2**64
rv = (~h & (2**64-1)) >> 8
th = round((1 - rate) * 2**56)
sampled = rv >= th
```
and verified it reproduces the existing, already-verified `SAMPLING_RATE_0_1` fixtures exactly (rv matches to the hex digit, and the sampled/dropped decision derived from `rv >= th` matches too, for trace IDs 1, 10, 100, 1000, 83).

Applying that same validated formula to the two new boundary vectors:

- `rate-0.1-kept` (trace_id `0x03A93EE8B1999F00`): I get `rv = e6666666666666`, `sampled = False` — but the test asserts `rv = e6666666666668` (equal to `th`) and `expected_sampled = True`.
- `rate-0.05-dropped` (trace_id `5401449561355763072`): I get `rv = f3333333333331`, `sampled = True` — but the test asserts `rv = f333333333332f` and `expected_sampled = False`.

In both cases the `rv` value and the sampled/dropped direction come out flipped relative to what the test expects. This should fail against a tracer that implements the algorithm correctly (or could mask a bug in one that doesn't). Might be worth regenerating these two vectors — unlike the pre-existing `SAMPLING_RATE_*` tables, these weren't cross-checked against the verified `sampling_rates.csv`.

MilanGarnier:
> Applying that same validated formula to the two new boundary vectors:
> 
> * `rate-0.1-kept` (trace_id `0x03A93EE8B1999F00`): I get `rv = e6666666666666`, `sampled = False` — but the test asserts `rv = e6666666666668` (equal to `th`) and `expected_sampled = True`.
> * `rate-0.05-dropped` (trace_id `5401449561355763072`): I get `rv = f3333333333331`, `sampled = True` — but the test asserts `rv = f333333333332f` and `expected_sampled = False`.
> 
> In both cases the `rv` value and the sampled/dropped direction come out flipped relative to what the test expects. This should fail against a tracer that implements the algorithm correctly (or could mask a bug in one that doesn't). Might be worth regenerating these two vectors — unlike the pre-existing `SAMPLING_RATE_*` tables, these weren't cross-checked against the verified `sampling_rates.csv`.

@dougqh This is intentional. In these edge cases the imprecision comes initially from computing `th`, so rv is set to a value explicitly agreeing the sampling decision made internally using dd sampling.

nccatoni (approved):

PR #7554, opened 2026-08-20
EX-3426: add serial ID exposure event system tests (gated as missing_feature)

danyal002:
## Motivation

[EX-3426](https://datadoghq.atlassian.net/browse/EX-3426) · epic [EX-3048](https://datadoghq.atlassian.net/browse/EX-3048)

The compiler rewrites a holdout into an ordinary allocation before an SDK receives the flag configuration, so an exposure event records no holdout. The serial id of the split the subject landed in is the only link back to it. Every SDK is adding a top-level `serial_id` to the exposure event, and these tests pin that contract in one place instead of once per language.

Prior art: #7316 uses the same shape for `observeFullEvaluationData`.

## Changes

`make_ufc_fixture` and `make_exposure_ufc_fixture` take an optional `serial_id`. A default of `None` omits the key, so existing tests are unchanged. The gate is a `None` test, not a truthiness test, because serial ids start at `0` for each organization and `0` is a valid value.

Three classes in `tests/ffe/test_exposures_datadog_agent.py`:

| Class | Contract |
|---|---|
| `Test_FFE_Exposure_Serial_Id` | the exact value reaches the event, and the key is absent, not `null`, when the split has none |
| `Test_FFE_Exposure_Caching_Serial_Id_Appears` | a serial id that appears on a refresh reports a new exposure, with the allocation and variant unchanged |
| `Test_FFE_Exposure_Caching_Serial_Id_Cycle` | `340132 → 340133 → 340132` reports three exposures |

The present and absent cases are one test, not two. An SDK that never sends the field satisfies the absent case on its own, so the two only discriminate together. Two separate tests report `XPASS` on every SDK that has not implemented the field.

The two caching classes follow `Test_FFE_Exposure_Caching_Allocation_Cycle`. They cover a fault that three SDKs had: a cache that stores `(allocation_key, variant)` hides a serial id that appears or changes. This is a usual event, not a rare one, because `ffe.ufc.vac_miss` is more than zero at all times in production and those misses resolve on a later configuration refresh.

Every SDK is declared `missing_feature`. Each one enables the tests when it ships the field.

## Validation

Run against `golang`, weblog `net-http`:

```
released dd-trace-go            3 xfailed
dd-trace-go DataDog/dd-trace-go#5214   3 passed in 52.91s
```

The second run used a local checkout in `binaries/dd-trace-go`. `./format.sh --check` is clean.

Only `tests/` and `manifests/` are modified.

[EX-3426]: https://datadoghq.atlassian.net/browse/EX-3426?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[EX-3048]: https://datadoghq.atlassian.net/browse/EX-3048?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/manifests/dotnet.yml
+++ b/manifests/dotnet.yml
@@ -752,6 +752,9 @@ manifest:
   tests/ffe/test_exposure_egress.py::Test_FFE_Exposure_Egress_Agentless_Direct: missing_feature (Not yet implemented)
   tests/ffe/test_exposure_egress.py::Test_FFE_Exposure_Egress_Agentless_Sidecar: missing_feature (Not yet implemented)
   tests/ffe/test_exposures_datadog_agent.py: v3.36.0
+  tests/ffe/test_exposures_datadog_agent.py::Test_FFE_Exposure_Caching_Serial_Id_Appears: missing_feature (EX-3416)
+  tests/ffe/test_exposures_datadog_agent.py::Test_FFE_Exposure_Caching_Serial_Id_Cycle: missing_feature (EX-3416)
+  tests/ffe/test_exposures_datadog_agent.py::Test_FFE_Exposure_Serial_Id: missing_feature (EX-3416)
   tests/ffe/test_flag_eval_evp.py: missing_feature (FFL-2446)
   tests/ffe/test_flag_eval_metrics.py: v3.44.0
   tests/integration_frameworks/llm/anthropic/test_anthropic_llmobs.py::TestAnthropicLlmObsMessages::test_create_error: bug (MLOB-1234)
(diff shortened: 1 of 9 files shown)


danyal002:
Closed automatically when the head branch was renamed. Replaced by #7556, same commit (844d465).

PR #7556, opened 2026-08-20
EX-3426: add serial ID exposure event system tests (gated as missing_feature)

danyal002:
## Motivation

[EX-3426](https://datadoghq.atlassian.net/browse/EX-3426) · epic [EX-3048](https://datadoghq.atlassian.net/browse/EX-3048)

The compiler rewrites a holdout into an ordinary allocation before an SDK receives the flag configuration, so an exposure event records no holdout. The serial id of the split the subject landed in is the only link back to it. Every SDK is adding a top-level `serial_id` to the exposure event, and these tests pin that contract in one place instead of once per language.

Prior art: #7316 uses the same shape for `observeFullEvaluationData`.

## Changes

`make_ufc_fixture` and `make_exposure_ufc_fixture` take an optional `serial_id`. A default of `None` omits the key, so existing tests are unchanged. The gate is a `None` test, not a truthiness test, because serial ids start at `0` for each organization and `0` is a valid value.

Three classes in `tests/ffe/test_exposures_datadog_agent.py`:

| Class | Contract |
|---|---|
| `Test_FFE_Exposure_Serial_Id` | the exact value reaches the event, and the key is absent, not `null`, when the split has none |
| `Test_FFE_Exposure_Caching_Serial_Id_Appears` | a serial id that appears on a refresh reports a new exposure, with the allocation and variant unchanged |
| `Test_FFE_Exposure_Caching_Serial_Id_Cycle` | `340132 → 340133 → 340132` reports three exposures |

The present and absent cases are one test, not two. An SDK that never sends the field satisfies the absent case on its own, so the two only discriminate together. Two separate tests report `XPASS` on every SDK that has not implemented the field.

The two caching classes follow `Test_FFE_Exposure_Caching_Allocation_Cycle`. They cover a fault that three SDKs had: a cache that stores `(allocation_key, variant)` hides a serial id that appears or changes. This is a usual event, not a rare one, because `ffe.ufc.vac_miss` is more than zero at all times in production and those misses resolve on a later configuration refresh.

Every SDK is declared `missing_feature`. Each one enables the tests when it ships the field.

## Validation

Run against `golang`, weblog `net-http`:

```
released dd-trace-go            3 xfailed
dd-trace-go DataDog/dd-trace-go#5214   3 passed in 52.91s
```

The second run used a local checkout in `binaries/dd-trace-go`. `./format.sh --check` is clean.

Only `tests/` and `manifests/` are modified.

[EX-3426]: https://datadoghq.atlassian.net/browse/EX-3426?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[EX-3048]: https://datadoghq.atlassian.net/browse/EX-3048?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/ffe/test_exposures_datadog_agent.py
+++ b/tests/ffe/test_exposures_datadog_agent.py
@@ -4,7 +4,6 @@ import json
 
 from tests.ffe.utils.exposures import (
     EXPOSURES_PATH,
-    EXPOSURE_CACHE_SETTLE_SECONDS,
     EXPOSURE_WAIT_TIMEOUT_SECONDS,
     exposure_events_from_data,
 )
@@ -50,16 +49,6 @@ def wait_for_min_exposure_count(flag_key: str, expected: int, subject_id: str | 
         count = count_exposure_events(flag_key, subject_id)
 
     return count
-
-
-def settle_and_count_exposures(flag_key: str, subject_id: str | None = None) -> int:
-    """Let late payloads arrive, then count. An exact count taken as soon as the expected
-    number is reached cannot see an over-emitting cache that delivers the extra event later.
-    """
-    if not interfaces.agent.replay:
-        interfaces.agent.wait(EXPOSURE_CACHE_SETTLE_SECONDS)
-
-    return count_exposure_events(flag_key, subject_id)
 
 
 # Simple UFC fixture for testing with doLog: true
@@ -529,7 +518,7 @@ class Test_FFE_Exposure_Caching_Different_Subjects:
             wait_for_min_exposure_count(self.flag_key, 1, subject)
 
         # Count total exposure events for this flag
-        total_exposure_count = settle_and_count_exposures(self.flag_key)
+        total_exposure_count = count_exposure_events(self.flag_key)
 
         # Each unique subject should generate exactly one exposure
         assert total_exposure_count == len(self.subjects), (
@@ -632,8 +621,7 @@ class Test_FFE_Exposure_Caching_Allocation_Cycle:
         # - Exposure #1: default-allocation
         # - Exposure #2: different-allocation (allocation changed)
         # - Exposure #3: default-allocation (allocation changed back)
-        wait_for_min_exposure_count(self.flag_key, 3, self.targeting_key)
-        exposure_count = settle_and_count_exposures(self.flag_key, self.targeting_key)
+        exposure_count = wait_for_min_exposure_count(self.flag_key, 3, self.targeting_key)
 
         assert exposure_count == 3, (
             f"Expected exactly 3 exposure events for subject '{self.targeting_key}' "
@@ -728,8 +716,7 @@ class Test_FFE_Exposure_Caching_Variant_Cycle:
         # - Exposure #1: variant-a
         # - Exposure #2: variant-b (variant changed)
         # - Exposure #3: variant-a (variant changed back)
-        wait_for_min_exposure_count(self.flag_key, 3, self.targeting_key)
-        exposure_count = settle_and_count_exposures(self.flag_key, self.targeting_key)
+        exposure_count = wait_for_min_exposure_count(self.flag_key, 3, self.targeting_key)
 
         assert exposure_count == 3, (
             f"Expected exactly 3 exposure events for subject '{self.targeting_key}' "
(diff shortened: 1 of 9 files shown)


tests/ffe/test_exposures_datadog_agent.py
```diff
@@ -51,6 +52,16 @@ def wait_for_min_exposure_count(flag_key: str, expected: int, subject_id: str |
     return count
 
 
+def settle_and_count_exposures(flag_key: str, subject_id: str | None = None) -> int:
+    """Let late payloads arrive, then count. An exact count taken as soon as the expected
+    number is reached cannot see an over-emitting cache that delivers the extra event later.
+    """
+    if not interfaces.agent.replay:
+        interfaces.agent.wait(EXPOSURE_CACHE_SETTLE_SECONDS)
```
sameerank: `interfaces.agent.wait()` delegates to `time.sleep()`

Did you confirm that this is required for exact counts? Otherwise we are adding 3 seconds to every call

tests/ffe/test_exposures_datadog_agent.py
```diff
@@ -716,14 +728,219 @@ def test_ffe_exposure_caching_variant_cycle(self):
...
+                "defaultValue": "default",
+                "targetingKey": self.targeting_key,
+                "attributes": {},
+            },
+        )
+
+        # Step 2: same allocation and variant, serial id now present
+        rc.tracer_rc_state.set_config(
+            f"{RC_PATH}/{config_id}/config",
+            make_ufc_fixture(self.flag_key, "variant-a", allocation_key="default-allocation", serial_id=340132),
```
sameerank: Nit: A bit of an edge case, but it might make for a stronger test if you exercise serial_id 0 here to distinguish this valid value from its absence. An SDK that drops zero through a truthiness check or omitempty-style serialization would still pass due to these tests using nonzero ids

tests/ffe/test_exposures_datadog_agent.py
```diff
@@ -716,14 +728,219 @@ def test_ffe_exposure_caching_variant_cycle(self):
...
+        wait_for_min_exposure_count(self.flag_key, 2, self.targeting_key)
+        exposure_count = settle_and_count_exposures(self.flag_key, self.targeting_key)
+
+        assert exposure_count == 2, (
+            f"Expected exactly 2 exposure events for subject '{self.targeting_key}' "
+            f"(no serial id, then serial id 340132, with the allocation and variant unchanged), "
+            f"but found {exposure_count} events"
+        )
+
+        events = find_exposure_events(self.flag_key, self.targeting_key)
```
sameerank, line 874: Do you cover serial ID removal? So far, I see coverage for 
* absent→present
* present-A→present-B→present-A

but not present→absent

tests/ffe/test_exposures_datadog_agent.py
danyal002: hmm, serialId = 0 is a valid case though right? Isn't that the first serialId that gets assigned to a new org? I was previously curious about this and checked.

sameerank: Yeah it's valid. But SDKs could treat it as falsey/absent, and the system test could help enforce that this doesn't happen

danyal002: ah I get it. I'll set this to 0.

sameerank: But also this feedback is a nit. This is just one case, and an uncommon one at that

tests/ffe/test_exposures_datadog_agent.py
danyal002: Upon further inspection, I think Codex was wrong. Mechanically, it's argument is possible, but it won't happen because the scenario is not reachable. I'm gonna revert this change.

tests/ffe/test_exposures_datadog_agent.py
danyal002: Good catch. Adding that right now.

sameerank (approved):
🙌

PR #7609, opened 2026-08-27
Wait for the agent to resolve container tags in origin detection tests

ddmatan:
Origin detection tests fail intermittently with the trace present, on time, fully enriched — and only `_dd.tags.container` missing. The container is running fine, so it looks like a backend problem. It isn't.

That tag is attached by the **agent**, not the backend. The tracer sends only a container id (`Datadog-Container-ID`); the agent resolves it into tags through its tagger, which discovers containers asynchronously. It holds the payload up to 12s waiting for that, then sends the trace untagged and moves on. Nothing errors.

Where that comes from:
- containers-tagging RFC in `architecture` — tracer sends the id, agent calls the tagger and attaches `_dd.tags.container`; the backend's only job is rewriting it into `ddtags`
- `datadog-agent/pkg/trace/agent/agent.go` — `tagContainersTags`, written from `conf.ContainerTags(containerID)`
- `datadog-agent/pkg/trace/containertags/buffer.go` — `maxBufferDuration = 12s`, then shipped untagged
- dd-go / dd-source only ever *read* the tag, never create it

So the test measures too early: we wait for the weblog port, but never for the agent. This waits for `datadog-agent tagger-list` to know the weblog containers before sending the request we measure.

Gotchas:
- Best effort. If the tagger never catches up we send the request anyway and the assertion fails as it does today — this doesn't hide a regression, and it can't turn a test failure into a job timeout.
- Only for `origin_detection=True`, and only on the SSH path (krunvm is untouched).
- No containers running, or no docker on the VM → no wait, unchanged behavior.
- Branch name is stale. This replaces the earlier pytest-rerunfailures approach; the net diff is the last commit only (34 lines, no new dependency).

@cbeauchesne this drops the retry logic entirely. On the mocked backend — I don't think it would have caught this particular failure, since the tag never reaches the backend in the first place. Happy to talk it through.

🤖 Written with [Claude Code](https://claude.com/claude-code)


natitsechanski (approved):
shipit

cbeauchesne (changes requested):
Thanks for working on improving the reliability of these tests. However, I don’t think adding retry logic is the right direction here.

The underlying issue is that these scenarios depend on a real backend whose availability and performance we do not control. Retries may reduce the visible failure rate, but they do not address that design flaw. They would also add significant complexity to the framework and could hide genuine failures or make them slower to diagnose.

The known solution is to use a mocked backend that we control. I think our effort should go toward implementing that solution rather than building increasingly complex retry mechanisms around the current behavior (see https://datadoghq.atlassian.net/browse/APMSP-2478). We already have many bits in place, as we use a mocked backend for other scenarios. The biggest part will be the piping between SSI components, and this mock.

For those reasons, I don’t think we should move forward with this approach.

robertomonteromiguel:
> Thanks for working on improving the reliability of these tests. However, I don’t think adding retry logic is the right direction here.
> 
> The underlying issue is that these scenarios depend on a real backend whose availability and performance we do not control. Retries may reduce the visible failure rate, but they do not address that design flaw. They would also add significant complexity to the framework and could hide genuine failures or make them slower to diagnose.
> 
> The known solution is to use a mocked backend that we control. I think our effort should go toward implementing that solution rather than building increasingly complex retry mechanisms around the current behavior (see https://datadoghq.atlassian.net/browse/APMSP-2478). We already have many bits in place, as we use a mocked backend for other scenarios. The biggest part will be the piping between SSI components, and this mock.
> 
> For those reasons, I don’t think we should move forward with this approach.

In this case, the flakiness does not originate from the backend.

robertomonteromiguel (approved):

tests/auto_inject/utils.py
```diff
@@ -150,6 +152,32 @@ def _appsec_validator(self, _: str, trace_data: dict):
...
+            "TAGS=$(sudo datadog-agent tagger-list 2>/dev/null); MISSING=0; "
+            "for CID in $(sudo docker ps -q --no-trunc 2>/dev/null); do "
+            'case "$TAGS" in *"$CID"*) ;; *) MISSING=1 ;; esac; done; '
+            f'[ "$MISSING" = 0 ] && echo READY && break; sleep {delay}; '
+            "done"
+        )
+        if "READY" in self.execute_command(virtual_machine, command):
+            logger.info("Agent can tag the weblog containers")
+        else:
+            logger.warning(f"Agent still cannot tag the weblog containers after {attempts * delay}s")
```
cbeauchesne: Here, the test will fail. In that situation, we must stop, to avoid adding noise with other assertion : 



```suggestion
        command_output:str = self.execute_command(virtual_machine, command)
        assert "READY" in command_output, f"datadog-agent tagger-list failed:\n{command_output}"
        logger.info("Agent can tag the weblog containers")
```

cbeauchesne (changes requested):

tests/auto_inject/utils.py
ddmatan: I added this, but wouldn't it just fail the test logically now? so no retries at all?

cbeauchesne: The key point is debugging. If we *know* that a test wil lfail for a reason `X`, it've very important to have `X` at the bottom of the output, as it's the first place users will look. If ever it's bumped over dozain of test failure logs, it'll make the debugging harder. The key rule is : fail fast :)

cbeauchesne (approved):

PR #7617, opened 2026-08-31
Implement /security/thread_context_sharing endpoints for Java, Python and PHP

danielmercier:
## Motivation

End to end test for trace correlation between tracers and agent.

## Changes

Added the /security/thread_context_sharing endpoint for Java, Python and PHP

### DIFF
--- a/utils/_context/_scenarios/thread_context_sharing.py
+++ b/utils/_context/_scenarios/thread_context_sharing.py
@@ -36,6 +36,10 @@ class ThreadContextSharingScenario(EndToEndScenario):
     def configure(self, config: pytest.Config):
         super().configure(config)
 
+        if self.weblog_infra.library_name == "java":
+            # Should be removed once java enables this with appsec
+            self.weblog_container.environment["DD_PROFILING_ENABLED"] = "true"
+
         agent = self.agent_container
 
         agent.environment["DD_RUNTIME_SECURITY_CONFIG_ENABLED"] = "true"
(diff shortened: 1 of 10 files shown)


utils/_context/_scenarios/thread_context_sharing.py
```diff
@@ -36,6 +36,10 @@ def __init__(self, name: str, doc: str) -> None:
     def configure(self, config: pytest.Config):
         super().configure(config)
 
+        if self.weblog_infra.library_name == "java":
```
nccatoni: Would it be possible to avoid having a special case for java here?

nccatoni (approved):
Just a small change if possible, otherwise LGTM (@DataDog/system-tests-core)

utils/_context/_scenarios/thread_context_sharing.py
danielmercier: This is definitely a workaround until Java publishes trace/spans either by default or when appsec is enabled (separate tracked task). I can wait until this is done to add the java endpoint tho.

nccatoni: The issue is that we have no way to track this kind of special per-library handling, so we can’t guarantee that it will be removed once it is no longer needed. By contrast, if you disable the test in the manifest, we’ll know when it starts passing and can re-enable it.

TL;DR: If you’re confident that you’ll remember to remove the special handling and if the coverage provided by the test is important right now, then it’s fine. The trade-off is that you lose the assurance that the test will eventually run in its intended configuration once the feature is fully implemented.

danielmercier: I agree, updating the manifest and removing the workaround

PR #7619, opened 2026-08-31
Add script to build agent image from a local datadog-agent branch

ajgajg1134:
## Motivation

Testing datadog-agent trace-agent changes against system-tests today means waiting on a
datadog-agent CI image build, or running `dda inv agent.hacky-dev-image-build` -- which builds
host-native binaries + rtloader and requires a Linux devcontainer, so it doesn't work directly
on macOS and is slow either way.

Most APM/tracing system-tests scenarios only exercise the **trace-agent**, so this PR adds a
much faster path: compile just `cmd/trace-agent` from a local datadog-agent checkout/worktree
inside a Linux `golang` container, and overlay that one binary onto a published agent base
image. system-tests already supports pointing at a custom agent image via `binaries/agent-image`
(see `utils/_context/containers.py`), so the resulting image drops straight into an existing
system-tests workflow.

## What's added

- `utils/scripts/build-local-agent-image.sh`
  - Takes the datadog-agent source path and output image tag as positional args, `--agent-src`
    / `--image` flags, or env vars (`AGENT_SRC` / `IMAGE`) -- no personal paths or tags baked in.
  - Detects the Docker server arch (`docker version --format '{{.Server.Arch}}'`) and builds for
    it, so it works on both arm64 Macs (colima/Docker Desktop) and amd64 Linux.
  - Reads the Go version from the agent checkout's `.go-version` and uses the matching
    `golang:<version>` image.
  - Builds with the trace-agent's build tags (`docker containerd datadog.no_waf kubelet otlp
    netcgo podman`) and `CGO_ENABLED=1` (the zstd dependency needs cgo; a pure
    `CGO_ENABLED=0` cross-compile from macOS fails).
  - Mounts the host Go module cache and a persistent `GOCACHE` so repeat builds are fast
    (~45s warm).
  - Sets version ldflags (`pkg/version.{AgentVersion,Commit,FullCommit}`) so the built binary
    reports the same agent version as the base image -- otherwise it reports the generic
    `6.0.0`, which skews system-tests' agent-version gating.
  - Overlays the binary via a generated Dockerfile inside a small build context dir (a
    Dockerfile referenced with `-f` from outside the context, e.g. `/tmp`, fails on macOS with
    xattr permission errors).
  - Verifies the result by running `trace-agent version` in the built image.
  - Prints the exact `echo ... > binaries/agent-image` line to wire the image into
    system-tests.

- `docs/execute/binaries.md`: new subsection under `## Agent` covering when to use this (changes
  limited to `pkg/trace`/`cmd/trace-agent`), the exact commands, how to confirm the custom binary
  is running (`logs_<scenario>/docker/agent/stdout.log` TRACE lines show `/src/...` paths;
  `logs_<scenario>/docker/agent/image.json` shows the image tag), and the limitation that only
  the trace-agent binary is replaced -- for anything else, use
  `dda inv agent.hacky-dev-image-build` from a Linux devcontainer or a CI-built
  `datadog/agent-dev:<branch>` image.

## Validation

- End-to-end: the resulting image was used to run the `DEFAULT` scenario for `golang`, which
  passed green, and the agent container logs confirmed the custom binary was in use (TRACE log
  lines showed `/src/...` source paths from the container's build mount).
- This PR's script itself was re-run against a real local datadog-agent worktree with a throwaway tag, and produced:
  ```
  trace-agent 7.84.0-devel - Commit: 7e639d07d86 - Serialization version:  - Go version: go1.26.6
  ```
  confirming the devel version and commit were baked in correctly. The throwaway image was
  removed afterward, and no files under `binaries/` were committed.
- `shellcheck utils/scripts/build-local-agent-image.sh` passes with no findings.

### DIFF
--- a/docs/execute/binaries.md
+++ b/docs/execute/binaries.md
@@ -7,6 +7,38 @@ But we often want to run system tests against unmerged changes. The general appr
 
 * Add a file `agent-image` in `binaries/`. The content must be a valid docker image name containing the datadog agent, like `datadog/agent` or `datadog/agent-dev:master-py3`.
 
+### Building an agent image from a local datadog-agent branch
+
+If your datadog-agent changes are limited to `pkg/trace` or `cmd/trace-agent`, you don't need
+to wait for a datadog-agent CI image build (or run the slow, Linux-devcontainer-only
+`dda inv agent.hacky-dev-image-build`) to test them. `utils/scripts/build-local-agent-image.sh`
+compiles just the trace-agent binary from your local checkout/worktree and overlays it onto a
+published agent base image:
+
+```bash
+./utils/scripts/build-local-agent-image.sh /path/to/datadog-agent datadog/agent-dev:my-branch
+echo datadog/agent-dev:my-branch > binaries/agent-image
+./build.sh golang
+TEST_LIBRARY=golang ./run.sh DEFAULT
+```
+
+Run `./utils/scripts/build-local-agent-image.sh --help` for all options (base image, output
+image tag/path via flags or env vars, etc). The script builds the trace-agent with `CGO_ENABLED=1`
+(required by the zstd dependency, so it can't cross-compile from macOS with `CGO_ENABLED=0`)
+inside a `golang` container matching your Docker host's architecture and your agent checkout's
+`.go-version`, and sets version ldflags so the resulting binary reports the same agent version as
+the base image (otherwise system-tests' agent-version gating sees a generic `6.0.0`).
+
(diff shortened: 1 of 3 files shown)


cbeauchesne (changes requested):
Can you add an exclusion rule in https://github.com/DataDog/system-tests/blob/main/utils/scripts/libraries_and_scenarios_rules.yml for the script file (it'll save a ton of time in CI every time you need to modify it, including this present PR

cbeauchesne (approved):

PR #7625, opened 2026-09-01
Explain slow and stalled remote config applies

christophe-papazian:
APPSEC-69943

An RC apply that is not acknowledged in time currently surfaces only as `assert <UNKNOWN: 0> == <ACKNOWLEDGED: 2>`, which says nothing about why. Chasing one of these took a long time for want of information that was already being collected.

- report every apply that takes over 5s or times out, with elapsed time, expected vs last-seen `targets_version`, and which configs were unacknowledged
- log the warnings the library sent through telemetry: libraries report their own slow remote config callbacks there, and system-tests already captures it
- capture the weblog thread stacks at 10s and 20s into a stalled apply, so a thread parked on one call can be told from one that is moving. Read from outside the process (`py-spy` for python, added to their images; `jcmd` for java). Other libraries stay silent
- fix `state` being shadowed by two loop variables, which made the timeout message report fields off a config-state dict instead of the client state

Nothing new is sent to the library and nothing is logged for a healthy apply, which stays on a debug line. Every diagnostic path is exception-safe.

Verified by replaying the telemetry of a real failure, which now surfaces the cause directly:

```
[WARN] RC callback operation exceeded threshold (product:asm_features,callback_type:payload,elapsed_time:30.542)
```

Threshold note: 5s is calibrated on python (20 000 measured applies, max 1.80s). If another library routinely acknowledges more slowly, `SLOW_APPLY_THRESHOLD` is a single constant to raise.


utils/_remote_config.py
```diff
@@ -46,6 +48,153 @@ def from_json(d: dict) -> "RemoteConfigStateResults":
...
+MAX_TELEMETRY_WARNINGS = 20
+"""Most recent telemetry warnings to log, so a long run cannot flood the output."""
+
+MAX_STACK_DUMP_CHARS = 20000
+"""Longest stack dump to log, so a process with many threads cannot flood the output."""
+
+_STACK_DUMP_COMMANDS: dict[str, str] = {
+    # `timeout` bounds the command so a hung dump cannot extend the apply, and
+    # --nonblocking keeps py-spy from pausing the process we are measuring
+    "python": f"timeout {STACK_DUMP_TIMEOUT} py-spy dump --pid 1 --subprocesses --nonblocking",
```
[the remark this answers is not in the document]
christophe-papazian: Checked this against py-spy 0.4.1 inside the built weblog image, and `--subprocesses` is supported on `dump`:

```
$ docker run --rm --entrypoint sh system_tests/weblog -c 'py-spy --version && py-spy dump --help'
py-spy 0.4.1
    -s, --subprocesses      Profile subprocesses of the original process
        --nonblocking       Don't pause the python process when collecting samples
```

It is also load-bearing rather than merely valid. Running the scenario locally with the dump deadlines lowered so they fire on a normal apply produced 8 dumps and 0 failures, covering both processes:

```
WARNING Weblog thread stacks 0.4s into a stalled RC apply:
Process 1: gunicorn -w 1 --threads 1 -b 0.0.0.0:7777 app:app -k gevent
Thread 1 (idle): "MainThread"
    sleep (gunicorn/arbiter.py:360)
--------------------------------------------------------------------------------
Process 28: gunicorn ... (Parent Process 1)
Thread 28 (idle)
    run (gevent/hub.py:647)
Thread 34 (idle)
    periodic (ddtrace/internal/remoteconfig/_subscribers.py:48)
```

PID 1 is the gunicorn master, which is idle in `arbiter.sleep`. The thread we need is `Thread 34` in the worker (Process 28), reached only via `--subprocesses`. Dropping the flag would capture just the master and lose the remote-config poller entirely, so I am leaving it in.

The exit-code check added in 7fdac44e also means a genuinely failing command is now reported as a failure rather than logged under a "thread stacks" heading.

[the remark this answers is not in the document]
christophe-papazian: Correct, and thanks: I had misread this. `SYS_PTRACE` is added by `enable_core_dumps()`, whose only caller is `if library in ("php", "cpp_nginx")`, so python weblogs were not getting it. (The other `cap_add=["SYS_PTRACE"]` in that file is the mssql container.)

Confirmed the capability is genuinely required, by running the same command in a container without it:

```
$ docker exec <no-cap-container> py-spy dump --pid 1 --subprocesses --nonblocking
Error: Permission denied (os error 13)
```

Fixed in e0c291a5 with a narrow `enable_ptrace()` called for python only. It deliberately does not reuse `enable_core_dumps()`, which also sets `privileged=True` and core ulimits that are not needed here:

```
python   cap_add=['SYS_PTRACE']
java     cap_add=None
php      cap_add=['SYS_PTRACE', 'SYS_ADMIN']
golang   cap_add=None
```

With the fix in place, a local run of APPSEC_RUNTIME_ACTIVATION produced 8 real stack dumps and 0 failures, so the capability is both necessary and sufficient.

utils/build/docker/python/anthropic-py.Dockerfile
```diff
@@ -18,6 +18,10 @@ COPY utils/build/docker/python/anthropic_app/system_tests_library_version.sh sys
 COPY utils/build/docker/python/install_ddtrace.sh binaries* /binaries/
 
 RUN /binaries/install_ddtrace.sh
+
+# py-spy lets system-tests dump this weblog's thread stacks from outside the
+# process when a remote config apply stalls (see utils/_remote_config.py)
+RUN pip install --no-cache-dir py-spy==0.4.1
```
bouwkast: It would be beneficial to clarify why we aren't using the latest version of py-spy as it isn't obvious.

https://github.com/benfred/py-spy/releases/tag/v0.4.2

bouwkast (approved):
Only reviewed owned files in `utils/build/docker/`

It isn't obvious to me why we aren't using the latest py-spy.

utils/build/docker/python/anthropic-py.Dockerfile
christophe-papazian: Good catch, and there was no good reason: 0.4.1 was just the version I happened to verify first. It should be 0.4.2, because 0.4.2 fixes a bug in the exact mode we use here.

[benfred/py-spy#832](https://github.com/benfred/py-spy/pull/832) ("Fix corrupted `subprocess` output") describes `dump --subprocesses` printing each child process once per parent Python thread. We reproduced it locally on 0.4.1 without noticing at first: the gunicorn worker appeared 4 times in a single dump, once per python thread in the master.

Process headers within one dump, same scenario, before and after:

```
py-spy 0.4.1 -> {Process 1: 1, Process 28: 4}
py-spy 0.4.2 -> {Process 1: 1, Process 28: 1}
```

Bumped to 0.4.2 in 93717d4f. Re-ran APPSEC_RUNTIME_ACTIVATION locally against the rebuilt image: 8 dumps, 0 failures, each process printed once.

Keeping it pinned rather than floating, to match the other pinned pip installs in these Dockerfiles.

utils/_context/containers.py
```diff
@@ -1159,6 +1167,11 @@ def configure(self, *, host_log_folder: str, replay: bool):
         if library in ("php", "cpp_nginx"):
             self.enable_core_dumps()
 
+        if library == "python":
```
cbeauchesne, line 1170: Do you really want to activate this unconditionnally on all runs? Will it have an impact (CPU, mem)? 

My concern is having some issues that never happens in real conditions, the alternative would be to add an option to activate it on demand.

utils/build/docker/python/django-poc.Dockerfile
```diff
@@ -6,6 +6,10 @@ WORKDIR /app
 COPY utils/build/docker/python/install_ddtrace.sh binaries* /binaries/
 RUN /binaries/install_ddtrace.sh
 
+# py-spy lets system-tests dump this weblog's thread stacks from outside the
+# process when a remote config apply stalls (see utils/_remote_config.py)
+RUN pip install --no-cache-dir py-spy==0.4.2
```
cbeauchesne: Can you add it in the base file ? Same for others

utils/interfaces/_core.py
```diff
@@ -129,7 +129,12 @@ def get_data(self, path_filters: Iterable[str] | str | None = None):
         else:
             path_regexes = None
 
-        for data in self._data_list:
+        # snapshot under the lock: the watchdog thread appends to and re-sorts
+        # _data_list, which would make an in-place walk skip or repeat entries
+        with self._lock:
```
cbeauchesne, line 134: Not a python export, but is this thread safe? 

```
 for data in list(data_list):
```

cbeauchesne (changes requested):

utils/_context/containers.py
christophe-papazian: The impact is almost non existant both for cpu/memory and image size.
And currently I face at least one failure that I was not able to reproduce, even on the "do that test 8000 times", we need more visibility if something goes wrong, even if it's very rarely.

utils/interfaces/_core.py
christophe-papazian: with "with self._lock", yes it is.

christophe-papazian: otherwise no, the for loop is not thread safe with the list, even if it should be pretty safe if we just do append on the list.

utils/build/docker/python/django-poc.Dockerfile
christophe-papazian: Done in 3447d673.

Moved the install into the 7 python base Dockerfiles and bumped each one a version, so the 8 weblogs deriving from them inherit it instead of repeating the line:

| base image | old -> new |
| --- | --- |
| django-poc | v12 -> v13 |
| django-py3.13 | v11 -> v12 |
| fastapi | v9 -> v10 |
| flask-poc | v15 -> v16 |
| python3.12 | v14 -> v15 |
| tornado | v2 -> v3 |
| uwsgi-poc | v11 -> v12 |

References updated: the `docker build`/`docker push` comments in each base Dockerfile, the leaf `FROM` lines, `uds-flask` (derives from `flask-poc.base`), `docker-bake.hcl`, and `mirror_images.yaml` via `update_mirror_images.py --skip-lock`. The `fastapi` and `tornado` push comments were already stale (v8 and v1) and now match. Re-grepped each old version afterwards: no stale references left.

Verified rather than assumed: built `tornado.base-v3` locally, then built the tornado leaf on it, and `py-spy 0.4.2` is present in the final weblog image with no leaf install.

Two deliberate exceptions:

- `anthropic-py`, `openai-py` and `google_genai-py` build from `python:3.11-slim` with no system-tests base image, so they keep the install in the leaf.
- `python_otel/flask-poc-otel.Dockerfile` stays on `flask-poc.base-v1`; it pins an old base on purpose, so bumping it would have broken it.

Two things this now needs before CI can pass, since the leaves point at tags that do not exist yet:

1. the 7 new base images have to be built and pushed (commands are in each base Dockerfile's trailing comments) — I do not have push rights;
2. `mirror_images.lock.yaml` then needs regenerating with `python utils/scripts/update_mirror_images.py` (digests cannot be resolved before the images exist, and the mirror job never updates the lock itself).

Happy to do either if you can point me at the process, or to hand them over.

Unrelated small thing: `update_mirror_images.py` strips the header comment from `mirror_images.yaml` (the block saying the file is generated). I restored it by hand rather than commit the loss, but it will happen to the next person too.

utils/build/docker/python/anthropic-py.Dockerfile
```diff
@@ -18,6 +18,10 @@ COPY utils/build/docker/python/anthropic_app/system_tests_library_version.sh sys
 COPY utils/build/docker/python/install_ddtrace.sh binaries* /binaries/
 
 RUN /binaries/install_ddtrace.sh
+
+# py-spy lets system-tests dump this weblog's thread stacks from outside the
+# process when a remote config apply stalls (see utils/_remote_config.py)
+RUN pip install --no-cache-dir py-spy==0.4.2
```
cbeauchesne: QQ : why not putting this in the line 10 ? If it's on purpose, can you leave a comment, to prevent someone to do this later on.

utils/build/docker/python/django-poc.base.Dockerfile
```diff
@@ -11,5 +11,9 @@ ENV PIP_ROOT_USER_ACTION=ignore
 COPY utils/build/docker/python/django/requirements-django-poc.txt /tmp/django-requirements.txt
 RUN pip install --upgrade pip && pip install -r /tmp/django-requirements.txt
 
-# docker build --progress=plain -f utils/build/docker/python/django-poc.base.Dockerfile -t datadog/system-tests:django-poc.base-v12 .
-# docker push datadog/system-tests:django-poc.base-v12
+# py-spy lets system-tests dump a weblog's thread stacks from outside the
+# process when a remote config apply stalls (see utils/_remote_config.py)
+RUN pip install --no-cache-dir py-spy==0.4.2
```
cbeauchesne: Is it possible to put it in `utils/build/docker/python/django/requirements-django-poc.txt` ?

utils/build/docker/python/django-py3.13.base.Dockerfile
```diff
@@ -10,5 +10,9 @@ ENV PIP_ROOT_USER_ACTION=ignore
 COPY utils/build/docker/python/django/requirements-django-py3.13.txt /tmp/django-requirements.txt
 RUN pip install --upgrade pip && pip install -r /tmp/django-requirements.txt
 
-# docker build --progress=plain -f utils/build/docker/python/django-py3.13.base.Dockerfile -t datadog/system-tests:django-py3.13.base-v11 .
-# docker push datadog/system-tests:django-py3.13.base-v11
+# py-spy lets system-tests dump a weblog's thread stacks from outside the
+# process when a remote config apply stalls (see utils/_remote_config.py)
+RUN pip install --no-cache-dir py-spy==0.4.2
```
cbeauchesne: Same, is it possible to put it in `utils/build/docker/python/django/requirements-django-py3.13.txt`?

utils/build/docker/python/fastapi.base.Dockerfile
```diff
@@ -13,5 +13,9 @@ RUN pip install --upgrade pip && pip install -r /tmp/fastapi-requirements.txt
 RUN mkdir app
 WORKDIR /app
 
-# docker build --progress=plain -f utils/build/docker/python/fastapi.base.Dockerfile -t datadog/system-tests:fastapi.base-v8 .
-# docker push datadog/system-tests:fastapi.base-v8
+# py-spy lets system-tests dump a weblog's thread stacks from outside the
+# process when a remote config apply stalls (see utils/_remote_config.py)
+RUN pip install --no-cache-dir py-spy==0.4.2
```
cbeauchesne: Is it possible to as it in `utils/build/docker/python/fastapi/requirements-fastapi.txt` ?

utils/scripts/update_mirror_images.py
```diff
@@ -135,6 +135,40 @@ def _refresh_lock_file() -> str | None:
     return original_lock
 
 
+def _read_yaml_preamble() -> str:
```
cbeauchesne: Is this change belongs to this PR ?

utils/_remote_config.py
```diff
@@ -46,6 +48,153 @@ def from_json(d: dict) -> "RemoteConfigStateResults":
         return RemoteConfigStateResults(version=d["version"], state=d["state"], configs=d["configs"])
 
 
+SLOW_APPLY_THRESHOLD = 12.0
```
cbeauchesne: Does this part requires a careful review ? 😅

cbeauchesne (changes requested):
* Can you also  update `mirror_images.lock.yaml`  ? 
* I've left few comments for where are stored the new dep, but not on all files

utils/build/docker/python/django-poc.base.Dockerfile
christophe-papazian: Done in 9e6f152d — `py-spy==0.4.2` is now in `requirements-django-poc.txt` and the separate `RUN` is gone.

Applied to all 7 python base images, not just the three you flagged, since they all install from a `requirements-*.txt`:

| base | file |
| --- | --- |
| django-poc | `django/requirements-django-poc.txt` |
| django-py3.13 | `django/requirements-django-py3.13.txt` |
| python3.12 | `django/requirements-python3.12.txt` |
| fastapi | `fastapi/requirements-fastapi.txt` |
| flask-poc | `flask/requirements-flask-poc.txt` |
| uwsgi-poc | `flask/requirements-uwsgi-poc.txt` |
| tornado | `tornado/requirements-tornado.txt` |

One layer fewer per image, and it sits with the other pinned dependencies. Verified by rebuilding `tornado.base-v3` from scratch: `py-spy 0.4.2` is present with no `RUN` line.

The base image version bumps still stand, since the requirements files are baked into those images.

utils/build/docker/python/django-py3.13.base.Dockerfile
christophe-papazian: Done, same commit: it is in `django/requirements-django-py3.13.txt` now.

utils/build/docker/python/fastapi.base.Dockerfile
christophe-papazian: Done, same commit: it is in `fastapi/requirements-fastapi.txt` now.

utils/build/docker/python/anthropic-py.Dockerfile
christophe-papazian: Not on purpose, so I moved it to line 10 rather than leave a comment defending the position. It was just where my patch landed relative to `install_ddtrace.sh`.

`anthropic-py`, `openai-py` and `google_genai-py` have no `requirements-*.txt`, so py-spy now joins the existing pip install line instead:

```
RUN python -m pip install fastapi==0.89.1 anyio==4.14.2 uvicorn==0.20.0 opentelemetry-exporter-otlp==1.36.0 py-spy==0.4.2
```

utils/scripts/update_mirror_images.py
christophe-papazian: No, it does not. Removed from this branch.

For context on where it came from: adding the base image tags made `update_mirror_images.py` add images, and `mirror_images.py add` rewrites the file through a YAML parser that drops comments, so regenerating deleted the header of `mirror_images.yaml`, including the line explaining how to regenerate it. It only happens when the script actually adds something, which is why CI never caught it.

So it is a real bug, but independent of remote config diagnostics. I will open it separately.

utils/_remote_config.py
christophe-papazian: The value was computed from more than a hundred points in the last system tests CI to only trigger on suspiciously slow remote config requests.

PR #7626, opened 2026-09-01
Support agent-side v1 trace payload conversion in tests

ajgajg1134:
## Motivation

The trace agent can convert incoming v0.4 payloads to the efficient trace payload format (`idxTracerPayloads`) before forwarding them. Three test groups still assumed the legacy agent payload shape and broke under that conversion.

These changes are **format-agnostic**: they pass both with an agent that converts and with one that does not, so they can land ahead of the agent change.

## Changes

**Config consistency** — the efficient format merges `meta` and `metrics` into a single typed `attributes` map, so a tag a tracer reports both as a string tag and as a numeric metric collides, and the numeric value survives. The Rust tracer does this for `http.status_code`, producing `assert 400.0 == '400'` while every other library passed. Adds `DataDogAgentSpan.get_tag()`, which normalizes integral numbers back to their string form **for that format only**. The legacy format keeps `meta` and `metrics` separate, so values are returned untouched there and a numeric value in `meta` remains a failure, as it should.

**Exception replay (java)** — a converted span carries a `_dd.convertedv1` provenance marker, and v0.4 `meta` values stay strings where a natively emitted v1 payload keeps real booleans. Drop the marker and canonicalize booleans on *both* sides of the approval comparison, so a single set of approval files stays valid either way. Normalizing is preferred over re-recording the baselines, which would bake the current conversion path into them.

**Code origin (python)** — `_wait_for_code_origin_span` indexed the raw payload with legacy-only keys (`tracerPayloads`, `resource`, `type`, `meta`), so under v1 it never matched a span and timed out into a misleading "code origin not enabled" assertion. Now iterates `interfaces.agent.get_spans()`, whose typed wrappers resolve those fields for both formats.

**Exception replay (php)** — the `*_spans_expected_v1.json` approvals were never generated, so the tests raised `FileNotFoundError`. Recorded into the existing `1.18.0` folder alongside the legacy files.

## Compatibility

- The `_v1` approval suffix is selected at runtime from the observed span format, so a legacy-format agent still reads the untouched `spans_expected` files.
- Under v1, approvals hold real booleans while received data holds strings when the agent converted a v0.4 payload — normalizing both sides matches either direction. Dropping `_dd.convertedv1` is a no-op when the marker is absent.
- `get_tag` is gated on the format enum, so legacy behavior is unchanged.

## Validation

Carried over from the exploratory branch, validated per language with the converting agent **and** with the default agent, confirming the payload format in each run via the captured agent interface files:

| weblog | scenario | result |
| --- | --- | --- |
| java/spring-boot | `DEBUGGER_EXCEPTION_REPLAY` | 8 passed both ways |
| python/flask-poc | `DEBUGGER_INPRODUCT_ENABLEMENT` | 5 passed both ways |
| php/apache-mod-8.0 | `DEBUGGER_EXCEPTION_REPLAY` | 9 passed both ways |

`get_tag` verified in replay mode against the rust/axum CI artifact: the two failing tests pass and the full `DEFAULT` scenario matches CI exactly (67 passed / 386 skipped / 43 xfailed / 3 xpassed) with no other change.

`ruff check`, `ruff format --check`, and `mypy` pass on the changed Python files.


cbeauchesne (approved):
AGTM. I advise you to get a review from someone in debugger team

tylfin (approved):

PR #7628, opened 2026-09-01
Send APM_TRACING configs as sdk_config when the library advertises SDK_CONFIGURATION

rachelyangdog:
### Motivation

[dd-trace-js#9392](https://github.com/DataDog/dd-trace-js/pull/9392) is the first library to move to the unified **Config at Runtime** contract, and it fails 7 system-tests jobs:

- `DEBUGGER_INPRODUCT_ENABLEMENT` × 5 weblogs — `Test_Debugger_InProduct_Enablement_Dynamic_Instrumentation::test_inproduct_enablement_di`
- `PARAMETRIC` × 2 — 10 behaviour tests + 7 capability tests in `test_dynamic_configuration.py`

That PR swaps the whole APM_TRACING contract:

|  | before | after |
| --- | --- | --- |
| capabilities | `APM_TRACING_SAMPLE_RATE`, `_LOGS_INJECTION`, `_HTTP_HEADER_TAGS`, `_CUSTOM_TAGS`, `_ENABLED`, `_SAMPLE_RULES`, `_ENABLE_DYNAMIC_INSTRUMENTATION`, `_ENABLE_CODE_ORIGIN`, `_ENABLE_LIVE_DEBUGGING` | `SDK_CONFIGURATION` (bit 49) |
| payload | `lib_config: {dynamic_instrumentation_enabled: true}` | `sdk_config: {service_name, env, config: [{key: "DD_DYNAMIC_INSTRUMENTATION_ENABLED", value: "true"}]}` |

The `lib_config` branch is **removed**, not kept as a fallback — `RCClientManager.addConfig` only reads `conf.sdk_config?.config`. system-tests only ever emits `lib_config`, so every APM_TRACING config becomes a no-op: DI is never enabled, probes never reach `EMITTING`, and the test fails.

### What does this PR do?

The RC helpers now read the capability bit the library advertises on `/v0.7/config` and, when `SDK_CONFIGURATION` is set, send the same settings in the new shape. `service_target` is untouched — it still drives matching and priority.

- **`utils/_remote_config.py`** — `to_sdk_config_payload()` rewrites an APM_TRACING config from `lib_config` to `sdk_config`: each setting keyed by its canonical `DD_*` name, serialized to its environment variable form (booleans, sample rate, `tracing_header_tags` → `"h:tag,…"`, `tracing_tags` → `"k:v,…"`, `tracing_service_mapping` → `"a:b,…"`, `tracing_sampling_rules` → JSON with the `[{key, value_glob}]` tag clauses folded into a map). `library_supports_sdk_configuration()` reads the bit.
- **The two APM_TRACING builders** take `use_sdk_config`; the `send_*` wrappers detect it. `lib_config` stays the internal representation and `prev_payloads` keeps it, so the "empty config keeps the previous value" inheritance the debugger helpers rely on is unchanged.
- **`tests/parametric/test_dynamic_configuration.py`** — translation happens in `_set_rc`, the single choke point that owns `test_agent`. `assert_rc_capability()` accepts the legacy per-setting bit **or** `SDK_CONFIGURATION`, since one bit now covers them all.
- **`tests/parametric/capabilities.yml`** — nodejs's per-setting bits scoped to `<7.0.0-0`.
- **`tests/test_library_conf.py`** — `Test_HeaderTags_DynamicConfig` routed through the same translation (`missing_feature` for nodejs today, but it builds its own APM_TRACING payload).
- **`tests/test_the_test/test_remote_config.py`** — 7 unit tests covering the translation, the value serializers, and that the mapping stays exhaustive over what the builders can emit.

**Nothing changes for any library that does not advertise the bit** — today that is every library, including nodejs on `main`.

### Notes for reviewers

- **`capabilities.yml`**: `SDK_CONFIGURATION` is deliberately *not* listed as expected yet. dd-trace-js `main` is `7.0.0-pre` and does not advertise it until #9392 lands; a pre-release is allowed to report capabilities beyond the expected set, so the `<7.0.0-0` bound alone covers both sides of the transition without a chicken-and-egg break. A follow-up should add `'>=7.0.0-0': [SDK_CONFIGURATION]` once #9392 merges. The bound assumes the removal ships in 7.0.0; if a 7.0.0 is cut before #9392 lands, the bound needs moving.
- **`dynamic_sampling_enabled` and `live_debugging_enabled`** are remote-config-only settings with no environment variable, so they cannot be expressed as `sdk_config` and are dropped for libraries on the new contract. Both are already ignored by dd-trace-js, so nothing regresses; the mapping records them explicitly as `None` rather than guessing a name.
- **`provenance` on sampling rules** is carried through in the `DD_TRACE_SAMPLING_RULES` JSON. dd-trace-js's `SamplingRule` reads it, so `_dd.p.dm` `-11`/`-12` should still be produced through the env-var path — worth confirming on the #9392 run, as it is a library-side behaviour, not a payload one.

### Testing
- `./run.sh TEST_THE_TEST` — 445 passed
- `./format.sh` — clean
- Replayed the 4-step RC sequence of `test_inproduct_enablement_di` through the js reader's allowlist/merge logic: unset → `{}`, enable → `DD_DYNAMIC_INSTRUMENTATION_ENABLED=true`, empty → inherits `true`, disable → `false`.

### DIFF
--- a/utils/_remote_config.py
+++ b/utils/_remote_config.py
@@ -13,11 +13,11 @@ import threading
 import time
 import uuid
 from typing import Any, Literal
-from collections.abc import Iterator, Mapping
+from collections.abc import Callable, Iterator, Mapping
 
 from utils._context.core import context
 
-from utils.dd_constants import RemoteConfigApplyState as ApplyState
+from utils.dd_constants import Capabilities, RemoteConfigApplyState as ApplyState
 from utils.interfaces import library
 from utils._logger import logger
 from utils.proxy.mocked_response import (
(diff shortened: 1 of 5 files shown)


rachelyangdog:
### Update: two bugs found in the first CI run, both fixed in 79eb84b

#### 1. Capability bit 49 is ambiguous today (this is the interesting one)

`SDK_CONFIGURATION` is bit 49 per the remote config source of truth — [`dd-source remote-config/shared/libs/rc/capabilities.go`](https://github.com/DataDog/dd-source/blob/main/domains/remote-config/shared/libs/rc/capabilities.go):

```go
// CapabilitySDKConfiguration is the capability for remotely configuring any in-scope
// SDK setting at runtime via a single, env-var-keyed capability bit
CapabilitySDKConfiguration = Capability{
	Name:     "SDK_CONFIGURATION",
	bitIndex: 49,
}
```

But **libdatadog gives the same bit to a different capability**:

```c
DDOG_REMOTE_CONFIG_CAPABILITIES_ASM_RAW_RESPONSE_BODY = 49,
```

dd-trace-php ships that enum, so it advertises bit 49 today while still reading `lib_config`. Captured from the failing job's artifact (php 1.24.2, `apache-mod-7.4-zts`), the client capabilities `[2, 44, 15, 160, 235, 247, 252]` decode to bits `…, 42, 43, 45, 49`. The first revision read that as "php supports SDK_CONFIGURATION", sent it `sdk_config`, and DI never enabled — `test_inproduct_enablement_di` failed on all 40 php weblog jobs. Java (highest bit 45) was unaffected, which is why only php broke.

**Fix:** the switch now also requires the per-setting APM_TRACING capabilities to be *gone*. Dropping `APM_TRACING_SAMPLE_RATE`/`_LOGS_INJECTION`/`_HTTP_HEADER_TAGS`/`_CUSTOM_TAGS`/`_ENABLED`/`_SAMPLE_RULES` is the entire point of the unified bit, and a library still advertising them still understands `lib_config` — so the two are told apart without depending on a bit whose meaning is currently contested. Verified against the real bitmasks captured from this run:

| library | bit 49 | resolved shape |
| --- | --- | --- |
| php 1.24.2 (real) | set | `lib_config` |
| java prod (real) | unset | `lib_config` |
| nodejs main (real) | unset | `lib_config` |
| nodejs + dd-trace-js#9392 | set | `sdk_config` |

`assert_rc_capability` uses the same rule, so bit 49 alone can no longer let a libdatadog-based tracer skip a per-setting capability assertion.

> [!IMPORTANT]
> **The bit collision itself still needs an owner.** libdatadog and dd-source disagree about bit 49, and dd-trace-js#9392 is about to start using it for real. One of the two has to move before `SDK_CONFIGURATION` can be trusted on its own. The guard here keeps system-tests correct meanwhile, but it is a workaround, not the fix.

#### 2. Stale-ACK race in the parametric helper

`set_and_wait_rc` calls `test_agent.clear()` before reusing a `config_id`, to discard the ACKs of the previous config at that path. The capability read sat *after* that clear and blocks until the next RC request is recorded — and that request still carries the tracer's ACK of the old config. `wait_for_rc_apply_state` matched it and returned before the update had been applied, so the test read a trace still sampled at the old rate:

```
E       assert 0.5 == 0.6 ± 6.0e-07
```

That failed `TestDynamicConfigV1::test_trace_sampling_rate_override_env` on rust, golang, nodejs, ruby, java and cpp — language-independent, because it is purely a helper-ordering bug.

**Fix:** resolve the payload shape at the top of `set_and_wait_rc`, before the clear, and memoize it once the capabilities are conclusive so `_set_rc` never polls inside the clear/update window.

#### Not caused by this PR

`test_config_consistency.py` and `test_128_bit_traceids.py` on cpp, and the `golang net-http-orchestrion` build job — unrelated to remote config, present independently of this change.

#### Testing
`./run.sh TEST_THE_TEST` — 446 passed. New unit test pins the resolution against the real php / java / nodejs bitmasks captured above.

cbeauchesne (approved):

utils/_remote_config.py
```diff
@@ -283,6 +283,206 @@ def send_symdb_command(version: int = 1) -> RemoteConfigStateResults:
...
+    return None
+
+
+# Memoized once the capabilities are conclusive, both to skip re-scanning the whole /v0.7/config
+# history on every payload and because a library that has not registered its APM_TRACING
+# capabilities yet reports a partial set, which must not be cached.
+_sdk_configuration_support: dict[str, bool] = {}
+
+
+def library_supports_sdk_configuration() -> bool:
```
mtoffl01: This seems like almost the exact same function as `uses_sdk_configuration`; can we collapse them into one shared helper in this utils file?

utils/_remote_config.py
```diff
@@ -283,6 +283,206 @@ def send_symdb_command(version: int = 1) -> RemoteConfigStateResults:
...
+# Canonical environment variable name of each `lib_config` setting. A `None` marks a setting
+# that only ever existed as a remote config and has no environment variable counterpart, so
+# there is nothing to send for it under SDK_CONFIGURATION: it is dropped rather than guessed.
+APM_TRACING_ENV_VAR_NAMES: dict[str, str | None] = {
+    "code_origin_enabled": "DD_CODE_ORIGIN_FOR_SPANS_ENABLED",
+    "data_streams_enabled": "DD_DATA_STREAMS_ENABLED",
+    "dynamic_instrumentation_enabled": "DD_DYNAMIC_INSTRUMENTATION_ENABLED",
+    "dynamic_sampling_enabled": None,
+    "exception_replay_enabled": "DD_EXCEPTION_REPLAY_ENABLED",
+    "live_debugging_enabled": None,
```
mtoffl01, line 551: What are the settings "that only ever existed as a remote config and has no environment variable counterpart"? Can we just delete them from this list, then?

mtoffl01 (approved):

utils/_remote_config.py
```diff
@@ -283,6 +283,206 @@ def send_symdb_command(version: int = 1) -> RemoteConfigStateResults:
...
+    libdatadog hands the same bit to `ASM_RAW_RESPONSE_BODY`, so a libdatadog-based library such as
+    dd-trace-php advertises it while still reading `lib_config`. Requiring the per-setting
+    capabilities to be *gone* tells the two apart: dropping them is the whole point of the unified
+    bit, and a library still advertising them still understands `lib_config`.
+    """
+    advertises_sdk_configuration = Capabilities.SDK_CONFIGURATION in capabilities
+    advertises_legacy = bool(capabilities & LEGACY_APM_TRACING_CAPABILITIES)
+
+    if advertises_sdk_configuration and not advertises_legacy:
+        return True
```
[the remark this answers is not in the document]
rachelyangdog: Valid, and fixed in 23eb101b2 — thanks, this was a real race, not a false positive.

The repo's own helper documents exactly the window you describe: *"capabilities are added as products start but never removed. Earlier requests may therefore carry a partial set (e.g. only AppSec capabilities, before APM capabilities are registered)"* (`utils/docker_fixtures/_test_agent.py`). And bit 49 **is** an AppSec capability in libdatadog (`ASM_RAW_RESPONSE_BODY`), so during that window PHP shows bit 49 with none of the legacy APM bits — read as "migrated", memoized, and `sdk_config` sent to a library that only reads `lib_config`.

The flaw was treating *absence* as evidence, when absence is also what "not registered yet" looks like. Now the absence only counts once the library has advertised some APM_TRACING capability:

```python
if not capabilities & APM_TRACING_CAPABILITIES:
    return None                                    # nothing to conclude yet
if capabilities & LEGACY_APM_TRACING_CAPABILITIES:
    return False                                   # per-setting bits -> lib_config
return Capabilities.SDK_CONFIGURATION in capabilities
```

`APM_TRACING_CAPABILITIES` excludes `SDK_CONFIGURATION` itself, so the ambiguous bit cannot serve as its own evidence; a unit test pins that. Verified against the real bitmask captured from the failing PHP job: full set -> `lib_config`, mid-startup (ASM + bit 49) -> inconclusive -> `lib_config`, nodejs post-migration -> `sdk_config`.

One small correction: only the parametric caller consumes the most recent non-empty poll. The end-to-end one unions across all requests, so it was less exposed — but exposed all the same.

utils/_remote_config.py
rachelyangdog: Agreed — done in 23eb101b2.

They now share `resolve_sdk_configuration_support(get_capabilities)`, which owns the memo, the contract call and the inconclusive policy. Worth noting this caught something I had left sloppy: there were **two separate memo dicts**, one per module, which only worked because a single session never exercises both paths. Now there is one. It also meant the Codex fix in this same push landed in one place instead of two — which matters for a race-condition guard that only one of the copies had tests pointed at.

I stopped short of collapsing the waits and the error handling, since those differ for a reason: the end-to-end path waits on `library.wait_for(.../v0.7/config)` and catches broad `Exception` because it runs inside `setup_*` methods, which per the repo rules must never raise; the parametric path polls `wait_for_rc_capabilities` and runs in test bodies where raising is fine. Unifying those would either weaken the setup-cannot-fail guarantee or over-broaden the parametric handler, so they stay at the call sites — the parametric wrapper is now a single delegating call.

utils/_remote_config.py
rachelyangdog: They are `dynamic_sampling_enabled` and `live_debugging_enabled`. Fair question — I have named them in the comment in 23eb101b2 so it is not a hunt.

I would rather not delete them, though. Both are real settings, not dead entries: dd-go's `LibConfig` declares each one, and `live_debugging_enabled` has its own capability bit (41, `APM_TRACING_ENABLE_LIVE_DEBUGGING`) distinct from `dynamic_instrumentation_enabled`. They simply never got a `DD_*` env var, so there is nothing to put in an env-var-keyed payload.

Removing them would cost two things:

1. They would fall through to the *unknown setting* branch, so every payload carrying one would log `No environment variable known for lib_config.dynamic_sampling_enabled` — and parametric's `test_trace_sampling_rules_with_tags` does send it. That warning is meant to mean "someone forgot a mapping", so it would read as a bug where the omission is deliberate.
2. It breaks `test_apm_tracing_command_settings_are_all_mapped`, which exists to catch an actually-forgotten mapping. I tried it to be sure:

```
AssertionError: APM_TRACING_ENV_VAR_NAMES is missing an entry for
  {'live_debugging_enabled', 'dynamic_sampling_enabled'}
```

So the `None` is what separates "known to have no env var" from "someone forgot". Happy to revisit if you would rather encode that distinction another way.

PR #7632, opened 2026-09-02
test(dotnet): revert easywins for OtelTracestate sampling, disable IAST in sampling scenario

MilanGarnier:
These were activated by easy win script but some will be marked as bug
Moreover, IAST flags the payload in `make_distant_call` and makes a force keeps which would break the test in python, dotnet, Java and PHP. Action: move the impacted test to the sampling scenario, and restrict the a SamplingTest to its own spans (because the moved test involve manual keep)

## Summary

- disable the .NET parametric and end-to-end OTel tracestate sampling suites
- run every end-to-end OTel tracestate test in SAMPLING instead of DEFAULT
- disable AppSec and IAST in the SAMPLING scenario
- remove only the OTel tracestate activations introduced by the automated easy-win PR

All .NET declarations retain the APMAPI-2171 missing-feature marker.

### DIFF
--- a/manifests/dotnet.yml
+++ b/manifests/dotnet.yml
@@ -1097,14 +1097,7 @@ manifest:
       component_version: <=2.41.0
   tests/parametric/test_otel_span_with_baggage.py::Test_Otel_Span_With_Baggage: missing_feature
   tests/parametric/test_otel_tracer.py::Test_Otel_Tracer: v2.8.0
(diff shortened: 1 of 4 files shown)


genesor (approved):

MilanGarnier:
Tried to use decision maker (checking manual keep), but dd-trace-dotnet does not emit it for the root span so I fall back on limiting the scope to the spans emitted for the test.

cbeauchesne (approved):

PR #7663, opened 2026-09-07
Keep mirror_images.yaml's header when regenerating

christophe-papazian:
APPSEC-69943

`mirror_images.py add` rewrites `mirror_images.yaml` through a YAML parser, which discards comments — but only when it actually adds an image. Regenerating after adding one therefore deleted the file header, including the line documenting how to regenerate it. Steady-state CI never catches this, because nothing is added, so the file is not rewritten.

Fix: capture the committed preamble before the `add` and re-apply it after. It preserves what is in the file rather than rewriting from this script's `MIRROR_YAML_HEADER` constant, whose wording has drifted from it.

Verified by removing an entry and regenerating: the entry comes back, the header survives, and the file is byte-identical to before.

Split out of #7625, where adding the python base image tags hit this.


utils/scripts/update_mirror_images.py
```diff
@@ -135,6 +136,47 @@ def _refresh_lock_file() -> str | None:
...
+            break
+    return "".join(preamble)
+
+
+def _has_document_marker(text: str) -> bool:
+    """Whether `text` contains a YAML document start on a line of its own."""
+    return any(line.strip() == "---" for line in text.splitlines())
+
+
+def _restore_yaml_preamble(preamble: str) -> None:
```
nccatoni: I would be inclined to make MIRROR_YAML_HEADER authoritative and prevent edits to the yaml file directly.

I can push a commit to this branch to do that if you don't mind

nccatoni (dismissed):

nccatoni:
It needs upstream changes to be merged first
https://github.com/DataDog/dd-repo-tools/pull/12

nccatoni (approved):

PR #7685, opened 2026-09-09
perf(ffe): remove unnecessary nodejs waits

rochdev:
## Motivation

PR #7488 removes the largest FFE timeout regression, but the Node.js Express 4
scenario still spends about 100 seconds in fixed Remote Config settling delays
and another 30 seconds executing a known expected failure.

Local profiling showed that Node.js acknowledges each FFE configuration only
after it has been applied. The generic two-second post-ack delay therefore adds
no value for this scenario.

## Changes

* Add a per-library Remote Config settle override to end-to-end scenarios while
  preserving the existing two-second default everywhere.
* Set the Node.js FFE settle delay to zero. Other languages, scenarios, and
  Node.js Remote Config suites retain the existing behavior.
* Mark `Test_FFE_EXP_5_Missing_Targeting_Key` as slow so its declared Node.js
  failure is skipped instead of consuming its full 30-second timeout. It will
  run normally once the manifest declaration is removed.

## Validation

* Full Express 4 FFE scenario on this stack: 29 passed, 12 skipped, 3 xfailed,
  and 8 xpassed in 73.58 seconds (84.14 seconds wall clock).
* Equivalent baseline before these two changes: 204.94 seconds of pytest time.
* Mypy: 556 source files passed.
* Ruff formatting and lint: passed with no changes.
* YAML formatting/lint and manifest validation: passed.
* The umbrella formatter later stops in the existing macOS Bash 3 shellcheck
  wrapper with `@: unbound variable`; no shell files are changed here.

## Workflow

    * Framework is modified, or non obvious usage of it -> get a review from
      [R&P team](https://dd.enterprise.slack.com/archives/C025TJ4RZ8X)

## Reviewer checklist

* [ ] Anything but `tests/` or `manifests/` is modified? I have the approval
  from [R&P team](https://dd.enterprise.slack.com/archives/C025TJ4RZ8X)
    * [ ] The relevant `build-XXX-image` label is present

*Generated by Codex.*


greghuels (approved):

utils/_remote_config.py
```diff
@@ -376,9 +376,10 @@ def remote_config_applied(data: dict) -> bool:
     else:
         logger.debug(f"RC apply version={version} acknowledged in {elapsed:.3f}s")
 
-    # ensure the library has enough time to apply the config to all subprocesses,
-    # and to flush the telemetry that explains a slow apply
-    time.sleep(2)
+    # By default, give subprocesses time to apply the config and flush telemetry.
+    # Scenarios can remove this delay for libraries that acknowledge only after
+    # the configuration has been fully applied.
+    time.sleep(context.scenario.rc_apply_settle_seconds.get(context.library.name, 2))
```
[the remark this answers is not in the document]
rochdev: I don't understand this, requests should always be acknowledged.

cbeauchesne, line 382: This may have an impact on stability, could you run the debugger harness to ensure we won't add more non-determinism ? https://github.com/DataDog/system-tests/actions/workflows/debug-harness.yml

cbeauchesne (changes requested):

utils/_remote_config.py
rochdev: It would do the opposite. Relying on sleeps is non-deterministic by design. If something is slower, it will just fail and then the sleep needs to be constantly increased to make sure it properly **hides the underlying race condition**. If there is a proper feedback loop in place, then it can run fast **because it is deterministic**.

How does the debug harness work?

rochdev: @cbeauchesne ^

cbeauchesne: > How does the debug harness work?

Sorry, I missed your comment, let's zoom for this

cbeauchesne: Added a meet tomorrow

